PR #21, opened 2023-07-18
Refactor

ericaxu:
Creates new `FormatImporter` class so we can have a bunch of importers under the names of `[app]-[extension].ts`.

Let me know if this is what you had in mind!

### DIFF
--- a/format-importer.ts
+++ b/format-importer.ts
@@ -1,0 +1,12 @@
+import { App } from "obsidian";
+import { ImportResult } from "yarle/interfaces";
+
+export abstract class FormatImporter {
+	app: App;
+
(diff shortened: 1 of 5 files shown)


format-importer.ts
```diff
@@ -0,0 +1,12 @@
...
+import { ImportResult } from "yarle/interfaces";
+
+export abstract class FormatImporter {
+	app: App;
+
+	constructor(app: App) {
+		this.app = app;
+	}
+
+        abstract import(filePaths: string[], outputFolder: string): Promise<ImportResult>;
```
lishid, line 11: Formatting.

lishid (approved):

PR #22, opened 2023-07-18
HTML Importer

ericaxu:
Adds `HTMLImporter` that import files the naive way. Builds on the `refactor` branch.

### DIFF
--- a/html.ts
+++ b/html.ts
@@ -30,12 +30,13 @@ export class HtmlImporter extends FormatImporter {
                 }
 
                 for (let path of filePaths) {
-                        let htmlContent = await fs.readFileSync(path, 'utf-8');
-                        let mdContent = htmlToMarkdown(htmlContent);
-
                         try {
-                                await this.saveAsMarkdownFile(folder as TFolder, baseFileName(normalizePath(path)), mdContent);
-                                results.total++;
+                                if (folder instanceof TFolder) {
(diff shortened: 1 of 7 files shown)


html.ts
```diff
@@ -0,0 +1,55 @@
...
+                let folder = app.vault.getAbstractFileByPath(outputFolder);
+
+                if (folder === null || !(folder instanceof TFolder)) {
+                        await app.vault.createFolder(outputFolder);
+                        folder = app.vault.getAbstractFileByPath(outputFolder);
+                }
+
+                for (let path of filePaths) {
+                        let htmlContent = await fs.readFileSync(path, 'utf-8');
+                        let mdContent = htmlToMarkdown(htmlContent);
```
lishid, line 34: We should move everything into the try-catch block.

html.ts
```diff
@@ -0,0 +1,55 @@
...
+
+                        try {
+                                await this.saveAsMarkdownFile(folder as TFolder, baseFileName(normalizePath(path)), mdContent);
```
lishid, line 37: Can we pre-verify folder to be TFolder?

lishid (commented):
Need to run the auto formatter please.

ericaxu:
> Need to run the auto formatter please.

On `html.ts`? I ran in and didn't notice a difference.

PR #31, opened 2023-07-29
Fix `FormatImporter#readPath` not actually async

polyipseity:
Just something I discovered while working on the HTML attachments.


ericaxu:
Great catch, thank you @polyipseity !

Do you think we should keep this as sync or change to async? @lishid

arthurtyukayev:
Not sure why it would need to be async.

lishid:
I would prefer to switch to async.

PR #32, opened 2023-07-29
Added Bear importer and file converter

arthurtyukayev:
- Adds importer configuration for Bear's `bear2bk` file format.
- Copies assets from each `.textbundle` into the Vault's configured Media folder, under a directory `bear-assets`. 

`.bear2bk` are just archive files (.zip) containing a folder of `.textbundle`'s. 

**This adds a dependency for reading zip files as a stream in memory, not sure how this performs with larger archives**


arthurtyukayev:
This is a file that I exported with all of Bear's default markdown files. Changing the file extension  to `.bear2bk`, should work for testing.
[notes.zip](https://github.com/obsidianmd/obsidian-importer/files/12207773/notes.zip)

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    let results: ImportResult = {
+      total: 0,
+      skipped: [],
+      failed: []
+    };
+
+    const attachmentsFolderPath = await (this.app.vault as any).getAvailablePathForAttachments(EXPORTED_ASSETS_FOLDER_NAME)
+    const existingAssetAttachmentsFolder = this.app.vault.getAbstractFileByPath(attachmentsFolderPath);
+    const assetMatcher = /!\[\]\(assets\//g
```
ericaxu, line 38: nit: missing semicolon

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,79 @@
...
+            let filePath = normalizePath(mdFilename);
+            await this.saveAsMarkdownFile(folder, pathToFilename(filePath), mdContent);
+            results.total++;
+            continue;
+          }
+
+          if (file.path.match(/\/assets\//g)) {
+            const assetFilename = path.basename(file.path);
+            const assetFileVaultPath = `${attachmentsFolderPath}/${assetFilename}`;
+            await this.app.vault.createBinary(assetFileVaultPath, await file.buffer())
```
ericaxu, line 69: nit: missing semicolon

ericaxu (changes requested):

lishid:
Is there any chance we can swap out `unzipper` for a different library? I don't like that it's nodejs-only (meaning there's no way to support mobile in the future) and has 10 dependencies.

I've done a round of research around zip utilities for javascript (seriously there isn't a standard API for it, wtf??) and it seems like I'm inclined to use `zip.js` (https://www.npmjs.com/package/@zip.js/zip.js).

What I like about `zip.js`:
- No dependencies
- Can use with `fetch` and other streaming APIs
- Allows ES6 imports of different parts for better tree shaking

I'm unsure how much of a hassle it is to use the API though, looks pretty clean.

arthurtyukayev:
That makes sense to me, I didn't spend a ton of time looking at potential libraries for zip. I think zip.js is good, I like the no dependencies as well.

ericaxu:

arthurtyukayev:

lishid:
Amazing!

Just FYI I've made some changes on master to make this plugin support mobile devices, if you can rebase on top of master, we'll be able to drop the NodeJS usage (fs, path, buffer) and solely rely on the shared common code to get the file data ❤️

arthurtyukayev:
This branch should be rebased on top of latest master.

I am using path to generate the file name for the markdown files, is there an example in main that should reference how I can replace the usages of `fs` and `path`?

lishid:
> I am using path to generate the file name for the markdown files, is there an example in main that should reference how I can replace the usages of fs and path?

You should be able to use `file.basename` directly!

In addition, I have just updated master to include zip.js in a nicely wrapped API. You can use it like this:

```ts
await file.readZip(async zip => {
	for (let zipFile of await zip.getEntries()) {
		// Metadata
		console.log(zipFile.filename, zipFile.uncompressedSize);
		// Contents
		console.log(await zipFile.getData(new TextWriter()));
	}
});
```

arthurtyukayev:
Thanks! That looks good! Just wondering if it's better to use promises to avoid having to use callbacks?

lishid:
> Thanks! That looks good! Just wondering if it's better to use promises to avoid having to use callbacks?

Generally yeah, unfortunately in this case the file descriptor had to be closed after the read operations are all done, and I can't think of a better way of handling that, other than letting the user remember to close the handle...

lishid:
Also just fixed a bug with the zip interface where it was closing the file descriptor too early, if you're already working on top of master you'll have to rebase again, sorry!

lishid:
Oh actually, I misunderstood what you needed to replace `path` with. You can now use `parseFilePath` from `util.ts` (it will work on a file name just as well) to split the basename + extension.

arthurtyukayev:
I am fighting a weird bug that I've yet to encounter before. After attempting to adapt my code to the file format from `filesystem.ts`, I am getting the following error in console, it's odd.
image

arthurtyukayev (commented):
Once this is fixed upstream, I can remove this comment.

lishid:
Were you able to nail down what was causing the issue exactly?

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    let results: ImportResult = {
+      total: 0,
+      skipped: [],
+      failed: []
+    };
+
+    // @ts-ignore
+    const attachmentsFolderPath = await this.createFolders(EXPORTED_ASSETS_FOLDER_NAME);
```
lishid, line 35: I think we should just call this "assets", as a subfolder of the user-chosen output folder. Also, what's the need for ts-ignore for?

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,80 @@
...
+    const assetMatcher = /!\[\]\(assets\//g;
+
+    for (let file of files) {
+      await file.readZip(async zip => {
+        for (let zipFileEntry of await zip.getEntries()) {
+          try {
+            if (!zipFileEntry) continue;
+            if (zipFileEntry.filename.match(/\.md|.markdown$/)) {
```
lishid, line 43: I don't think this regex does what you intend.

```suggestion
            let { basename, extension } = parseFilePath(zipFileEntry.filename);
            if (extension  === 'md' || extension  === 'markdown')) {
```

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,80 @@
...
+              const paths = zipFileEntry.filename.replace(`/${parseFilePath(zipFileEntry.filename).basename}`, '').split('/');
+              const mdFilename = paths[paths.length - 1]?.replace('.textbundle', '');
+              let mdContent = await zipFileEntry.getData(new TextWriter());
+              if (mdContent.match(assetMatcher)) {
```
lishid, line 47: Seems like files without assets don't get saved...

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,80 @@
...
+                // Replace asset paths with new asset folder path.
+                mdContent = mdContent.replace(assetMatcher, `![](${attachmentsFolderPath.path}/`);
+                let filePath = normalizePath(mdFilename);
+                await this.saveAsMarkdownFile(folder, filePath, mdContent);
+                results.total++;
+                continue;
+              }
+            }
+
+            if (zipFileEntry.filename.match(/\/assets\//g)) {
```
lishid, line 57: Maybe use `else if` here?

main.ts
```diff
@@ -2,6 +2,7 @@ import { App, Modal, Plugin, Setting } from 'obsidian';
 import { FormatImporter } from './format-importer';
 import { EvernoteEnexImporter } from './formats/evernote-enex';
 import { HtmlImporter } from './formats/html';
+import { Bear2bkImporter } from 'formats/bear-bear2bk';
```
lishid, line 5: Should use relative import here.

main.ts
```diff
@@ -33,6 +34,10 @@ export default class ImporterPlugin extends Plugin {
 				name: 'HTML (.html)',
 				importer: HtmlImporter,
 			},
+      'bear': {
```
lishid, line 37: Seems like there's some formatting differences with the rest of the code.

lishid (commented):
I think it's pretty much ready minus some small adjustments!

formats/bear-bear2bk.ts
arthurtyukayev: Good catch, I missed this. I moved the if-statement close to the correct line.

formats/bear-bear2bk.ts
arthurtyukayev: Done, also added a final else to add any files that aren't assets or markdown files to the skipped arra .

esbuild.config.mjs
```diff
@@ -44,7 +44,7 @@ const context = await esbuild.context({
 
 	// We don't need to include code to create zip files (deflate), only read them (inflate),
 	// so this cuts it out and makes the final bundle smaller.
-	alias: {'@zip.js/zip.js': '@zip.js/zip.js/lib/zip-no-worker-inflate.js'},
+	// alias: {'@zip.js/zip.js': '@zip.js/zip.js/lib/zip-no-worker-inflate.js'},
```
arthurtyukayev, line 47: This seems to be the issue, more specifically it's related to the workers. 

I'm not sure entirely what is causing this issue, but it seems like whenever the code reaches anything related to `zip.js`, it attempts to fetch it via web worker?

lishid: Hmm but that's weird, I'm specifically mapping it to `zip-no-worker-inflate` which shouldn't be using workers at all. Also in my testing I couldn't reproduce the issue; I was able to list files and get their contents from the zip.

arthurtyukayev: Alright, well it doesn't block me for development as I am able to get around it. This ticket should be ready to go.

formats/bear-bear2bk.ts
```diff
@@ -0,0 +1,80 @@
...
+      return;
+    }
+
+    let results: ImportResult = {
+      total: 0,
+      skipped: [],
+      failed: []
+    };
+
+    const attachmentsFolderPath = await this.createFolders(`${(await this.getOutputFolder()).path}/assets`);
```
lishid, line 32:
```suggestion
    const attachmentsFolderPath = await this.createFolders(`${folder.path}/assets`);
```

lishid (approved):
LGTM. I have some minor styling changes I'll just touch-up once this PR merges! Thank you so much for your work ❤️

lishid:
Are you ready to merge to master? I am good to go as soon as the last suggestion is applied.

lishid:
Oh and could you reformat with tabs

arthurtyukayev:
Should be good now.

arthurtyukayev:
Ready when you are for merge.

lishid:
Do you happen to have any sample test files I can use for further development?

PR #33, opened 2023-07-30
Add importing HTML attachments

### LINKED ISSUE #28: HTML attachments - $300
## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $300 USD
**Timeframe:** 7 days

## Requirements

Improve the HTML importer by adding support for attachments

- [ ] Add support for attachments (images, video, audio)
- [ ] Follow the user's setting for `Default location for new attachments`
- [ ] Use Obsidian-style embed links `![[image.png]]`
- [ ] Avoid importing icons/logos that are not part of the content of the page

polyipseity:
Closes #28.

__Requirements__

* [x] Add support for attachments (images, video, audio).
  * Supports file (local) and http(s) links.
* [x] Follow the user's setting for `Default location for new attachments`.
* [x] Use Obsidian-style embed links `![[image.png]]`.
  * This implementation follows the setting `Files & Links > Use [[Wikilinks]]`.
* [x] Avoid importing icons/logos that are not part of the content of the page.
  * Uses image size. User configurable. Default is skip if less than 65 pixels in either dimension.

__Additional__

* Report failed and skipped attachments in the results.
* Attachment size limit. User configurable. Default is ~~10 MB (10e7 bytes)~~ no size limit.
* Append the correct file extension for the detected MIME type if no or wrong extension so that the file can be displayed in Obsidian.
* Add instructions to readme.
* As a side effect, it fixes links and embeds outputted by `htmlToMarkdown` such as `` which does not work in Obsidian.

~~__Dependencies__~~

* ~~`file-type`, `is-svg`: Detect MIME type from file content when detection from `Content-Type` header and file extension fails.~~
* ~~`image-size`: Detect image size to skip icons and logos.~~
* ~~`mime` (`@types/mime`): Interconvert between MIME types and file extensions. Imports from `mime/lite` instead of `mime` to reduce bundle size.~~
* ~~Bundle size: 91.3 kiB to 162 kiB (+70.7 kiB/77.4%)~~

~~Let me know if the bundle size is too large:~~
- ~~We could replace `is-svg` that also checks whether the text is valid XML with a check that simply finds `<svg`, which reduces 26 kiB from the bundle size to 185 kiB.~~
- ~~Or we could drop MIME type detection from file content altogether and remove `file-type` and `is-svg` (which will mean if the type cannot be determined from `Content-Type` or the file extension of the URL, the file is skipped), decreasing the bundle size by 90 kiB to 122 kiB.~~


ericaxu (changes requested):

ericaxu:
Overall it looks okay, although I'm not quite comfortable with the AST parsing approach.

The more intuitive approach would be to leverage Obsidian's own parsing, but I don't know how soon metadataCache would be ready to be consulted after we added a new file, and would that be a big performance hit to the importer. Will let @lishid make that judgement call.

formats/html.ts
```diff
@@ -1,15 +1,64 @@
...
 
 export class HtmlImporter extends FormatImporter {
+	attachments: Record<string, ReturnType<typeof this.downloadAttachment>> = {};
+
+	attachmentSizeLimit: number;
+	minimumImageSize: number;
+
 	init() {
 		this.addFileOrFolderChooserSetting('HTML (.htm .html)', ['htm', 'html']);
+		this.addAttatchmentSizeLimit(10);
```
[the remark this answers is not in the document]
polyipseity: Changed to 0. Multiple filters are combined using `&&`, so the attachment needs to satisfy both the size limit (if set) and the minimum image size (if set) to be included, which should be what you want.

formats/html.ts
```diff
@@ -1,15 +1,64 @@
...
 		this.addOutputLocationSetting('HTML');
 	}
 
-	async import(): Promise<void> {
+	addAttatchmentSizeLimit(defaultInMB: number) {
+		this.attachmentSizeLimit = defaultInMB * 10 ** 6;
+		new Setting(this.modal.contentEl)
+			.setName("Attachment size limit (MB)")
+			.setDesc("Set 0 to disable.")
+			.addText(text => text
```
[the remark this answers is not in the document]
polyipseity: Did that.

But it only checks whether it can be converted using `Number` and doesn't allow you to set further constraints on the input, so most validation code in `onChange` is still there. I think the only benefit would be that mobile users would be shown the special numeric keyboard (if the plugin is ever supported on mobile).

I tried using `inputmode="numeric" pattern="(custom pattern here)"`, but I realized the pattern doesn't constrain the input but only marks the input as invalid.

formats/html.ts
```diff
@@ -1,15 +1,64 @@
...
+					this.attachmentSizeLimit = num * 10 ** 6;
+				}));
+	}
+
+	addMinimumImageSize(defaultInPx: number) {
+		this.minimumImageSize = defaultInPx;
+		new Setting(this.modal.contentEl)
+			.setName("Minimum image size (px)")
+			.setDesc("Set 0 to disable.")
+			.addText(text => text
```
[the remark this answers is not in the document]
polyipseity: Ditto.

polyipseity:
> The more intuitive approach would be to leverage Obsidian's own parsing, but I don't know how soon metadataCache would be ready to be consulted after we added a new file, and would that be a big performance hit to the importer. Will let @lishid make that judgement call.

I would certainly prefer using `metadataCache`, and I think it wouldn't affect the performance very much (unless the initial indexing is ongoing). The problem for me is that this approach cannot be used to fix embed links outputted by `htmlToMarkdown`. For example, something as simple as
```html
<html>
	<body>
		<image src="a b#c.png" />
	</body>
</html>
```
becomes
```md
![](a b#c.png)
```
which would not work on Obsidian due to the space and the hash symbol (needs to be escaped as `%20` and `%23` respectively). The broken embeds are not inside `metadataCache`. Fixing those broken links is important because filenames with spaces are common enough.

(While writing this, I realized that the HTML notes probably contain links to each other, which is not handled by the importer right now. The importer just leaves them as `[](example.html)` instead of `[](example.md)` or `[[example]]`. It should definitely be fixed, but is out of scope for this PR...)

lishid:
I don't love the additional dependencies - would prefer to either come up with alternatives (builtins or simple code we can put right into the repo). I remember yarle also had a mime dependency which is at `/formats/yarle/utils/mime.ts`.

formats/html.ts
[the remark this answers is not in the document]
ericaxu: Yes, please keep all the validation code. In general I think it's good practice to set `type` to `number` if you're only accepting numbers, that's all.

polyipseity:
> I don't love the additional dependencies - would prefer to either come up with alternatives (builtins or simple code we can put right into the repo).

Managed to eliminate all additional dependencies but `file-type`. Would an alternative to `file-type` mean replicating its functionality?

> I remember yarle also had a mime dependency which is at `/formats/yarle/utils/mime.ts`.

Thanks for the hint.

polyipseity:

The code should already be ready.

(I feel like there should be some automation for the CLA thingy.)

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
+		const data = (await fsPromises.readFile(nodeUrl.fileURLToPath(url.href))).buffer;
+		return { mime: await detectMime(url, data), data };
+	}
+
+	async requestHTTP(url: URL) {
+		url = new URL(url.href);
+		let response;
+		try {
+			url.protocol = "https:";
+			response = await requestUrl({
```
lishid, line 224: Can we try to use `fetch` first before falling back to requestUrl? I believe it's much more efficient since it doesn't have to go through a cross-process call through the electron main process.

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
-				let mdContent = htmlToMarkdown(htmlContent);
-				await this.saveAsMarkdownFile(folder, file.basename, mdContent);
-				results.total++;
+				//@ts-ignore
+				const path: string = await this.app.vault.getAvailablePathForAttachments(basename, extension, mdFile);
+				return await this.app.vault.createBinary(path, data);
 			} catch (e) {
-				console.error(e);
-				results.failed.push(file.toString());
+				// retry in case `path` is the same for multiple invocations of `writeAttachment`
```
lishid, line 255: This can be solved by un-parallelizing the calls.

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
+	data: ArrayBufferLike;
+}
+
+function getURLFilename(url: URL) {
+	return pathToFilename(normalizePath(decodeURI(url.pathname)));
+}
+
+async function detectMime(url: URL, data: ArrayBufferLike) {
+	return mime(splitFilename(getURLFilename(url)).extension) ||
+		((await fileType.fileTypeFromBuffer(data))?.mime ??
```
lishid, line 305: Hmm I've been thinking about it - I think I am ok to forego binary based detection. We should prioritize it like this:
- Mime type from the response, if available, and if it converts to an extension correctly
- Extension from the URL if possible
- Skip the file otherwise.

formats/utils/mime.ts
```diff
@@ -996,3 +997,7 @@ export function extension(type: string) {
 
 	return ext || '';
 }
+
+export function mime(extension: string) {
```
lishid, line 1001: I'm seeing that this is only used to convert an extension to a mime and then immediately converted back to an extension. Why not just use the extension?

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
+			results.skipped.push(file.name);
+			return results;
+		}
+		let mdFile: TFile | null = null;
+		try {
+			let mdContent = htmlToMarkdown(await file.readText());
+			const pathURL = nodeUrl.pathToFileURL(file.filepath);
+			mdFile = await this.saveAsMarkdownFile(folder, file.basename, "");
+
+			const ast = [];
```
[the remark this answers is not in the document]
joshuatazrein: I also worked on translating wikilinks in the [Notion importer](https://github.com/joshuatazrein/obsidian-importer/blob/929590bd13f19490c028ef37e0b401f5f0f705ed/formats/notion/convert-to-md.ts#L191).

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
 			return;
 		}
 
-		let results: ImportResult = {
-			total: 0,
+		const results = await Promise.all(files
```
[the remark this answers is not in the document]
polyipseity: I like it to be functional. But anyway, I changed it to your suggestion.

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
+	}
+
+	async processFile(folder: TFolder, file: PickedFile) {
+		const results: ImportResult = {
+			total: 1,
+			failed: [],
 			skipped: [],
-			failed: []
 		};
+		if (!(file instanceof NodePickedFile)) {
```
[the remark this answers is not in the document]
polyipseity: Fixed. Now it will only not import local files on mobile. It seems like you cannot import local files because you cannot read an arbitrary filepath using browser APIs only.

Well, except that you can actually use Capacitor plugins in Obsidian plugins as long as the corresponding native Capacitor plugin is included in the official Obsidian build, like `@capacitor/filesystem`: https://github.com/polyipseity/obsidian-plugin-library/blob/a8caa2e3fb7bc45883c876f86413b1b52fd93c6d/sources/obsidian.ts#L491-L496. So this might be a way to provide attachment support on mobile.

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
...
+				.map(({ link: { path } }) => path));
+
+			mdContent = transformedAST.map(({ text }) => text).join("");
+			await this.app.vault.modify(mdFile, mdContent);
+		} catch (e) {
+			console.error(e);
+			results.failed.push(file.toString());
+			if (mdFile) {
+				try {
+					await this.app.vault.delete(mdFile);
```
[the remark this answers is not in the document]
polyipseity: > Hmm maybe it doesn't make sense to delete the file if any attachment download fails?

This does not happen because there is a `try-catch` block for downloading attachments: https://github.com/obsidianmd/obsidian-importer/pull/33/files/b1c500b454a1dfe1b5f40112f2faff455ac116b1#diff-3817194cf9ed55204dc4518448fc9ad1f788b08fec4b4d3aa16b8a989a28ba86R147-R150

For the updated code, `Promise.allSettled` is used instead, which achieves the above same result: https://github.com/obsidianmd/obsidian-importer/pull/33/files#diff-3817194cf9ed55204dc4518448fc9ad1f788b08fec4b4d3aa16b8a989a28ba86R97

> Perhaps we should not save to file immediately after converting to md, and instead only save when everything has been downloaded? I know we'll need a path for link generation, but since it's only used for relative link resolution, I think we can generate a fake one of where we think the file will end up at.

The undocumented method `getAvailablePathForAttachments` needs a `TFile` as the third argument: https://github.com/obsidianmd/obsidian-importer/pull/33/files#diff-3817194cf9ed55204dc4518448fc9ad1f788b08fec4b4d3aa16b8a989a28ba86R241-R242

formats/html.ts
polyipseity: Done.

formats/utils/mime.ts
polyipseity: It is removed from the PR now because it is no longer needed.

formats/html.ts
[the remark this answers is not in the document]
polyipseity: > Instead of rolling our own parser here, perhaps it would be easier to work on the original HTML? What I suggest doing is to use DOMParser to parse the HTML into a DOM tree first, then querySelector on the `` elements and working off them. [(See how this is implemented in the Notion importer)](https://github.com/obsidianmd/obsidian-importer/pull/34/files#diff-dcb7a54e5cb2f4c3a888c6eb457ebfc648ec9534b52beeb4202fea10e9514e30R85)

Done. ``, `<audio>`, and `<video>` elements are now queried.

> It unfortunately won't help us generate wikilinks instead of markdown links, but one possibility is to go back to match the URL in the final markdown text and attempt to regex the before/after from the position where the URL was matched.

The Markdown parsing is now done using the builtin parser via `MetadataCache` (by first fixing the embed links outputted by `htmlToMarkdown`), so regex is not even needed to find the embed links to replace. The wikilink generation is already handled by `generateMarkdownLink`.

formats/html.ts
polyipseity: Did that by creating a small utility `PromiseExecutor`.

formats/html.ts
polyipseity: Okay, the binary-based detection is gone now.

Actually, with this idea and the idea above to use `DOMParser` to read the attachments, the entire MIME detection can be eliminated. Basically the steps are:
1. Use the `DOMParser` to read the attachments. Each attachment is labelled whether it is from a `<audio>`, ``, or `<video>` HTML element.
2. (some unrelated intermediate steps later)
3. Check whether the downloaded file has a content type or file extension. Lookup or keep the extension if so. Otherwise, `.noext.mp3` (or any other audio file extension) is appended if the attachment is from an `<audio>`, `.noext.bmp` (or any other image file extension) for ``, and `.noext.mp4` for `<video>` (or any other video file extension). It doesn't matter whether the file format actually matches the extension as the Markdown renderer only uses the extension to determine whether to use a `<audio>`, ``, `<video>`, or no embed. The browser will figure out the actual file format by itself. This way, if we cannot detect the extension, the file can still be displayed. `.noext` here is used to label that they originally have no extensions to the user.

formats/html.ts
```diff
@@ -1,44 +1,339 @@
...
+	}
+}
+
+interface TypedResponse {
+	type: "audio" | "img" | "video";
+	data: ArrayBufferLike;
+	extension: string;
+}
+
+function escapeRegExp(str: string) {
```
polyipseity, line 293: There is also `escapeStringRegexp` inside `yarle`, but using that causes "Invalid escape" error when creating a new `RegExp`. I don't want to break the yarle importer so I didn't fix and use that function.

formats/html.ts
```diff
@@ -21,24 +73,340 @@ export class HtmlImporter extends FormatImporter {
 			return;
 		}
 
-		let results: ImportResult = {
-			total: 0,
+		const results = await Promise.all(files
+			.map(file => this.processFile(folder, file)));
```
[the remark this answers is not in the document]
polyipseity: > * Memory limits. If we work in parallel, this could lead to out-of-memory cases, especially for concurrently downloading multiple attachments. This is a much bigger problem on mobile devices, if/when we want to support them in the future.

For memory limits, I believe this can be mitigated by limiting the number of processing files and attachments that can run at once.

> * One-time use. Because the use case of the importer isn't a day-to-day thing, I think most people will be ok to wait a few more seconds for their import. I'd want to make sure we don't do anything obviously inefficient, but I think maximum parallelization isn't something we need to achieve.

Can't argue against that. (Though I imagine a few wanting to use multiple tools, making frequent use of it...)

> For this specific importer there's also a potential issue of rate-limiting since we are downloading some of these files from the internet.

Same as memory limits. In fact, I have already limited the max number of concurrent http connections to 5: https://github.com/obsidianmd/obsidian-importer/blob/453d8371ddff574e0c7b85f6f2fff66bcdc2a7d7/formats/html.ts#L12-L13

---

Here's a patch that adds a "parallel" toggle to the settings and logs the time elapsed for comparing concurrent and sequential processing:
- [parallel-toggle.patch](https://github.com/obsidianmd/obsidian-importer/files/12248430/parallel-toggle.patch) (base: 453d8371ddff574e0c7b85f6f2fff66bcdc2a7d7)

Testing indicates that concurrent processing speeds up reading local files and http requests, and doesn't do much when reading and writing files in the vault (probably because `Vault` operations are forced to be serial). So it depends on the amount of local attachments and external attachments. Also, it looks like Turndown, the tool for turning HTML into Markdown, takes quite some time (0.1~1 s per file), so I think the best way would be to sequentially process Markdown files, and download attachments concurrently.

formats/utils/mime.ts
polyipseity: Also, speaking of `mime.ts`, the data seems inadequate. For example, `image/jpeg` only maps to `.jpeg` but not `.jpg`.

ericaxu:
Btw we increased the bounty amount. Please see kepano's comment under the issue. Thanks for putting in the good work!

formats/html.ts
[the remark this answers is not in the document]
lishid: Ah your are correct for both!

lishid:
I really like the PromiseExecutor code, it's pretty elegant. However I think it does not address the memory concern since it only limits concurrent execution of that specific phase. For example, for the HTTP downloads, so it's possible that we concurrently download 4 files at a time, but they are all held in memory until the subsequent single-threaded save code save each of them. Similarly we could start processing 1000 `html` files, and all of them will start the conversion concurrently (and hold all of the raw HTML + converted markdown strings in memory), then they will go through the bottleneck of HTTP downloads all while holding the in-progress variables in memory.

lishid:
I think in order for PromiseExecutor to actually work for memory limits, it needs to be started from a single thread, or another limited pool (instead of being fed from a fully-concurrent process). Here for example, the HTTP executor is fed from `processFile` which is fully concurrent.

formats/html.ts
lishid: That's amazing. One suggestion I had is to default to png or jpg instead of bmp, since those are much more commonly used these days.

polyipseity: > That's amazing. One suggestion I had is to default to png or jpg instead of bmp, since those are much more commonly used these days.

Replaced wih `png` now.

polyipseity:
> I really like the PromiseExecutor code, it's pretty elegant. However I think it does not address the memory concern since it only limits concurrent execution of that specific phase. For example, for the HTTP downloads, so it's possible that we concurrently download 4 files at a time, but they are all held in memory until the subsequent single-threaded save code save each of them. Similarly we could start processing 1000 `html` files, and all of them will start the conversion concurrently (and hold all of the raw HTML + converted markdown strings in memory), then they will go through the bottleneck of HTTP downloads all while holding the in-progress variables in memory.

> I think in order for PromiseExecutor to actually work for memory limits, it needs to be started from a single thread, or another limited pool (instead of being fed from a fully-concurrent process). Here for example, the HTTP executor is fed from processFile which is fully concurrent.

I am aware of that. I had not made the Markdown processing limited by replacing `Promise.all` with `new PromiseExecutor` or a simple `for`-loop. It should be now fixed: https://github.com/obsidianmd/obsidian-importer/blob/8ffd780bbd0800e80a85cf0afca4fe159ff80c0c/formats/html.ts#L82-L86

polyipseity:
Also, thanks for the new shiny utilities.

formats/html.ts
```diff
@@ -1,44 +1,358 @@
...
+							if (file.path === mdFile.path) {
+								try {
+									resolve(cache);
+								} finally {
+									this.app.metadataCache.offref(ref);
+								}
+							}
+						});
+					});
+					await this.app.vault.process(mdFile, data => `${data}${mdContent}`);
```
lishid, line 158: Since we saved the file as empty before, we should just call modify here instead of process.

formats/html.ts
```diff
@@ -1,44 +1,358 @@
...
+					);
+					const attachments2 = Object.fromEntries(Object.entries(attachments).map(([key, value]) => [decodeURIComponent(key), value]));
+
+					const cache = new Promise<CachedMetadata>(resolve => {
+						const ref = this.app.metadataCache.on("changed", (file, _1, cache) => {
+							if (file.path === mdFile.path) {
+								try {
+									resolve(cache);
+								} finally {
+									this.app.metadataCache.offref(ref);
```
lishid, line 153: Suggest calling offref here first before resolving.

formats/html.ts
```diff
@@ -1,44 +1,358 @@
...
+							}
+							return [original, this.app.fileManager.generateMarkdownLink(attachment, mdFile.path, "", displayText)] as const;
+						})
+						.filter(entry => entry));
+					if (Object.keys(embeds).length > 0) {
+						const embedOriginals = alternativeRegExp(Object.keys(embeds));
+						await this.app.vault.process(mdFile, data => data.replace(embedOriginals, orig => embeds[orig]));
+					}
+				} else {
+					mdContent = mdContent.replace(alternativeRegExp(links), sstr => sstr.replace(/ /gu, "%20"));
```
lishid, line 174: Shouldn't we do this directly to the DOM node's href?

lishid (commented):
Ok probably one of the last rounds of review, this is pretty close to ready for merging.

I honestly admire the amount of effort you're willing to go with concurrency optimizations, but ultimately since maintenance falls under responsibility, I'd prefer if we keep things simple and easy to reason about. Two things I'd prefer if you are willing to refactor the code for:

- Sequential processing.
- Reduce the use of functional processing of arrays and objects.

By functional I am talking about the extensive use of `.map` and `.filter`. For example, instead of

```ts
const downloads = await Promise.allSettled(
	Array.from(dom.querySelectorAll<HTMLAudioElement | HTMLImageElement | HTMLVideoElement>("audio, img, video"))
	.map(async element => {
		/*...*/
	});
/*...*/
result.failed = result.failed.concat(downloads
	.filter((dl): dl is typeof dl & { status: "rejected" } => dl.status === "rejected")
	.map(({ reason }) => decodeURIComponent(reason)));
result.skipped = result.skipped.concat(downloads
	.filter((dl): dl is typeof dl & { status: "fulfilled" } => dl.status === "fulfilled" && !dl.value[1])
	.map(({ value: [src] }) => decodeURIComponent(src)));
```

I'd prefer a more straightforward
```ts
for(let element of dom.querySelectorAll<HTMLAudioElement | HTMLImageElement | HTMLVideoElement>("audio, img, video")) {
	try {
		/*...*/
		if (skipped) {
			result.skipped.push(decodeURIComponent(element.src));
		}
	} catch(e) {
		result.failed.push(e);
	}
	...
```

formats/html.ts
polyipseity: Done. I was thinking about plugins that write text to newly created files.

formats/html.ts
polyipseity: Done.

formats/html.ts
polyipseity: Done. Brilliant idea!

polyipseity:
The ugly `filter`s and `map`s are gone now. The code is sequential but I have left it at a state that should be easy to convert to concurrent should the need arises. For example,
```TypeScript
const attachments = [];
for (const element of Array.from(dom.querySelectorAll<HTMLAudioElement | HTMLImageElement | HTMLVideoElement>("audio, img, video"))) {
	const ret = await this.processAttachment(result, mdFile, element, base);
	if (ret) {
		attachments.push(ret);
	}
}
```
can easier become
```TypeScript
const attachments = (await Promise.all(
	Array.from(dom.querySelectorAll<HTMLAudioElement | HTMLImageElement | HTMLVideoElement>("audio, img, video"))
		.map(element => this.processAttachment(result, mdFile, element, base))
))
	.filter(entry => entry);
```
without any changes to other parts of the code (and without too many ugly `filter`s and `map`s).

lishid (approved):
Looks good to me. I am ready to merge as soon as the conflict is resolved.

lishid:
Do you happen to have some sample files for testing?

polyipseity:
Well, the testing files I have been using are scrapped from random
websites, so I can't commit them here legally. But I could make some
synthetic HTML files like the test cases for Google Keep.

lishid:
Ok cool. Could you fix the merge conflict? After that we are good to merge.

polyipseity:
> Ok cool. Could you fix the merge conflict? After that we are good to merge.

Resolved the conflict. Just had some further bugs to squash after creating those new test cases.

lishid:
Hmmmm if we're committing the test files into the repo then we should pick something smaller for the audio and video files - also preferably mp3 and mp4.

polyipseity:
I have reencoded them to mp3 and mp4 now. They are also about 66% smaller (about 50 kB per file) now. I would have preferrred to use symlinks to give each attachments 3 different names instead of duplicating the attachment 3 times, but Windows by default do not support them.

lishid:
That's fine. Thank you for the contribution!

We will be in touch about the bounty very soon!

ericaxu:
Congrats on completing the bounty! Will DM for bounty payout.

Issue #14, opened 2023-07-14
Import from Notion - $1,000


kepano:
Please upvote this issue if you would like to see import from Notion.

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $1,000 USD
**Timeframe:** 10 days

## Requirements

Must support the following:

- [ ] Uses either Notion's HTML or MD export (decide which one is best)
- [ ] Convert page title to file name and remove it from the top of the file
- [ ] Maintain folder structure
- [ ] Proper Obsidian Markdown formatting
- [ ] Parse internal links
- [ ] Parse page properties, including internal links must be imported into YAML frontmatter using the correct field types
  - [ ] Text
  - [ ] Date
  - [ ] Time
  - [ ] Number
  - [ ] List
  - [ ] Checkbox

Internal link format (added in Obsidian 1.4+)

```yaml
---
link: "[[Link]]"
linklist:
  - "[[Link]]"
  - "[[Link2]]"
---
```

## Current solutions

- See [Obsidian forum](https://forum.obsidian.md/t/meta-migration-workflows/15252)

Fevol:
I'm interested in implementing the Notion importer. I have a pretty decent understanding of how Obsidian development works (developer of the [Translate](https://www.github.com/Fevol/obsidian-translate) (_released_) and [Commentator](https://www.github.com/Fevol/obsidian-criticmarkup) (**WIP**) plugins).  

As a caveat, I do not have a lot of experience with Notion itself. I am looking into it currently, but if anyone other developer has more experience with the program and format, I'll apply for a different bounty.

joshuatazrein:
Here is a link to a repository implementing a Notion importer which I have used to import my own notes. It reads a folder of .md exports and converts them to internal links, also converting formatting: https://github.com/joshuatazrein/obsidian-notion-import

kepano:
> Here is a link to a repository implementing a Notion importer which I have used to import my own notes. It reads a folder of .md exports and converts them to internal links, also converting formatting: https://github.com/joshuatazrein/obsidian-notion-import

Is the repo private?

joshuatazrein:
> > Here is a link to a repository implementing a Notion importer which I have used to import my own notes. It reads a folder of .md exports and converts them to internal links, also converting formatting: https://github.com/joshuatazrein/obsidian-notion-import
> 
[...]

PR #34, opened 2023-07-31
Add Notion Importer

joshuatazrein:
The Notion Importer!

Process:
- Reads all files from the selected `.zips` and creates two objects in memory, `idsToFileInfo` and `pathsToAttachmentInfo`.
- Then reconciles duplicate filenames (allowed in Notion but not Obsidian) by prefixing numbers onto duplicates
- Writes files to the target folder, setting appropriate metadata from file properties
    - Converts .html to .md, translating links and relations to the correct Obsidian internal links
    - Also moves all attachments into the user's Attachments directory or nested folders, depending on the setting

Notes:
- Any relations/links to Notion files which aren't included in the import become Obsidian links to files that don't exist yet. This enables gradually importing Notion without losing internal links.


ericaxu:
This is for bounty #14, we still have a full week left. It's ready for testing for everyone interested.

@joshuatazrein : I see you've resolving some issues with helpful testers on Discord, do you wanna keep doing that before we review the code?

joshuatazrein:
I've resolved the issues so far mention on Discord, so I believe it's ready for testing now!

joshuatazrein:
Ok, some issues which have arisen just now:
- [ ] I'm stumped on this, but somehow [some notes are not being imported](https://discord.com/channels/686053708261228577/1133074995271188520/1135714952531951706) which are found neither in "failed" or "skipped" files. I'm not sure which notes these are or what is failing. Could be a filename duplicate?
- [ ] Need to handle behavior for these attachment folder settings:
    - [ ] Same folder as current file
    - [ ] In subfolder under current folder

lishid (commented):
When iterating objects, I'd prefer to use simple for-in with hasOwnProperty, or a for-of with Object.keys, rather than using `Object.entries()` 🙏 


One thing to note is that Obsidian's `htmlToMarkdown` will actually also accept a DOM node. Instead of manually matching HTML strings, consider using `DOMParser` to get a DOM tree to work with.

formats/notion/copy-files.ts
```diff
@@ -0,0 +1,119 @@
...
+	idsToFileInfo: Record<string, NotionFileInfo>;
+	pathsToAttachmentInfo: Record<string, NotionAttachmentInfo>;
+	attachmentFolderPath: string;
+	targetFolderPath: string;
+	app: App;
+	results: ImportResult;
+}) {
+	const createdFolders = new Set<string>();
+
+	await Promise.all(
```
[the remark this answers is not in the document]
joshuatazrein: Wow, that saves so much headache. I will switch to sequential processing.

joshuatazrein:
> When iterating objects, I'd prefer to use simple for-in with hasOwnProperty, or a for-of with Object.keys, rather than using `Object.entries()` 🙏
> 
> One thing to note is that Obsidian's `htmlToMarkdown` will actually also accept a DOM node. Instead of manually matching HTML strings, consider using `DOMParser` to get a DOM tree to work with.

Very good to know. The `DOMParser` will fix a lot of the finnicky RegEx. I'll switch to that as well as the for-of with Object.keys.

formats/notion.ts
```diff
@@ -0,0 +1,109 @@
...
+		});
+
+		results.total =
+			filePaths.length + Object.keys(pathsToAttachmentInfo).length;
+
+		const appSettings = await app.vault.adapter.read(
+			normalizePath(`${app.vault.configDir}/app.json`)
+		);
+		const parsedSettings = JSON.parse(appSettings ?? '{}');
+		const attachmentFolderPath = parsedSettings.attachmentFolderPath ?? '';
```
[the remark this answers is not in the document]
joshuatazrein: Great, updated!

formats/notion/copy-files.ts
```diff
@@ -0,0 +1,119 @@
+import { readFile } from 'fs/promises';
+import { ImportResult } from 'main';
+import moment from 'moment';
```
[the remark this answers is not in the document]
joshuatazrein: I'm confused—does `obsidian` export `readFile`? I'm not sure where that is, it's not showing up in VSCode.

formats/notion/copy-files.ts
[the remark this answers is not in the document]
joshuatazrein: Actually after switching to sequential processing, the import process takes much longer. Even if the internal fs serializes I think keeping parallel processing speeds it up significantly. What do you think?

ericaxu:
I labelled it "Changes requested" for now. You can re-label it when you're done addressing the comments, thanks!

formats/notion/copy-files.ts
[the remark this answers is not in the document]
joshuatazrein: > Hmm interesting. How much of a difference is it? Do you have any intuition as to why that is? Could you run a performance profile for before and after?
> 
> I'm currently thinking it is possibly due to the parallelization of the `fs.readFile` call for the attachments.
> 
> If that's the case we can look into parallelization again, though I'd need to think about how to keep the code comprehensible 🤣

Let me do that. For importing 1000 notes it used to take 2 seconds now it takes more than 10. I think `readFile` is a big part, but am not sure. I'll profile tomorrow

joshuatazrein:
@ericaxu I have finished incorporating changes but can't change tags on the pull request for some reason.

formats/notion/copy-files.ts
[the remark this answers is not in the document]
joshuatazrein: Here's the profile for sequential processing...`json` isn't supported so I zipped it. As a side note it could have been some plugins I had enabled that responded to `metadataCache` events adding a lot of the time. Still though something to think about.

[Profile-20230802T093420.json.zip](https://github.com/obsidianmd/obsidian-importer/files/12241925/Profile-20230802T093420.json.zip)

ericaxu:
> @ericaxu I have finished incorporating changes but can't change tags on the pull request for some reason. @lishid I have incorporated the `DOMParser` node which simplified things quite significantly.

Will change it for you!

formats/notion/copy-files.ts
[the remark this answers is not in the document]
lishid: Thank you! I've gone back to putting more thought into this problem and I think I am inclined to forego parallelization for the following reasons:

- Memory limits. If we work in parallel, this could lead to out-of-memory cases, especially for concurrently downloading multiple attachments. This is a much bigger problem on mobile devices, if/when we want to support them in the future.
- One-time use. Because the use case of the importer isn't a day-to-day thing, I think most people will be ok to wait a few more seconds for their import. I'd want to make sure we don't do anything obviously inefficient, but I think maximum parallelization isn't something we need to achieve.

What do you think?

lishid:
Hmm might need a merge with master to get the latest change in. In general I would prefer not to perform formatting changes for files outside of/unrelated to this PR to avoid conflicts ❤️.

The new master changes lays the groundwork for mobile support soon. I would also prefer to see us rely less on NodeJS APIs here if possible, though selecting folders don't work on mobile yet due to some technical limitations, so it might end up having to read everything from a zip.

formats/notion/convert-to-md.ts
```diff
@@ -19,226 +23,226 @@ export function convertNotesToMd({
...
 }
 
-const replaceNestedFormatting = (body: string) => {
-	body = body
-		.replace(/<strong>(<strong>)+/g, '<strong>')
-		.replace(/<\/strong>(<\/strong>)+/g, '</strong>')
-		.replace(/<em>(<em>)+/g, '<em>')
-		.replace(/<\/em>(<\/em>)+/g, '</em>');
+const replaceNestedTags = (body: HTMLElement, tag: 'strong' | 'em') => {
+	body.querySelectorAll(tag).forEach((el) => {
```
lishid, line 58: Hmm wouldn't querySelectorAll give you both the parent and children elements? Not sure if there's a better way to de-duplicate the hierarchy.

formats/notion/convert-to-md.ts
```diff
@@ -19,226 +23,226 @@ export function convertNotesToMd({
...
+		const nestedEls = el.querySelectorAll(tag);
+		if (nestedEls.length === 0) return;
+		const textContent = el.textContent;
```
lishid, line 61: How about hoisting the children (moving all children elements of the inner element up) instead of flattening it to `textContent`? It would help prevent destroying other formatting or non-text element like images and links.

formats/notion/convert-to-md.ts
```diff
@@ -19,226 +23,226 @@ export function convertNotesToMd({
...
-			body = body.replace(
-				italic[0],
-				italics.map((italic) => `<em>${italic}</em>`).join('\n')
-			);
-		}
-	}
+const splitBrsInFormatting = (body: HTMLDivElement, tag: 'strong' | 'em') => {
+	body.querySelectorAll(tag).forEach((el) => {
+		// TODO: check to see if newlines are stripped in textContent
+		if (!/<br \/>/g.test(el.textContent)) return;
```
lishid, line 70: Any chance we can do this by manipulating/splicing DOM elements instead of relying on textContent and innerHTML?

lishid: Also this doesn't work with more than one level of nesting right?

formats/notion/copy-files.ts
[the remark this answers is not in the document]
joshuatazrein: I agree. Given that the importer already needs to load a lot of data in memory it makes sense not to read/write files in parallel.

joshuatazrein:
> Hmm might need a merge with master to get the latest change in. In general I would prefer not to perform formatting changes for files outside of/unrelated to this PR to avoid conflicts ❤️.
> 
> The new master changes lays the groundwork for mobile support soon. I would also prefer to see us rely less on NodeJS APIs here if possible, though selecting folders don't work on mobile yet due to some technical limitations, so it might end up having to read everything from a zip.

Sorry, that's Prettier autoformatting. I can unstage that

formats/notion/convert-to-md.ts
joshuatazrein: Skipping any tags which have the same tag as a parent.

formats/notion/convert-to-md.ts
joshuatazrein: Honestly that seems way too complicated for me to do concisely. It worked perfectly with Regex, so I'm just going to do this transformation after all the DOM mutations on the HTML text.

formats/notion/convert-to-md.ts
joshuatazrein: Seems much better. I'll do that.

joshuatazrein:
Merged with master as well as updated @lishid's comments.

ericaxu:

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+}
+
+const replaceNestedTags = (body: HTMLElement, tag: 'strong' | 'em') => {
+	body.querySelectorAll(tag).forEach((el) => {
+		if (el.parentElement.tagName === tag.toUpperCase()) return;
+		const nestedEls = el.querySelectorAll(tag);
+		if (nestedEls.length === 0) return;
+		for (let i = 0; i < nestedEls.length; i++) {
```
lishid, line 65: Hmm I'm not sure if this does what we think it's trying to do - it might be doing something unexpected, for example, if the nesting goes more than one level deep (i.e. it skips levels down). Seems like this will hoist the nested children all the way up to the root level tag. For example if we had
```html
<b>
<i><b>test</b></i>
</b>
```
this code will move `test` directly into the top level `<b>` .

Perhaps what should happen here is that the childNodes get moved to a DocumentFragment, and we use `nestedEls[i].replaceWith(fragment)`.

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,257 @@
...
+						  attachmentInfo.nameWithExtension +
+						  '|' +
+						  attachmentInfo.nameWithExtension
+						: attachmentInfo.nameWithExtension
+				}]]`;
+				break;
+		}
+
+		obsidianLink.setText(linkContent);
+		link.a.insertAdjacentElement('afterend', obsidianLink);
```
lishid, line 254: Easier to just use `replaceWith`.

formats/notion.ts
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		const folderPathsReplacement = new RegExp(
+			`^(${folderPaths
+				.map((folderPath) => escapeRegex('/' + folderPath))
+				.join('|')})`
+		);
+
+		let adapter = app.vault.adapter;
+		if (!(adapter instanceof FileSystemAdapter)) return;
```
lishid, line 40: Any reason why this is needed? I get that we're using readPath, but I think ultimately I'd want to remove the hard requirement on NodeJS.

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,257 @@
...
+	getParentFolder,
+	matchFilename,
+} from '../../util';
+import {
+	assembleParentIds,
+	getNotionId,
+	stripNotionId,
+	parseDate,
+} from './notion-utils';
+import moment from 'moment';
```
lishid, line 14: Remember to use `obsidian` here

lishid (commented):
I'll give this another round tomorrow hopefully - didn't have enough time to go through the whole thing today unfortunately.

formats/notion/convert-to-md.ts
joshuatazrein: that sounds good!

joshuatazrein:
Uh so after the previous updates it seems like `folderPaths` has been removed from the `FormatImporter` class. This is something I need for my importer, can you please add it back in? Without knowing which folders the user picked it's impossible to normalize the paths. Also there doesn't seem to be a `folderChooserSetting` anymore and my importer only works with choosing folders. Since I can't test without these can you fix ASAP?

lishid:
Hmmm I see what you mean - Does notion expert into a zip the user has to manually unzip? I think we might be able to let the user keep using the zip format instead - I am working in parallel to incorporate a shared zip utility because it seems like multiple exports uses zips.

In the meantime feel free to continue working without a rebase, sorry for the oversight!

formats/notion.ts
joshuatazrein: That was vestigial code from when I was starting out. I've removed it.

joshuatazrein:
> Hmmm I see what you mean - Does notion expert into a zip the user has to manually unzip? I think we might be able to let the user keep using the zip format instead - I am working in parallel to incorporate a shared zip utility because it seems like multiple exports uses zips.
> 
> In the meantime feel free to continue working without a rebase, sorry for the oversight!

Yes, the user has to manually unzip. It would be a lot less confusing to let them import from just the zip.

lishid:
I have just implemented the zip reader - let me know if it works for you. Here's how to use it:

```ts
await file.readZip(async zip => {
	let entries = await zip.getEntries();
	for (let zipFile of entries) {
		// Metadata
		console.log(zipFile.filename, zipFile.uncompressedSize);
		// Contents as string
		console.log(await zipFile.getData(new TextWriter()));
		// Contents as ArrayBuffer
		console.log(await (await zipFile.getData(new BlobWriter())).arrayBuffer());
	}
});
```

joshuatazrein:
> I have just implemented the zip reader - let me know if it works for you. Here's how to use it:
> 
> 
> 
> ```ts
> 
> await file.readZip(async zip => {
> 
> 	let entries = await zip.getEntries();
> 
> 	for (let zipFile of entries) {
> 
> 		// Metadata
> 
> 		console.log(zipFile.filename, zipFile.uncompressedSize);
> 
> 		// Contents as string
> 
> 		console.log(await zipFile.getData(new TextWriter()));
> 
> 		// Contents as ArrayBuffer
> 
> 		console.log(await (await zipFile.getData(new BlobWriter())).arrayBuffer());
> 
> 	}
> 
> });
> 
> ```

This is great, solves all the previous issues quite elegantly! Thanks.

joshuatazrein:
> I have just implemented the zip reader - let me know if it works for you. Here's how to use it:
> 
> ```ts
> await file.readZip(async zip => {
> 	let entries = await zip.getEntries();
> 	for (let zipFile of entries) {
> 		// Metadata
> 		console.log(zipFile.filename, zipFile.uncompressedSize);
> 		// Contents as string
> 		console.log(await zipFile.getData(new TextWriter()));
> 		// Contents as ArrayBuffer
> 		console.log(await (await zipFile.getData(new BlobWriter())).arrayBuffer());
> 	}
> });
> ```

I'm not sure where the `TextWriter` class is; I'm not finding it. Where do I import it from?

lishid:
You get them from `'@zip.js/zip.js'`

joshuatazrein:
I've run into a problem. Here's the code:

```
         for (let zip of files) {
		await zip.readZip(async (zip) => {
			const entries = await zip.getEntries();

			for (let file of entries.filter((file) =>
				file.filename.endsWith('.html')
			)) {
				console.log('getting text', file.filename);

				const text = await file.getData(new TextWriter('utf-8'));

				console.log('got text', text);
```

The `TextWriter` seems to throw an error: 
Screenshot 2023-08-04 at 12 11 34 AM

lishid:
Oh huh, that seems to be the same problem as the other PR (https://github.com/obsidianmd/obsidian-importer/pull/32#discussion_r1283994590)

I wonder why I can't repro this issue.

joshuatazrein:
> 

Yeah it's very strange. Would any extra info help from my end?

arthurtyukayev:
As @lishid has mentioned, I got the same problem, @joshuatazrein if it's blocking you, comment out the alias for `@zip.js` in the `esbuild.config.mjs` file.

joshuatazrein:
Thanks so much, working now!

lishid:
I've reproduced the issue and committed a fix on master, let me know if that solves the problem!

joshuatazrein:
The `.zip` imports are now working, I think this is ready to merge!

lishid:
Oops, seems like the merge has duplicated all the files.

lishid:
I will do one last pass over this PR once it's fixed. Do you happen to have a test/sample file that you can share?

joshuatazrein:
Fixed! Here is my test folder (.zip of multiple .zips)
[zips.zip](https://github.com/obsidianmd/obsidian-importer/files/12264254/zips.zip)

lishid:
Sorry for the mess of the merge with master, seems like it's still reverting quite some changes 😬

joshuatazrein:
Is it OK now? Seems fine on my end.

lishid:
Hmm it seems like the conflict was resolved, but your PR now undoes the last few commits on master. For example see here: https://github.com/obsidianmd/obsidian-importer/pull/34/files#diff-3817194cf9ed55204dc4518448fc9ad1f788b08fec4b4d3aa16b8a989a28ba86

lishid:
Cool that fixed it up.

joshuatazrein:
Ok I've resolved all conflicts. Good to go?

lishid (commented):
So I'm going over the main logic of the parser and it seems that it is reading the contents of the entire zip, and reading every single attachment into memory before parsing and processing them. I'm not sure this works well for large exports or low-ram devices like on mobile.

I am not seeing any cross-file dependencies that would require reading all files before we can process them - seems like the deduplication process only works on file names, but let me know if I'm misunderstanding any of that. Seems like we are getting an ID from an <article>, I'm wondering if we can get it from the filename instead.

I would prefer if the code would do something more like:
- Read the list of files (but not the contents just yet)
- Make decisions about duplicates, etc
- Parse each note, process it, then save it to disk before moving onto the next
- Read each attachment and save it to disk

Doing so ensures that the memory used for previous files can get garbage collected, so that we aren't holding references to the contents every single file at any point in time.

lishid:
The other thing that would help is if some comments and examples can be provided for specific parts of the conversion, especially when a regex is involved and the conversion doesn't seem obvious ([example](https://github.com/obsidianmd/obsidian-importer/pull/34/files#diff-44efed1e3d651f9c32b180536bfe44e073c6f43211824afb0f0f9a97a0beaa5bR63)).

joshuatazrein:
> So I'm going over the main logic of the parser and it seems that it is reading the contents of the entire zip, and reading every single attachment into memory before parsing and processing them. I'm not sure this works well for large exports or low-ram devices like on mobile.
> 
> I am not seeing any cross-file dependencies that would require reading all files before we can process them - seems like the deduplication process only works on file names, but let me know if I'm misunderstanding any of that. Seems like we are getting an ID from an `<article>`, I'm wondering if we can get it from the filename instead.
> 
> I would prefer if the code would do something more like:
> 
> * Read the list of files (but not the contents just yet)
> * Make decisions about duplicates, etc
> * Parse each note, process it, then save it to disk before moving onto the next
> * Read each attachment and save it to disk
> 
> Doing so ensures that the memory used for previous files can get garbage collected, so that we aren't holding references to the contents every single file at any point in time.

Yeah, I switched to reading the attachments first today, to only read `.zip` files once. I can wait to load them until the end when I'll just re-read the `.zip`s at that point.

util.ts
```diff
@@ -1,18 +1,99 @@
...
+	const tagExp = /#[a-z0-9\-]+/gi;
+
+	if (!tagExp.test(body)) return body;
+	const lines = body.split('\n');
+	for (let i = 0; i < lines.length; i++) {
+		const hashtags = lines[i].match(tagExp);
+		if (!hashtags) continue;
+		let newLine = lines[i];
+		for (let hashtag of hashtags) {
+			const hashtagInLink = new RegExp(
```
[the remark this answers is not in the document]
joshuatazrein: Added: `// skipping any internal links [[ # ]], URLS [ # ]() or []( # ), or already escaped hashtags \#, replace all tag-like things #<word> in the document with \#<word>. Useful for programs (like Notion) that don't support #<word> tags.`

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,316 @@
...
+	});
+
+	return markdownBody;
+};
+
+const formatDatabases = (body: HTMLElement) => {
+	// Notion includes user SVGs which aren't relevant to Markdown, so change them to pure text. Safe because it's a part of Notion UI and not nested.
+	const users = body.querySelectorAll('span[class=user]');
+	users.forEach((user) => {
+		user.innerHTML = user.textContent;
```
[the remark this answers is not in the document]
joshuatazrein: I want to strip all children which aren't plaintext in the `user` object, which Notion inserts `svg`s and `img`s into for their UI. I don't know if `setText` will replace these non-text tags.

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,316 @@
...
+		const lastChild = select.parentElement.lastElementChild;
+		if (lastChild === select) return;
+		select.setText(select.textContent + ', ');
+	});
+
+	const linkValues = body.querySelectorAll(
+		'a[href]'
+	) as NodeListOf<HTMLAnchorElement>;
+	linkValues.forEach((a) => {
+		if (a.href.startsWith('app://obsidian.md')) {
```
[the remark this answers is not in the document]
joshuatazrein: Added: `// Any <a> with an email, phone number, or non-URL value registers as an internal link in Obsidian. [email@gmail.com](email@gmail.com) will create a new note when clicked. This strips these erroneous links.`

joshuatazrein:
I've changed the process to optimize memory:
1. Read `.zips` and store filenames (still parsing HTML, but not saving it)
2. Resolve duplicates within `idsToFileInfo`
3. Read `.zips` again and convert each `.html` file to `.md`, then immediately write to disc.

joshuatazrein:
I think all changes have been addressed. Is there anything else I should look at before we merge?

formats/notion/convert-to-md.ts
```diff
@@ -0,0 +1,419 @@
...
+		if (lastChild === select) return;
+		select.setText(select.textContent + ', ');
+	});
+
+	const linkValues = body.querySelectorAll(
+		'a[href]'
+	) as NodeListOf<HTMLAnchorElement>;
+	linkValues.forEach((a) => {
+		// Any <a> with an email, phone number, or non-URL value registers as an internal link in Obsidian. [email@gmail.com](email@gmail.com) will create a new note when clicked. This strips these erroneous links.
+		if (a.href.startsWith('app://obsidian.md')) {
```
lishid, line 273: Actually this doesn't work on mobile, because the page isn't hosted on `app://obsidian.md`. What you are looking for are ones where `.getAttribute('href')` (be aware it's different from `.href`) does not contain the protocol indicator character `:`.

lishid (commented):
Need to fix the following:
- Undo changes to `/formats/yarle`
- Roll back formatting changes to main.ts and manifest.json

Another thing I would prefer is instead of using `document.querySelectorAll` and passing around `NodeListOf<XXXElement>`, let's use `document.body.findAll` which returns proper arrays like `HTMLElement[]`.

formats/notion/convert-to-md.ts
joshuatazrein: I'll change this to be more reliable—originally it was detecting any `href` from Notion that wasn't a valid `https://` URI. I can just detect that directly.

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
 	"id": "obsidian-importer",
-	"name": "Importer",
-	"version": "1.0.4",
+	"name": "Importer (Notion fork)",
```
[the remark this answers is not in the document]
joshuatazrein: I'll revert those for the merge

joshuatazrein:
- Changed to using `.findAll` instead of `querySelectorAll`
- Updated to use `ProgressReporter`
- checkout `formats/yarle` and `manifest.json` from master
- Merge `main.ts` with master

joshuatazrein:
Ok, I've dealt with a few more issues on Discord and I think this is ready to merge.

Issue #35, opened 2023-07-31
Detailed import log and progress bar - $500


kepano:
Allow users to see a detailed log, similar to Sync log with list of every file. This would be combined with the progress meter into a new screen that  is displayed while the import is running.

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $500 USD
**Timeframe:** 7 days

## Requirements

- [ ] Should be generic and compatible with existing/future converters
- [ ] Import progress modal should replace the "Import data" modal once import begins
- [ ] Progress meter
  - [ ] Number of remaining files (if available)
  - [ ] Number of skipped files
  - [ ] Number of failed files
  - [ ] Number of imported files
  - [ ] Progress bar should use total count of files if available, otherwise use an indefinite cycling progress bar  
- [ ] List of all files with states
  - [ ] Imported — timestamp, original file name, location of the new file (text color `--text-muted`)
  - [ ] Failed — timestamp, original file name, error message (text color `--text-warning`)
  - [ ] Skipped — timestamp, original file name, skipped message (text color: `--text-error`)
- [ ] Toggle "errors only" to only show failed and skipped files

## Reference

Mockup modified from current Sync log

<img width="979" alt="Screenshot 2023-07-31 at 2 06 47 PM" src="https://github.com/obsidianmd/obsidian-importer/assets/10565871/d27c17f5-d460-4213-b0e5-4e4a561d45eb">

Progress bar from [Google Keep Import plugin](https://github.com/daledesilva/obsidian_google-keep-import) by @daledesilva

<img width="979" alt="Screenshot 2023-07-31 at 2 06 47 PM" src="https://github.com/daledesilva/obsidian_google-keep-import/raw/main/docs/media/progress-preview.gif">

daledesilva:
I won't create a new issue incase you want to role it into this one, but I would suggest that a drag and drop and better summary section before import would be nice too. Currently, if you browse and select 1000s of files, you have to scroll a long way down to find the import button.

Here's an example from the same plugin
<img width="655" alt="Screenshot 2023-08-01 at 11 19 47 am" src="https://github.com/obsidianmd/obsidian-importer/assets/18710973/e70a08e5-ef7f-4d61-99ac-c18115f90e20">

Quorafind:
[...]

Issue #12, opened 2023-07-14
Import from Google Keep - $500


kepano:
Please upvote this issue if you would like to see Google Keep as an import format.

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $500 USD
**Timeframe:** 10 days

## Requirements

By using the existing HTML importer you should be to convert Google Keep's HTML export

- [ ] Must work on macOS, Windows, Linux
- [ ] Convert to proper Obsidian Markdown formatting
- [ ] Maintain folder structure
- [ ] Apply the created/modified date to the created Markdown file
- [ ] Follow the user's setting for `Default location for new attachments`
- [ ] Use Obsidian-style embed links for attachments `![[image.png]]`

## Current solutions

- [See Obsidian forum](https://forum.obsidian.md/t/google-keep-notes-to-obsidian-markdown-python-converter/12714)
- [obsidian_google-keep-import](https://github.com/daledesilva/obsidian_google-keep-import)

dbarenholz:
This (apparently) exists: https://github.com/daledesilva/obsidian_google-keep-import

daledesilva:
Hi!
I would like to apply for this bounty.
I am the developer of the [Google Keep Import community plugin](https://github.com/daledesilva/obsidian_google-keep-import) and so I am already familiar with format and the requirements.

In maintaining that plug-in, I have also been able to learn about different users setups and expectations in importing their keep notes that would allow me to more efficiently rewrite it for core inclusion.

kepano:
@daledesilva ok! You're assigned! Let us know in the Discord channel if you have any questions 😊

kepano:
Please make sure to submit your PR by end of day (GMT) on August 4th, 2023. Thanks!

daledesilva:
I've submitted my PR [here](https://github.com/obsidianmd/obsidian-importer/pull/37), but noted that I didn't meet the requirement of following the users default location for new attachments.

I was going to fix it, but I'd like to question it. I don't think this is the best approach.

Importing from Google Keep is a one time operation, so if the files land in an different assets folder, it is really just a quick drag and drop for them to move it where they would like, or to delete them if the import didn't meet their expectations somehow.

[...]

PR #37, opened 2023-08-01
Google Keep Import Pull Request

daledesilva:
Please note that I have included a partial Google Keep export in the repository for easy testing. This means json files, and binary files can be ignored during code review.

I've also push the edges of the existing modal layout so that I could provide the user relevant information and options when importing. Screenshot included below.

Screenshot 2023-08-01 at 4 49 53 pm


daledesilva:
Testing data is also zipped here, but please also test any of your own exports from Google Keep.
[testing-data.zip](https://github.com/obsidianmd/obsidian-importer/files/12225487/testing-data.zip)

daledesilva:
FYI. I just realised I didn't adhere to the requirement of "**Follow the user's setting for Default location for new attachments**"
I'll modify that to update the PR soon and post here.

daledesilva:
I've made a suggestion on the [original ticket](https://github.com/obsidianmd/obsidian-importer/issues/12#issuecomment-1660145980) where I advise we don't **Follow the user's setting for default attachment location**
So I'll consider this this PR ready for review unless someone makes a call otherwise.

polyipseity:
> FYI. I just realised I didn't adhere to the requirement of "**Follow the user's setting for Default location for new attachments**" I'll modify that to update the PR soon and post here.

In case you do have to follow the setting, there is an undocumented method `Vault.getAvailablePathForAttachments` that returns an available attachment path following the setting: https://github.com/obsidianmd/obsidian-importer/blob/9d5a2a9e1be3bb2ee7f66b2d05a9c9ef52af5b59/formats/html.ts#L245-L246

(I am guessing that slight use of undocumented methods are okay here.)

lishid:
Could we move the testing data to a separate PR, and put that in a `/tests/keep/` folder instead of mixing with the other sources?

daledesilva:
I've now updated the PR to support the recent changes to how file picking works.
I've also moved the testing files to [this PR](https://github.com/obsidianmd/obsidian-importer/pull/40) which should be reviewed after this one is merged.

**This is now ready to be reviewed.**
Since the testing files aren't included in here, you can refer to the first post above to download those testing files.

util.ts
```diff
@@ -1,18 +1,139 @@
...
+
 export function genUid(length: number): string {
 	let array: string[] = [];
 	for (let i = 0; i < length; i++) {
 		array.push((Math.random() * 16 | 0).toString(16));
 	}
 	return array.join('');
 }
+
+export function pathToFilename(path: string) {
```
lishid, line 55: This is no longer needed.

util.ts
```diff
@@ -1,18 +1,139 @@
...
+/**
+ * Retrieves a reference to a specific folder in a vault. Creates it first if it doesn't exist.
+ */
+export async function getOrCreateFolder(folderPath: string): Promise<TFolder> {
+	let normalizedPath = normalizePath(folderPath)
+	if(normalizedPath === '') {
+		normalizedPath = '/';
+	}
+
+	const folder = this.app.vault.getAbstractFileByPath(normalizedPath);
```
lishid, line 78: `this` has no reference.

util.ts
```diff
@@ -1,18 +1,139 @@
...
 function escapeRegex(str: string): string {
 	return str.replace(/[.?*+^$[\]\\(){}|-]/g, '\\$&');
 }
 
-const ILLEGAL_CHARACTERS = '\\/:*?<>\"|';
-const ILLEGAL_FILENAME_RE = new RegExp('[' + escapeRegex(ILLEGAL_CHARACTERS) + ']', 'g');
+const ILLEGAL_FILENAME_CHARACTERS = '\\/:*?<>\"|';
+const ILLEGAL_FILENAME_RE = new RegExp('[' + escapeRegex(ILLEGAL_FILENAME_CHARACTERS) + ']', 'g');
+
+const ILLEGAL_HASHTAG_CHARACTERS = '\\:*?<>\"|!@#$%^&()+=\`\'~;,.';
```
lishid, line 13: I would prefer if we consistently name these `tag` instead of mixing in `hashtag`

util.ts
lishid: Hmm seems like a lot of these are like that - could you pass `vault` in as a parameter?

util.ts
```diff
@@ -1,18 +1,139 @@
...
+}
+
+/**
+ * Adds an alias to the note's frontmatter.
+ * Only linebreak sanitzation is performed in this function.
+*/
+export async function addAliasToFrontmatter(alias: string, fileRef: TFile) {
+	const sanitizedAlias = alias.split('\n').join(', ');
+	await this.app.fileManager.processFrontMatter(fileRef, (frontmatter: any) => {      
+		if(!frontmatter['alias']) {
```
lishid, line 122: We now recommend the key `aliases` instead of `alias`.

daledesilva: Should tag still be singular or plural too?

lishid: Should be `tags`.

util.ts
daledesilva: Done. Good pickup. Weird that it even worked.

daledesilva:
All the above comments are now addressed 🙂

ericaxu:

daledesilva:
Thanks. I can see that, but I don't really no how to reply. There's no comment fields.

kepano:
> Thanks. I can see that, but I don't really no how to reply. There's no comment fields.

you can reply right here in the PR comments 😊

daledesilva:
oh.. ha ha. Whoops.

util.ts
```diff
@@ -1,18 +1,126 @@
...
 
 export function sanitizeFileName(name: string) {
 	return name.replace(ILLEGAL_FILENAME_RE, '');
 }
 
+/**
+ * Searches a string for characters unsupported by Obsidian in the hashtag body and returns a snaitized string.
+ * If the # symbol is included at the start or anywhere else it will be removed.
+ */
+export function sanitizeHashtag(name: string): string {
```
lishid, line 26: Also remember to rename `Hashtag` here and below.

util.ts
```diff
@@ -1,18 +1,126 @@
...
 		array.push((Math.random() * 16 | 0).toString(16));
 	}
 	return array.join('');
 }
+
+/**
+ * Retrieves a reference to a specific folder in a vault. Creates it first if it doesn't exist.
+ * Must pass in the relevant vault.
+ */
+export async function getOrCreateFolder(folderPath: string, vault: Vault): Promise<TFolder> {
```
lishid, line 58: I've made a helper function to cover this use case in FormatImporter that should help! (latest master)

formats/keep/add-keep-frontmatter.ts
```diff
@@ -0,0 +1,37 @@
+import { FileManager, TFile } from "obsidian";
+import { addAliasToFrontmatter, addTagToFrontmatter, sanitizeHashtag } from '../../util';
+import { KeepJson } from "./models/KeepJson";
+
+
+
+export async function addKeepFrontMatter(fileRef: TFile, keepJson: KeepJson, fileManager: FileManager) {
```
lishid, line 7: This is quite inefficient because it keeps re-reading the file from disk, writing it back. Can we wrap it all in a single `processFrontMatter` call, and just pass the frontmatter object to each helper?

formats/keep/models/KeepJson.ts
```diff
@@ -0,0 +1,50 @@
+export interface KeepListItem {
```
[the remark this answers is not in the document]
daledesilva: Sure.
I was keeping consistent with the models in the yarle folder. Happy to change it.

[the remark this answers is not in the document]
lishid: Good point, I will come back and fix those later!

formats/keep/add-keep-frontmatter.ts
daledesilva: Will do :)

daledesilva:
All of the above is addressed.
This is ready for further review.

formats/keep-json.ts
```diff
@@ -0,0 +1,202 @@
+import { FormatImporter } from "../format-importer";
+import { DataWriteOptions, Notice, Setting, TFile, TFolder } from "obsidian";
+import { ImportResult, ProgressReporter } from '../main';
+import { convertJsonToMd } from "./keep/convert-json-to-md";
+import { KeepJson, convertStringToKeepJson } from "./keep/models/keep-json";
+import { addAliasToFrontmatter, addTagToFrontmatter, toSentenceCase } from "../util";
+import { PickedFile, PickedFolder, parseFilePath } from "filesystem";
+import { BlobWriter, Entry, TextWriter } from "@zip.js/zip.js";
+import { Transform } from "stream";
```
lishid, line 9: This is unused.

formats/keep-json.ts
```diff
@@ -0,0 +1,202 @@
...
+					} else if(ATTACHMENT_EXTS.contains(innerFileProps.extension)) {
+						const rawContent = await entry.getData(new BlobWriter());
+						const arrayBuffer = await rawContent.arrayBuffer();
+						await this.copyFile(arrayBuffer, assetFolderPath, innerFileProps.name, progress);
+					}
+					// else: Silently skip any other unsupported files in the zip
+		
+				} catch(e) {
+					console.error(`${file.name} ::: ${curInnerFilename ?? curInnerFilename}`, e);
+					progress.reportFailed(`${file.name} / ${curInnerFilename}`, e);
```
lishid, line 134: Prefer to remove the extra spaces here around the `/`.

formats/keep-json.ts
```diff
@@ -0,0 +1,202 @@
...
+		
+					} else if(ATTACHMENT_EXTS.contains(innerFileProps.extension)) {
+						const rawContent = await entry.getData(new BlobWriter());
+						const arrayBuffer = await rawContent.arrayBuffer();
+						await this.copyFile(arrayBuffer, assetFolderPath, innerFileProps.name, progress);
+					}
+					// else: Silently skip any other unsupported files in the zip
+		
+				} catch(e) {
+					console.error(`${file.name} ::: ${curInnerFilename ?? curInnerFilename}`, e);
```
lishid, line 133: Let's avoid adding extra console messages since they are already reported in `reportFailed`.

formats/keep-json.ts
```diff
@@ -0,0 +1,202 @@
...
+
+		await this.convertKeepJson(keepJson, folder, title);
+		progress.reportNoteSuccess(`${title}.json`);
+	}
+
+	// Keep assets have filenames that appear unique, so no duplicate handling isn't implemented
+	async copyFile(arrayBuffer: ArrayBuffer, folderPath: string, filename: string, progress: ProgressReporter) {
+		let assetFolder = await this.createFolders(folderPath);
+		await this.vault.createBinary(`${assetFolder.path}/${filename}`, arrayBuffer);
+		progress.reportNoteSuccess(filename);
```
lishid, line 163: This should be `reportAttachmentSuccess`

main.ts
```diff
@@ -149,13 +154,15 @@ export class ImporterModal extends Modal {
 					el.addEventListener('click', async () => {
 						let progress = new ProgressReporter();
 						this.modalEl.addClass('is-loading');
+						new Notice('Import started.');
```
lishid, line 157: Let's not add stuff to the main script

util.ts
```diff
@@ -21,6 +54,46 @@ export function genUid(length: number): string {
...
+/**
+ * Takes a string and returns in lowercase with the first letter capitalised.
+ */
+export function toSentenceCase(str: string) {
+    return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+}
+
+/**
+ * Adds a single tag to the tag property in frontmatter and santises it.
+ * Must pass in app.fileManager.
```
lishid, line 66: This comment isn't accurate anymore.

util.ts
```diff
@@ -21,6 +54,46 @@ export function genUid(length: number): string {
...
+ * Only linebreak sanitization is performed in this function.
+ * Must pass in app.fileManager.
+*/
+export function addAliasToFrontmatter(frontmatter: any, alias: string) {
+	const sanitizedAlias = alias.split('\n').join(', ');
+	if(!frontmatter['aliases']) {
+		frontmatter['aliases'] = [sanitizedAlias];
+	} else {
+		if (!Array.isArray(frontmatter['aliases'])) {
+			frontmatter['aliases'] = frontmatter['aliases'].split(' ');
```
lishid, line 91: This should split by comma. Actually we should be using obsidian provided functions here for `parseFrontMatterTags` and `parseFrontMatterAliases`.

formats/keep/models/keep-json.ts
```diff
@@ -0,0 +1,50 @@
...
+    createdTimestampUsec: number;
+	userEditedTimestampUsec: number;
+    //
+	isArchived?: boolean;
+	isPinned?: boolean;
+	isTrashed?: boolean;
+    //
+	title?: string;
+	textContent?: string;
+	listContent?: Array<KeepListItem>;
```
lishid, line 31: Prefer to use `Type[]` instead of `Array<Type>`.

formats/keep/convert-json-to-md.ts
```diff
@@ -0,0 +1,36 @@
...
+            // Don't put in blank checkbox items
+            if(!listItem.text) continue;
+            
+            let listItemContent = `- [${listItem.isChecked ? 'X' : ' '}] ${listItem.text}\n`;
+            mdContent += sanitizeTags(listItemContent);
+        }
+    }
+
+    if(jsonContent.attachments) {
+        for(let i=0; i<jsonContent.attachments.length; i++) {
```
lishid, line 29: Use a for-of loop here instead?

lishid:
Does google keep export in zip format directly? What's the benefit of supporting flat folder of unzipped files? I'm thinking we should probably keep it zip-only because the file selector accepts a lot of files right now and it's quite some clutter.

daledesilva:
> Does google keep export in zip format directly? What's the benefit of supporting flat folder of unzipped files? I'm thinking we should probably keep it zip-only because the file selector accepts a lot of files right now and it's quite some clutter.

Google Keep doesn't let you selectively export notes, so you'll always get a zip of all your keep notes even if you only want some of them. For that reason, I think allowing users to selectively import is useful.

But I also don't think we should limit the functionality that's already built and working here simply because of a stylistic limitation of the dialog (Which will continue to create issues with other importers too). I mentioned this [here](https://github.com/obsidianmd/obsidian-importer/issues/35#issuecomment-1659423225).

You used the term "flat folder" though, which leads to believe that maybe you mean selecting the files by folder? I actually never intentionally allowed that. I used the method **addFileChooserSetting** back when it only gave me an individual file chooser, but at some point in pulling the master branch that method has changed to include both file and folder buttons.

formats/keep-json.ts
daledesilva: I'm curious why. This is about readability. The spaces make it much easier to identify the containing zip file and the inner file separately.

formats/keep-json.ts
daledesilva: Fixed and pushed.

formats/keep-json.ts
daledesilva: Fixed and Pushed.

main.ts
daledesilva: Fixed and pushed.
FYI. This was added because if the user is uploading a large number of files, it will appear to them that the program has hung - since there is no message and interface stays the same.

There is now https://github.com/obsidianmd/obsidian-importer/issues/35, but until that's implemented the issue will persist. Hence I implemented it in main as a placeholder because it shouldn't be implemented individually by importers.

util.ts
daledesilva: Fixed and pushed

util.ts
daledesilva: Fixed and pushed

formats/keep/models/keep-json.ts
daledesilva: Fixed and pushed

formats/keep/convert-json-to-md.ts
daledesilva: Fixed and pushed

daledesilva:
I've completed all but one edit above. Happy to do it but just thought I'd draw attention to the discussion first.
https://github.com/obsidianmd/obsidian-importer/pull/37#discussion_r1287733847

lishid:
Ah yes, so there's now an optional flag you can pass to indicate whether you allow multiple files to be chosen.

lishid:
> Google Keep doesn't let you selectively export notes, so you'll always get a zip of all your keep notes even if you only want some of them. For that reason, I think allowing users to selectively import is useful.

One thing I'm not understanding is allowing selection of everything including attachments, if you can only select single files to import, it doesn't make much sense to select single attachment files no?

daledesilva:
> One thing I'm not understanding is allowing selection of everything including attachments, if you can only select single files to import, it doesn't make much sense to select single attachment files no?

I'm not sure I'm interpreting what you're saying correctly.

You can select multiple files to import, therefore, the user would ideally select a json file or multiple and their related attachment at the same time, but even if they don't and they notice missing attachments later, they can rerun the importer and just select those attachments too.

Or, alternatively, they can just select the zip, or even multiple zips.

lishid:
Hmm I was replying to this part:

> You used the term "flat folder" though, which leads to believe that maybe you mean selecting the files by folder? I actually never intentionally allowed that. I used the method addFileChooserSetting back when it only gave me an individual file chooser, but at some point in pulling the master branch that method has changed to include both file and folder buttons.

Selecting the container folder would end up being the same as selecting all the files in that folder.

daledesilva:
> Selecting the container folder would end up being the same as selecting all the files in that folder.

Yes it is (apart from it being recursive). That's true with any importer.

I'm not sure if there's a change here you'd like. If it's to remove the folder option, that makes sense, but I'm already using addFileChooserSetting, does that now support options?

util.ts
```diff
@@ -1,8 +1,12 @@
+import { parseFrontMatterAliases, parseFrontMatterTags } from "obsidian";
+
 let illegalRe = /[\/\?<>\\:\*\|"]/g;
 let controlRe = /[\x00-\x1f\x80-\x9f]/g;
 let reservedRe = /^\.+$/;
 let windowsReservedRe = /^(con|prn|aux|nul|com[0-9]|lpt[0-9])(\..*)?$/i;
 let windowsTrailingRe = /[\. ]+$/;
+let potentialTagsRe = /(#[^ ^#]*)/g; // Finds any non-whitespace sections starting with #
```
lishid, line 8: Let's not mix these into the file name sanitizer code.

lishid: Actually, can we move all of these util functions into `formats/keep/util.ts`?

formats/keep/models/keep-json.ts
```diff
@@ -0,0 +1,50 @@
+export interface KeepListItem {
```
lishid, line 1: Can we move this up to `/formats/keep/models.ts`?

formats/keep/convert-json-to-md.ts
```diff
@@ -0,0 +1,33 @@
+import { sanitizeTags } from '../../util';
+import { KeepJson } from "./models/keep-json";
+
+/**
+ * Reads a Google Keep JSON file and returns a markdown string.
+ */
+export function convertJsonToMd(jsonContent: KeepJson): string {
+    let mdContent = '';
```
lishid, line 8: Let's make this into an array and use `join('\n')` at the end to avoid multiple string concatenations.

formats/keep-json.ts
```diff
@@ -0,0 +1,199 @@
...
+					progress.reportFailed(`${file.name} / ${curInnerFilename}`, e);
+				}
+			}
+		})
+	}
+
+	async importKeepNote(rawContent: string, folder: TFolder, title: string, progress: ProgressReporter) {
+		let keepJson = convertStringToKeepJson(rawContent);
+		if(!keepJson) {
+			progress.reportFailed(`${title}.json (Invalid format)`, 'NOT A GOOGLE KEEP JSON');
```
lishid, line 140: Can we update these to properly use a filename for the first parameter?

lishid (commented):
Ok just these and fixing the merge conflict and we should be good to go.

formats/keep-json.ts
daledesilva: Was going to flag this with you actually @lishid 
I'll change this, but please note that the ProgressReporter currently accepts a reason, but doesn't display it anywhere. Without that the user will have no idea why something was skipped, which will caused confusion.

Also, Are you happy with the format of the reasons specified in the second parameters?
I can't remember whether I put them in all caps for a reason.

lishid: That's right. ProgressReporter is a work-in-progress that I have setup for the backend API - we are working in parallel on the interface to display the progress as well as the reason/details.

Everything will be user-visible in the end; I would prefer if they were just regular prose instead of allcaps.
Also the file path might end up getting parsed for better formatting of containing folder vs filename itself, so that's why it's important to use a proper filename instead of adding debug information into it.

daledesilva: Makes sense, that was my expectation but just wanted to flag it.

formats/keep/convert-json-to-md.ts
daledesilva: Fixed & pushed

util.ts
daledesilva: Done & pushed.

formats/keep-json.ts
daledesilva: Done & pushed.

daledesilva:
All updated

util.ts
```diff
@@ -13,12 +14,29 @@ export function sanitizeFileName(name: string) {
 		.replace(windowsTrailingRe, '');
 }
 
-export function genUid(length: number): string {
```
lishid, line 16: Hmm this shouldn't have been moved...

lishid (commented):
Could you also run the auto formatter `npm run lint` to fix the inconsistent spaces?

PR #40, opened 2023-08-02
Google Keep Testing files

daledesilva:
This pull request only adds an unaltered Google Keep export file for future testing.
It's been included as a zip to keep it in the exact format exported from Keep.

The other [Google Keep Format PR](https://github.com/obsidianmd/obsidian-importer/pull/37) should be reviewed before this one as this one builds on it. Once that PR is merged this PR will only show one zip file as added.


ericaxu:
This PR also adds a bunch of new util functions. Is that intentional?

daledesilva:
They're actually added by the other PR that should be merged first. Once that's merged this will only have 1 file.

daledesilva:
This has been updated to reflect the recent release.

PR #42, opened 2023-08-04
Replace HTML interlinks

polyipseity:
Requires #33. ~~Comparison when #33 is merged: 6b88a4859651a803abe887164ecf44be416bf259...6c61e0b91762f54c7d1b2a9b8e1584175419a952~~ Rebased.

This makes the importer convert valid `<a>` links linking other HTML files into wikilinks or Markdown links linking the imported Markdown files depending on `Files & Links > Use [[Wikilinks]]`. Example:

- `a.html` -> `a.md`
- `b.html` -> `b.md`:
```html
<a href="./a.html">abc</a>
```
becomes
```md
[[a]]
```

(It's not `[[a|abc]]` because apparently `ReferenceCache.displayText` is always the empty string if the link is a Markdown link even if the Markdown link has a display like `[abc](./a.html)`. As the intermediate form uses Markdown links, this causes `[[a]]` instead of `[[a|abc]]` to be outputted. Maybe this needs to be fixed on your side? A temporary fix would be to parse the `displayText` out from `ReferenceCache.original`, but I am not exactly wanting to parse Markdown links myself yet...)


src/formats/html.ts
```diff
@@ -75,10 +75,85 @@ export class HtmlImporter extends FormatImporter {
...
+								const url = new URL(encodeURI(path), fileKey);
+								url.hash = '';
+								url.search = '';
+								linkKey = decodeURIComponent(url.href);
+							}
+							else {
+								linkKey = parseFilePath(path.replace(/#/gu, '%23')).name;
+							}
+							if (fileLookup.has(linkKey)) {
+								const newLink = this.app.fileManager.generateMarkdownLink(fileLookup.get(linkKey)!.tFile, tFile.path, subpath, displayText);
```
tgrosinger, line 139: Please rewrite this to avoid the assertion. For example, get the value from `fileLookup` first, then check to see if it is defined instead of using `.has`.

tgrosinger (changes requested):
> (It's not [[a|abc]] because apparently ReferenceCache.displayText is always the empty string

When I tested this on the files in `test/html` I was seeing that the link became `[[HTML import/embeds|embeds.html]]` as expected.

Looking good. Left a few changes and a few questions. Thank you.

polyipseity:
> > (It's not [[a|abc]] because apparently ReferenceCache.displayText is always the empty string
> 
> When I tested this on the files in `test/html` I was seeing that the link became `[[HTML import/embeds|embeds.html]]` as expected.

Presumably, it is fixed.

src/formats/html.ts
polyipseity: There is also one outside of my diff here: https://github.com/obsidianmd/obsidian-importer/blob/cfdfac1bd55eec84b2aca8cc35defae6365cc730/src/formats/html.ts#L172

src/formats/html.ts
```diff
@@ -75,10 +75,85 @@ export class HtmlImporter extends FormatImporter {
...
+				const resolved = new Promise<void>(resolve => {
+					const ref = this.app.metadataCache.on('resolved', () => {
+						this.app.metadataCache.offref(ref);
+						resolve();
+					});
+				});
+				const appended = '\n';
+				await this.app.vault.append(aFile, appended);
+				await resolved;
+				await this.app.vault.process(aFile, data => data.endsWith(appended) ? data.slice(0, -appended.length) : data);
```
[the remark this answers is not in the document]
polyipseity: This block of code in the `if` branch ensures that metadata cache for the newly createdfiles is up-to-date by waiting for `resolved`. Since `resolved` is triggered only once the metadata cache is up-to-date after a file modification, if `resolved` triggered by creating the new files is fired before `await resolved`, then `await resolved` will block forever. Making a file change here ensures `resolved` is triggered.

Anyway, I have thought of another way of not needing to append an newline and made this change in 280f148eba9c6e9edec453f6a4b67ade02967c8f.

tgrosinger (approved):
Looks good to me, but I'd like to get a quick look from @lishid too.

src/formats/html.ts
```diff
@@ -75,10 +75,89 @@ export class HtmlImporter extends FormatImporter {
...
+		const fileLookup = new Map<string, { file: PickedFile, tFile: TFile }>;
+		let resolved: Promise<void> | undefined;
+
+		ctx.reportProgress(0, files.length);
 		for (let i = 0; i < files.length; i++) {
 			if (ctx.isCancelled()) return;
 			ctx.reportProgress(i, files.length);
-			await this.processFile(ctx, folder, files[i]);
+			const prevResolved = resolved;
+			resolved = new Promise<void>(resolve => {
```
lishid, line 86: There is a race condition here was the resolved was emitted by a previous file before the next processFile was written, which means this wouldn't really work. Since we're using private APIs anyway we might as well use `MetadataCache.isCacheClean()` instead of keep generating these promises and hooking events.

lishid (approved):

src/formats/html.ts
tgrosinger: @lishid I updated this to use `MetadataCache.OnCleanCache()`. A quick test using the files in this repo seems to work as expected. Let me know what you think.

Issue #16, opened 2023-07-14
Import from Roam - $500


kepano:
Please upvote this issue if you would like to see import from Roam

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $500 USD
**Timeframe:** 10 days

## Requirements

- [ ] Resolve page names with allowed characters only
  - [ ] Resolve date format for daily notes, convert to the user's daily note format (`YYYY-MM-DD` by default)
- [ ] Convert outlines to bullet lists with block references
- [ ] Convert highlights and todos to Obsidian Markdown
- [ ] Support for attachments

## Current solutions

- See [Obsidian forum](https://forum.obsidian.md/t/meta-migration-workflows/15252)

8bitgentleman:
Hello, I have extensive experience building extensions around the Roam format and already have 13 extensions reviewed and available in the Roam extension depot (with 3 more awaiting approval).
I would like to put my name forward for this bounty, here is a link to all of my approved Roam extensions 
https://github.com/Roam-Research/roam-depot/tree/main/extensions/8bitgentleman

kepano:
@8bitgentleman have you used Obsidian, and can you share any code related to Obsidian (e.g. community plugins) — we just want to make sure you have some familiarity with how to implement things on the Obsidian side.

8bitgentleman:
Hi @kepano Yes I do use Obsidian. I don't have any Obsidian plugins published to the larger community but I do have a personal one that I've just put onto github so you can take a look. It's bare bones and catered for my use case now but if I were to publish it I'd probably add more error handling and settings options for a user to customize what metadata is imported 

https://github.com/8bitgentleman/obsidian-import-book-metadata

kepano:
@8bitgentleman You are assigned! Please make sure to submit your PR by end of day (GMT) on August 10th, 2023. Thanks!

Join the #plugin-dev/importer channel: https://discord.gg/obsidianmd

PR #50, opened 2023-08-07
Roam import

8bitgentleman:
Here is my submission for the Roam Importer Bounty #16 
- [x]  Resolve page names with allowed characters only
 - [x]  Resolve date format for daily notes, convert to the user's daily note format (YYYY-MM-DD by default)
 - [x]  Convert outlines to bullet lists with block references
 - [x]  Convert highlights and todos to Obsidian Markdown
 - [x]  Support for attachments

Because of the complexity I use a 3 stage import 

1. pre-process (create a map of
2. write to markdown
3. edit markdown to resolve blockRefs

This is the first plugin I've officially submitted to Obsidian (though not the first I've made) and it's a bit of a beast. Please let me know if there is anything to cleanup/change.


formats/roam-json.ts
```diff
@@ -0,0 +1,46 @@
...
+
+		let folder = await this.getOutputFolder();
+		if (!folder) {
+			new Notice('Please select a location to export to.');
+			return;
+		}
+
+		let { app } = this;
+		let adapter = app.vault.adapter;
+		if (!(adapter instanceof FileSystemAdapter)) return;
```
lishid, line 42: This isn't necessary.

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+	blockText = blockText.replaceAll(/\[.+?\]\((\(.+?\)\))\)/g, "$1")
+	// page alias
+	blockText = blockText.replaceAll(/\[(.+?)\]\(\[\[(.+?)\]\]\)/g, "[[$2|$1]]")
+
+	blockText = blockText.replaceAll("[[>]]", ">");
+	blockText = blockText.replaceAll("{{TODO}}", "[ ]");
+	blockText = blockText.replaceAll("{{[[TODO]]}}", "[ ]");
+	blockText = blockText.replaceAll("{{DONE}}", "[x]");
+	blockText = blockText.replaceAll("{{[[DONE]]}}", "[x]");
+	blockText = blockText.replaceAll("{{word-count}}", "");
```
lishid, line 111: Could we loop all of these that remove everything into a single regex so that we don't create a bunch of intermediate strings?

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+	//   blockText = blockText.replaceAll(/\!\[\]\((.+?)\)/g, "$1"); //imags with no description
+	//   blockText = blockText.replaceAll(/\[(.+?)\]\((.+?)\)/g, "$1: $2"); //alias with description
+	//   blockText = blockText.replaceAll(/\[\]\((.+?)\)/g, "$1"); //alias with no description
+	//   blockText = blockText.replaceAll(/\[(.+?)\](?!\()(.+?)\)/g, "$1"); //alias with embeded block (Odd side effect of parser)
+
+	return blockText;
+  };
+
+async function jsonToMarkdown(graphFolder:string, attachmentsFolder:string, downloadAttachments:boolean,json: JsonObject, indent: string = '', isChild: boolean = false ): Promise<string> {
+    let markdown = '';
```
lishid, line 161: Can we use an array and `join('')` at the end instead of keep concatenating the string?

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+        const graphName = sanitizeFileName(file.basename);
+        const graphFolder = path.join(folder.path, graphName);
+		const attachmentsFolder = path.join(folder.path, graphName, "Attachments")
+        // create the base graph folders
+        await importer.createFolders(graphFolder)
+		await importer.createFolders(attachmentsFolder)
+
+		// read the graph
+		// TODO is this async?
+        const data = fs.readFileSync(file.filepath, "utf8")
```
lishid, line 191: You need to use `await file.readText()` instead.

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    return markdown;
+}
+
+export async function importRoamJson (importer:RoamJSONImporter, progress: ProgressReporter, files:PickedFile[], folder:TFolder, downloadAttachments:boolean = true) {
+    for (let file of files) {
+        const graphName = sanitizeFileName(file.basename);
+        const graphFolder = path.join(folder.path, graphName);
```
lishid, line 183: `path.join` doesn't work on mobile, plus for internal use we can simply put it between `/`.

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+            // convert json to nested markdown
+
+			const markdownOutput = await jsonToMarkdown(graphFolder, attachmentsFolder, downloadAttachments, pageData);
+            
+            try {
+				//create folders for nested pages [[some/nested/subfolder/page]]
+                await importer.createFolders(path.dirname(filename))
+				const existingFile = app.vault.getAbstractFileByPath(filename);
+				if (existingFile) {
+					await app.vault.adapter.write(existingFile.path, markdownOutput);
```
lishid, line 212: Should use `vault.modify` instead of `vault.adapter.write`.

formats/roam/roam.ts
```diff
@@ -0,0 +1,327 @@
...
+				}
+				progress.reportNoteSuccess(filename);
+			} catch (error) {
+				console.error("Error saving Markdown to file:", filename, error);
+				progress.reportFailed(pageName);
+			}
+        }
+
+		// POST-PROCESS: fix block refs //
+		async function modifySourceBlockString(sourceBlockUID:string) {
```
lishid, line 226: This function is awfully bad for performance as it keeps reading files, blowing it up into lines, modifying a single line and re-save it to disk. Can we prepare a list of block changes per target file and do a single read-write (using `vault.process`) per file instead?

formats/roam/roam_dl_attachment.ts
```diff
@@ -0,0 +1,51 @@
...
+        } else {
+            // I expect this to be a bare link which is typically a binary file
+            link = line.match(binaryRegex);
+            syntaxLink = line.match(/https:\/\/firebasestorage.*?alt=media&.*?(?=\s|$)/);
+        } 
+        
+        if (link) {
+            const firebaseShort = 'https://firebasestorage' + link[1];
+            const firebaseUrl = link[0].slice(0, -1);
+            const response = await axios.get(firebaseUrl, { responseType: 'arraybuffer' });
```
lishid, line 30: Can't use axios. You can use fetch, and fallback to `requestUrl` from `obsidian`.

formats/roam/samples/help-graph.json
lishid, line 1: These needs to be moved to /tests/ folder.

util.ts
```diff
@@ -57,3 +83,30 @@ export function uint8arrayToArrayBuffer(input: Uint8Array): ArrayBuffer {
 export function stringToUtf8(text: string): ArrayBuffer {
 	return uint8arrayToArrayBuffer(new TextEncoder().encode(text));
 }
+export function getUserDNPFormat(){
+	// @ts-expect-error : Internal Method
+	const dailyNotePluginInstance = app.internalPlugins.getPluginById("daily-notes").instance;
```
lishid, line 88: app refers to global - we should use a proper reference.

util.ts
```diff
@@ -1,16 +1,42 @@
+import moment from "moment";
```
lishid, line 1: Moment needs to be imported from `obsidian` instead so we don't bundle a copy into the plugin.

tsconfig.json
```diff
@@ -17,7 +17,8 @@
 			"DOM",
 			"ES5",
 			"ES6",
-			"ES7"
+			"ES7",
+			"es2021"
```
lishid, line 21: What's this for?

package.json
```diff
@@ -26,6 +26,7 @@
 	},
 	"dependencies": {
 		"@zip.js/zip.js": "2.7.20",
+		"axios": "^1.4.0",
```
lishid, line 29: Let's not import this since it breaks mobile support.

formats/roam/roam_dl_attachment.ts
```diff
@@ -0,0 +1,51 @@
...
+            const timestamp = Math.floor(Date.now() / 1000);
+            const reg = firebaseShort.slice(-5).match(/(.*?)\.(.+)/);
+            if (reg) {
+                const ext = '.' + reg[2];
+
+                const newFilePath = path.join(attachmentsFolder, `${timestamp}${ext}`)
+                // Convert ArrayBuffer to Buffer
+                const data = Buffer.from(response.data);
+                // Write the file using Node.js's fs module
+                await fs.writeFileSync(path.join(app.vault.adapter.basePath, newFilePath), data);
```
lishid, line 40: `vault.createBinary` is what we should be using here.

lishid (commented):
Looks like a great start!

Since Roam uses json, I don't see why we need to use nodejs exclusively. This would mean that we can easily make this importer support mobile devices. To do so, we must avoid the use of `fs`, `path`, `Buffer`, and any node-based dependencies like axios.

There are replacements for everything this PR is using so I am optimistic.

Also, can we move the utils into `/formats/roam` for now?

tsconfig.json
8bitgentleman: Was getting the error **Property 'replaceAll' does not exist on type 'string'** and adding es2021 fixed it. I will remove

formats/roam/roam.ts
8bitgentleman: Not sure what you're looking for here. Do you mean create a single regex to replace things like these

{{word-count}}
{{date}}
{{[[POMO]]}}
{{POMO}}
{{slider}}
{{[[slider]]}}
{{encrypt}}
{{[[encrypt]]}}
{{TaoOfRoam}}
{{orphans}}
{{[[orphans]]}}
{{count}}
{{character-count}}
{{comment-button}}

But avoid strings that have custom replacements like these?
{{TODO}}
{{[[TODO]]}}
{{DONE}}
{{[[DONE]]}}
{{embed: ((asdf))}}
{{[[embed]]: [[asadf]]}}

lishid: Yeah the custom ones that are replaced with `[ ]` and `[x]` I guess you can keep, but I think it would be best to use a single regex for the other ones that just get replaced with empty string.

util.ts
8bitgentleman: don't I need to use global app to get an internal plugin? What's the alternative?

lishid: You can get an explicit reference to app via this.app on your FormatImporter.

lishid:
Hey @8bitgentleman, I see some updates since the last review but no replies - were you able to address them? Are there anything that was unclear or needs further discussion? And are you waiting on another pass over the code from my side?

kepano:
Sorry to hear that @8bitgentleman, do not worry about it. We will pay you the bounty for the work you have done so far, and find a way to complete the work on our side. Thank you so much, and best of luck ❤️

lishid:
I should be able to jump on the branch and finish the rest of the changes once I wrap up my current tasks. Sorry to hear about your situation.

ericaxu:
> Thank you for understanding @kepano and team

I've already reached out in DM ❤️ 🫂

tgrosinger:
I have continued the work started in this PR in #85.

Issue #11, opened 2023-07-14
Import from OneNote - $1,000


kepano:
Please upvote this issue if you would like to see OneNote as an import format.

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $1,000 USD
**Timeframe:** 10 days

## Requirements

- [ ] Export files from OneNote
- [ ] Must work on macOS and Windows
- [ ] Convert to proper Obsidian Markdown formatting
- [ ] Maintain folder structure
- [ ] Parse internal links
- [ ] Apply the created/modified date to the created Markdown file

## Current solutions

- [ConvertOneNote2MarkDown](https://github.com/theohbrothers/ConvertOneNote2MarkDown/issues/123)
  - More detail on [Obsidian forum](https://forum.obsidian.md/t/new-tool-for-migration-from-onenote-updated-and-improved-version/3055)

ericaxu:
Please note that OneNote has different versions and @lishid was stuck in one of the older versions like OneNote 2015 or something. I think we'll only offer importing from the latest export formats.

Quorafind:
Need it!

claremacrae:
Have you had any thoughts about platforms this would need to run on?

The linked tool requires Windows.

OneNote also runs on Macs, iPhone and iPad, for example.

kepano:
> Have you had any thoughts about platforms this would need to run on?

The solution must work on macOS and Windows. Importer is not currently available for mobile.

daledesilva:
I'm currently working on another format, but I'm very interested in this one.
In preparation for whoever takes it up, it might be worth considering which aspects of OneNote are definite ins and outs for this bounty.

I do a lot of pen based note taking on my iPad in OneNote and have been meaning to explore whether drawings are exposed to the API. Even if they are, they'll be a significant task to parse and translate to Excalidraw (I think that's the only option?) And it would also create a dependency on Excalidraw.

My guess for inclusion would be text, images, audio recordings, embedded files (if as easy as audio recordings), and tables. I doubt drawings could be included as my guess is it would blow this out, but just noting all these for discussion (Not an exhaustive list).

claremacrae:
That’s a really good point.
I had thought I would migrate all my OneNote notes in to Obsidian.

I understand why drawings would be hard, but without them, it’s not a complete migration and I would just leave everything in OneNote.

[...]

PR #51, opened 2023-08-07
Add OneNote importing

p3rid0t:
This pull request introduces the functionality to import content from OneNote, in accordance with the specifications detailed in the original [OneNote importer issue](https://github.com/obsidianmd/obsidian-importer/issues/11).

The implementation of this pull request utilizes the Microsoft Graph API. Therefore, the Obsidian team is required to generate their unique client ID by creating an application in the Azure AD Portal. Following this, the `CLIENT_ID` const must be updated and some specific adjustments need to be made within the Azure AD Portal. As always, I'm open to helping with this process ;)

As always, performance is really important to me, so here are a few things I did to improve it:
- I wrote the device code sign in feature myself
- I minimized the amount of requests made
- I used lesser-known techniques to boost performance, such as sorting by creation date, as detailed in this [official blog post](https://devblogs.microsoft.com/microsoft365dev/a-few-performance-tips-for-using-the-onenote-api/)
Those seemingly small changes have saved a few MB from being used for unnecessary dependencies for the SDK, and made the section list load faster by around 250ms on average.

Even though drawings were not mentioned in the original requirements, they are a very important part of OneNote for me and other users. Currently, the only available option was the [microsoft/InkMLjs](https://github.com/microsoft/InkMLjs) library which still has a dependency on ***JQuery 1.5***... I decided to take on the challenge of rewriting it in TypeScript, but sadly that was too big of a goal to finish before the deadline, so I plan on working on the rewrite as soon as I can.

I am looking forward to improving the importer as soon as I get any code/user feedback!

---


formats/onenote/models/device-code.ts
```diff
@@ -0,0 +1,16 @@
+export interface DeviceCode {
+    device_code: string;
```
ericaxu, line 2: Is this specified by some kind of API? If not we should switch to camel case

formats/onenote/inkml
```diff
@@ -0,0 +1,527 @@
+// @ts-ignore
```
ericaxu, line 1: This file has no extension... Should we review this for the initial submission?

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
+import { FormatImporter } from 'format-importer';
+import { ProgressReporter } from 'main';
+import { DataWriteOptions, Notice, Setting, TFile, TFolder, htmlToMarkdown, requestUrl } from 'obsidian';
+import { OnenotePage, OnenoteSection, Notebook, SectionGroup } from '@microsoft/microsoft-graph-types';
+import { parseHTML } from '../util';
+import { DeviceCode, TokenResponse } from './onenote/models/device-code';
+
+const GRAPH_CLIENT_ID: string = 'c1a20926-78a8-47c8-a2a4-650e482bd8d2'; // TODO: replace with an Obsidian team owned client_Id
```
ericaxu, line 8: Is there any rate limiting for this?

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+const GRAPH_SCOPES: string[] = ['user.read', 'notes.read'];
+const ATTACHMENT_EXTS: string[] = ['png','webp','jpg','jpeg','gif','bmp','svg','mpg','m4a','webm','wav','ogv','3gp','mov','mp4','mkv','pdf'];
```
ericaxu, line 10: nit: space after each comma

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			.setDesc('If disabled, attachments will be stored in the export folder in the OneNote Attachments folder.')
+			.addToggle((toggle) => {
+				toggle.setValue(true).onChange((value) => (this.useDefaultAttachmentFolder = value));
+			});
+		new Setting(this.modal.contentEl)
+			.setName('Import incompatible attachments')
+			.setDesc('Imports incompatible attachments which cannot be embedded in Obsidian, such as .exe files.')
+			.addToggle((toggle) => {
+				toggle.setValue(false);
+				toggle.onChange((value) => (this.importIncompatibleAttachments = value));
```
ericaxu, line 37: nit: chain this

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			});
+
+		// Create a wrapper for sign in related settings in order to hide them later
+		const contentArea = this.modal.contentEl.createDiv({
+			cls: 'contentArea',
```
ericaxu, line 42: remove this (see comment below)

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			cls: 'modal-title',
+		});
+		let description = contentArea.createEl('p');
+		// This could possibly use the version from DeviceCode.message, as it returns a string in the user's language?
+		description.innerHTML = `Go to <a href="https://microsoft.com/devicelogin">microsoft.com/devicelogin</a> on your PC or phone and enter this code: <b>${await this.generateLoginCode()}</b>`;
+
+		new Setting(contentArea)
+			.setName('Custom user access token')
+			.setDesc('If you are having troubles with the device code, use a custom access token from Microsoft Graph Explorer by going to the access token tab under the address bar.')
+			.addText((text) => text.setPlaceholder('Paste token here').onChange(async (e) => this.signIn(e)));
```
ericaxu, line 55: nit: it looks more clear if you write `.setPlaceholder` and it's own line and then `.onChange` on another line

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+	async showSectionPickerUI() {
+		const response = await fetch(
+			'https://graph.microsoft.com/v1.0/me/onenote/notebooks?$expand=sections($select=id,displayName)&$select=id,displayName&$orderby=createdDateTime',
+			this.requestParams
+		);
+		const data = await response.json();
+		const notebooks: Notebook[] = data.value;
+
+		// Replace the sign in area to declutter the UI
+		const contentArea: HTMLDivElement = this.modal.contentEl.querySelector('.contentArea')!;
```
ericaxu, line 116: keep a reference of this element instead of querying it by selector

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+		contentArea.innerHTML = '';
```
ericaxu, line 117: use `.empty()`

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+		contentArea.createEl('h3', {
+			text: 'Choose what to import',
+			cls: 'modal-title',
+		});
+		contentArea.createEl('hr');
+
+		notebooks.forEach((notebook) => {
+			contentArea.createEl('h5', {
+				text: notebook.displayName!,
```
ericaxu, line 126: why do you assume it's not `null`?

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+				cls: 'modal-title',
+			});
+			let sections: OnenoteSection[] = notebook.sections!;
+			// All notebooks have sections but not all notebooks have section groups
+			let sectionGroups: SectionGroup[] | undefined | null = notebook?.sectionGroups;
+
+			this.createSectionList(sections);
+			sectionGroups?.forEach((sectionGroup) => {
+				this.modal.contentEl.createEl('h6', {
+					text: sectionGroup.displayName!,
```
ericaxu, line 136: ditto

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			cls: 'modal-title',
+		});
+		contentArea.createEl('hr');
+
+		notebooks.forEach((notebook) => {
+			contentArea.createEl('h5', {
+				text: notebook.displayName!,
+				cls: 'modal-title',
+			});
+			let sections: OnenoteSection[] = notebook.sections!;
```
ericaxu, line 129: ditto

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+				this.modal.contentEl.createEl('h6', {
+					text: sectionGroup.displayName!,
+				});
+				this.createSectionList(sectionGroup.sections!);
+			});
+		});
+	}
+
+	createSectionList(sections: OnenoteSection[]) {
+		const contentArea: HTMLDivElement = this.modal.contentEl.querySelector('.contentArea')!;
```
ericaxu, line 144: same, please keep a reference and don't query DOM

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+
+		sections?.forEach((section) => {
+			let label = contentArea.createEl('label');
```
ericaxu, line 147: have you considered using Obsidian `Setting` for this? that would ensure the UI is consistent with rest of Obsidian

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			let mdContent = htmlToMarkdown(parsedPage).trim();
+			const fileRef = await this.saveAsMarkdownFile(folder, page.title!, mdContent);
+
+			await this.fetchAttachmentQueue(progress, fileRef);
+
+			// Add the last modified and creation time metadata
+			const writeOptions: DataWriteOptions = {
+				ctime: Date.parse(page.createdDateTime!.toString()) ||
+					   Date.parse(page.lastModifiedDateTime!.toString()) ||
+					   Date.now(),
```
ericaxu, line 206: if createdDateTime and lastModifiedDateTime are both not null, is `Date.now()` fallback needed?

formats/onenote.ts
```diff
@@ -0,0 +1,345 @@
...
+			}
+		});
+		return pageElement;
+	}
+
+	// This function gets all attachments and adds them to the queue, as well as adds embedding syntax for supported file format
+	getAllAttachments(pageHTML: HTMLElement): HTMLElement {
+		try {
+			const objects: HTMLElement[] = pageHTML.findAll('object');
+			const images: HTMLElement[] = pageHTML.findAll('img');
```
ericaxu, line 291: Is `video` a possibility here?

ericaxu:
I did a very rough first pass -- will need to test it out to give more comments. Thank for submitting it on time!

ericaxu:
Additional comment: It doesn't seem like it's ready for testing yet. See [Daniel's comment on Discord](https://discord.com/channels/686053708261228577/1133074995271188520/1138578991066927226).

p3rid0t:
I think it should be ready for re-review, all bugs I encountered while testing have been fixed, plus everything should import outside drawings (I'll add those later on). Thanks in advance!

lishid:
Did you look at the previous comments that @ericaxu left? I'll take a look as soon as you can fix the conflict with master and address those comments. (See them in between the code here: <https://github.com/obsidianmd/obsidian-importer/pull/51/files>)

p3rid0t:
Really sorry for the late reply, I didn't get any notifications or mails from GitHub... 
I addressed most of the previous comments and fixed most issues. 

I'll try my best to fix the conflict with the `master` branch tomorrow :)

p3rid0t:
@lishid Comments from the Discord server were addressed and merge issues fixed! 

I think the code is ready for review, looking forward for feedback. Thanks in advance!

Tomorrow, I'll work on the [help page for the OneNote importer](https://help.obsidian.md/import/onenote).

main.ts
```diff
@@ -0,0 +1,223 @@
+import { App, Modal, Plugin, Setting } from 'obsidian';
```
tgrosinger, line 1: This file was moved to `src/main.ts` and should not be added back in this PR please.

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+		let outputFolder = await this.getOutputFolder();
+		if (!outputFolder) {
+			new Notice('Please select a location to export to.');
+			return;
+		}
+
+		for (let section of this.selectedSections) {
+			let sectionFolder: TFolder = await this.createFolders(outputFolder.path + '/' + section.displayName);
+
+			let pages: OnenotePage[] = (await (await fetch(`
```
tgrosinger, line 200: It would be nice to have some error handling in here, so we can notify if we fail to load a page from a notebook.

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+		}
+		catch (e) {
+			progress.reportFailed(page.title!, e);
+		}
+	}
+
+	// OneNote returns page data and inking data in one file, so we need to split them
+	convertFormat(input: string): { html: string; inkml: string } {
+		const output = { html: '', inkml: '' };
+		const boundary = input.split('\n')[0];
```
tgrosinger, line 247: If we're only using the first line, then we can improve performance by only splitting out to the first line break.

```suggestion
		const boundary = input.split('\n', 1)[0];
```

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+		const parts = input.split(boundary);
+
+		for (let part of parts) {
+			if (part.trim() === '') continue;
+
+			let contentTypeLine = part.split('\n').find((line) => line.includes('Content-Type'));
+
+			let contentType = contentTypeLine!.split(';')[0].split(':')[1].trim();
```
tgrosinger, line 255: Is it safe to assert `contentTypeLine`? Maybe add a check beforehand and report a parsing error if we can't find the content type.

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+
+			// Extract the value from the part by removing the first two lines and then splitting by the boundary delimiter
+			let value = part
+				.split('\n')
+				.slice(2)
+				.join('\n')
+				.split(boundary)[0]
```
tgrosinger, line 262: `parts` was already created by splitting on `boundary`. Is this split going to do anything?

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+				.trim();
+
+			if (contentType === 'text/html') {
+				output.html = value;
+			}
+			else if (contentType === 'application/inkml+xml') {
+				const lines = value.split('\n');
+				lines.pop(); // Remove the last line
```
tgrosinger, line 270: What is the last line? Can you leave a comment about why it needs to be removed?

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+		const tables = pageElement.querySelectorAll('table');
+  
+		tables.forEach(table => {
+			if (table.rows.length > 1) table.deleteRow(0); 
+		});
+		
+		return pageElement;
+	}
+
+	async convertDrawings(element: HTMLElement, currentFile: TFile | undefined = undefined): Promise<HTMLElement> {
```
tgrosinger, line 464: `currentFile` is never used or provided.

tgrosinger, line 464: Why is this function async?

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+		const walker = document.createTreeWalker(element, NodeFilter.SHOW_COMMENT, null);
+		let hasDrawings: boolean = false;
+
+		while (walker.nextNode()) {
+		  const commentNode = walker.currentNode as Comment;
+		  if (commentNode.nodeValue?.trim() === 'InkNode is not supported') hasDrawings = true;
+		}
+
+		if (hasDrawings) {
+			const textNode = document.createTextNode('> [!caution] This page contained an drawing which was not converted. Try exporting again later, into the same output folder.');
```
tgrosinger, line 475:
```suggestion
			const textNode = document.createTextNode('> [!caution] This page contained a drawing which was not converted. Try exporting again later into the same output folder.');
```

tgrosinger: We might want to leave out the "Try exporting again later" unless we intend for this to change really soon, and to support idempotent importing. It would also introduce potentials for conflicts if the OneNote or the Obsidian side has the content changed between imports.

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+			// Insert the notice at the top of the page
+			element.insertBefore(textNode, element.firstChild);	
+		}
+		else {
+			for (let i = 0; i < element.children.length; i++) {
```
tgrosinger, line 480: Why do we need to descend into the elements children here? Won't the treeWalker retrieve all elements of concern?

src/formats/onenote.ts
```diff
@@ -0,0 +1,489 @@
...
+
+	async generateLoginCode(): Promise<string> {
+		const requestBody = new URLSearchParams({
+			client_id: GRAPH_CLIENT_ID,
+			scope: GRAPH_SCOPES.join(' '),
+		});
+
+		// Using requestUrl in order to prevent CORS issues
+		const tokenResponse = await requestUrl(
+			`https://login.microsoftonline.com/common/oauth2/v2.0/devicecode?${requestBody.toString()}`
```
tgrosinger, line 87: I'm concerned about the security and possibility of phishing with this auth method. The client ID is embedded in the plugin and could be used outside of this plugin by anyone. From my research, this auth method is intended only for limited-input devices (e.g. TVs).

Please use the [auth code grant](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow) authentication flow. Obsidian supports plugins [registering custom URIs](https://help.obsidian.md/Concepts/Obsidian+URI) which can be used for the auth redirect.

tgrosinger: See also https://docs.obsidian.md/Reference/TypeScript+API/Plugin/registerObsidianProtocolHandler

p3rid0t:
The PR should be ready for re-review. The latest commit fixes previous issues and addresses comments made by @tgrosinger. Additionally, it does a refactor by splitting all MS Graph related functions into another file.

src/formats/onenote.ts
```diff
@@ -0,0 +1,444 @@
...
+
+	async import(progress: ProgressReporter): Promise<void> {
+		let outputFolder = await this.getOutputFolder();
+
+		if (!outputFolder) {
+			new Notice('Please select a location to export to.');
+			return;
+		}
+
+		for (let section of this.selectedSections) {
```
tgrosinger, line 142: It would be nice to have the remaining portion of the importer progress bar work.

What about adding 1 to the remaining for each section that we have left. Then when we get into the section we can add the number of pages in that section too. So by the end the number remaining should be 0, but it will always show some remaining until then.

Currently there's no visual indication of whether it's done importing, or just stalled.

src/formats/onenote.ts
```diff
@@ -0,0 +1,444 @@
...
+	attachmentQueue: FileAttachment[] = [];
+	selectedSections: OnenoteSection[] = [];
+
+	graphHelper: MicrosoftGraphHelper = new MicrosoftGraphHelper();
+
+	init() {
+		this.addOutputLocationSetting('OneNote');
+		this.showUI();
+		// Required for the OAuth sign in flow
+		this.modal.plugin.registerObsidianProtocolHandler('importer-onenote-signin', (data) => {
```
tgrosinger, line 29: There's currently an issue if you:

1. Open the importer UI
2. Select OneNote as the file format
3. Change the file format to something else
4. Change it back to OneNote

Note the error in the console, as it attempts to register the protocolHandler again even though it has already been registered once.

One solution to this would be to use the `unregister` callback that is provided when calling `register`, although I am not sure when it should be called.

@lishid or @ericaxu do you have any suggestions?

tgrosinger: Clicking "Sign in" also does not work after following those steps.

tgrosinger (changes requested):
I successfully tested this and was able to import some OneNote notebooks. It's looking really good and getting close I think. Just a few more comments and a couple issues I found while testing.

Thank you!

p3rid0t:
@tgrosinger All your comments and issues got fixed in the latest commit.
 
Once you're done with reviewing, please message me on Discord (@p3rid0t), so we can set up the Azure AD project, as it's required for using school or work accounts with this importer (the current project is marked as development and is unverified)

tgrosinger (approved):
This looks great to me, thank you!

src/formats/onenote.ts
```diff
@@ -0,0 +1,467 @@
...
+		
+				// Replace the <object> tag with the new <p> element
+				object.parentNode?.replaceChild(markdownLink, object);
+			}
+		});
+
+		images.forEach(async (image) => {
+			let split: string[] = image.getAttribute('data-fullres-src-type')!.split('/');
+			const extension: string = split[1];
+			// TODO: there may be a similar function in the Obsidian API but I couldn't find it
```
lishid, line 329: We should be able to leverage moment for this?

src/formats/onenote.ts
```diff
@@ -0,0 +1,467 @@
...
+				video.parentNode?.replaceChild(linkNode, video);
+			}
+		});
+		return pageElement;
+	}
+
+	// Downloads attachments from the attachmentQueue once the file has been created.
+	async fetchAttachmentQueue(progress: ImportContext, currentFile: TFile) {
+		if (this.attachmentQueue.length >= 1) {
+			let attachmentPath: string = (await this.getOutputFolder())!.path + '/OneNote Attachments';
```
lishid, line 361: this.getOutputFolder() should only be called once and validated before the importer starts it work - then the referenced TFolder should be passed.

src/formats/onenote.ts
```diff
@@ -0,0 +1,467 @@
...
+			if (matchingStyle) {
+				const newElementTag = styleMap[matchingStyle];
+				const newElement = document.createElement(newElementTag);
+
+				if (newElementTag === 'pre') {
+					const code = newElement.createEl('code');
+					code.textContent = element.textContent;
+				}
+
+				else newElement.innerHTML = element.innerHTML;
```
lishid, line 422: This seems redundant? Why not just skip the replacement if we are keeping the same innerHTML?

src/formats/onenote.ts
```diff
@@ -0,0 +1,467 @@
...
+				element.replaceWith(newElement);
+			}
+		});
+
+		return pageElement;
+	}
+	
+	/* Commented out for possible future use. As of now, it seems like it's htmlToMarkdown's fault rather than OneNote's
```
lishid, line 430: Can you explain what's wrong here? Someone reading this code has no idea what this is intended to fix or what's wrong with htmlToMarkdown.

lishid (commented):
- Can you run the auto formatter? Seems like there's a few things that aren't formatted properly.
- I'm seeing a lot of `!`s, without checking for nulls. Are we certain they will not be null/undefined? If we aren't, proper checks should be added.
- I'm also seeing a lot of `?`s that shouldn't be needed since the exact variable was used previously without a `?` check and would have errored out. Can we clean those up and only leave behind the ones that actually could be nullable?
- I'd like us to use `for-of` loops instead of `array.forEach()` - they're mostly equivalent so this should be a simple change.

lishid:
In addition to my comments, it would be nice if we can go over the comments from Tony and resolve them if they are already fixed in code, and address any unresolved ones (or at least respond why not).

src/formats/onenote/graph-helper.ts
```diff
@@ -0,0 +1,86 @@
+import { ObsidianProtocolData, requestUrl } from 'obsidian';
+import { accessTokenResponse } from './models/tokenResponse';
+
+const GRAPH_CLIENT_ID: string = 'c1a20926-78a8-47c8-a2a4-650e482bd8d2'; // TODO: replace with an Obsidian team owned client_Id
+const GRAPH_SCOPES: string[] = ['user.read', 'notes.read'];
+const GRAPH_DATA = {
+	state: crypto.randomUUID(),
```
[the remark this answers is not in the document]
tgrosinger: @p3rid0t, Sorry to lead you astray here. Thanks @lishid.

src/formats/onenote/graph-helper.ts
```diff
@@ -0,0 +1,94 @@
...
+const GRAPH_CLIENT_ID: string = 'c1a20926-78a8-47c8-a2a4-650e482bd8d2'; // TODO: replace with an Obsidian team owned client_Id
+const GRAPH_SCOPES: string[] = ['user.read', 'notes.read'];
+const REDIRECT_URI: string = 'obsidian://importer-onenote-signin/';
+
+export class MicrosoftGraphHelper {
+	accessToken: string = '';
+	state: string = this.generateRandomState(32);
+
+	// Generate state for OAuth authentication
+	generateRandomState(length: number): string {
```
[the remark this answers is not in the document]
lishid: Sorry I missed this. The correct function to use is `genUid` from `util.ts`

lishid:
> Would it be okay for me to resolve the comments which got replied to or fixed?

Yes feel free to mark them as resolved!

src/formats/onenote.ts
tgrosinger: This was marked as resolved, but I don't see any code changes.

src/formats/onenote.ts
```diff
@@ -0,0 +1,467 @@
...
+	selectedSections: OnenoteSection[] = [];
+
+	graphHelper: MicrosoftGraphHelper = new MicrosoftGraphHelper();
+
+	init() {
+		this.addOutputLocationSetting('OneNote');
+		this.showUI();
+		// Required for the OAuth sign in flow
+		try {
+			this.modal.plugin.registerObsidianProtocolHandler('importer-onenote-signin', (data) => {
```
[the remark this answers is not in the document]
tgrosinger: @p3rid0t, I just wanted to bump this comment. Let us know if you have any questions about this change.

tgrosinger:
@p3rid0t, let us know if you have any questions about changing how the protocol handler is registered ([this comment](https://github.com/obsidianmd/obsidian-importer/pull/51/files#r1311928111)). Thanks!

PR #56, opened 2023-08-11
Readded Keep testing files

daledesilva:
This pull request only adds an unaltered Google Keep export file for future testing and 2 jsons that don't match the google keep data structure.

### DIFF
--- a/tests/keep/broken-json-note.json
+++ b/tests/keep/broken-json-note.json
@@ -1,0 +1,10 @@
+{
+    "color": "DEFAULT",
+    "isTrashed": false,
+    "isPinned": false,
+    "isArchived": false,
+    "textContent": "This is just a plain text note.     This sentence has more than one space before it.\n\n\nThis sentence has more than one new line before it.\n\n   This sentence has more than one new line and spaces before it.",
+    "title": "_Text note"
+    "userEditedTimestampUsec": 1690597779005000,
+    "createdTimestampUsec": 1690426307496000
+}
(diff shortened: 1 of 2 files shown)


daledesilva:
Closed the previous PR related to this and created a new one because I wasn't struggling with some merge issues.

tests/keep/broken-json-note.json
```diff
@@ -0,0 +1,10 @@
+{
+    "color": "DEFAULT",
+    "isTrashed": false,
+    "isPinned": false,
+    "isArchived": false,
+    "textContent": "This is just a plain text note.     This sentence has more than one space before it.\n\n\nThis sentence has more than one new line before it.\n\n   This sentence has more than one new line and spaces before it.",
+    "title": "_Text note"
```
lishid, line 7: I don't think we need to test whether the JSON parser can detect broken files. I think the actual exported zip file is good enough.

daledesilva: You wouldn't wanna know if it can break with unexpected input?

lishid: I don't think we need to test the browser implementation of JSON.parse

daledesilva: Not for that, but to Test that the keep format-importer code is correctly try/catching and reporting.
(For the broken file)

And the other file tests that it gracefully handles and reports a valid json that isn't an valid keep json.

In both cases it should show that it fails in the output log.

lishid: Oh ok, fair enough.

PR #63, opened 2023-08-11
Created textbundle file importer

arthurtyukayev:
This importer is similar to the Bear importer that file format just contains a zip of textbundles.

I was hoping for more code re-use, but the contents of the zip files are different.


src/formats/textbundle.ts
```diff
@@ -0,0 +1,77 @@
...
+		let folder = await this.getOutputFolder();
+		if (!folder) {
+			new Notice('Please select a location to export to.');
+			return;
+		}
+
+		let outputFolder = folder;
+
+		const attachmentsFolderPath = await this.createFolders(`${folder.path}/assets`);
+		const assetMatcher = /!\[\]\(assets\//g;
```
tgrosinger, line 28: Does Textbundle only support embedded assets? What about a link to a file (e.g. `[Photo](assets/photo.png)`)

src/formats/textbundle.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		for (let file of files) {
```
tgrosinger, line 30: Maybe we should check the `info.json` first and make sure `type` is `net.daringfireball.markdown` if it's provided (looks like it's an optional field)?

tgrosinger (commented):
Could you add the demo file from the TextBundle.org web page into the `tests/textbundle` directory?

Could we indicate in the UI that this supports version 1 and 2?

What do you think about adding support for TextPack as well? I think all we'd need to do is run it through `readZip` and then proceed as usual?

Thank you!

PR #85, opened 2023-08-31
Roam Importer

tgrosinger:
This is a continuation of #50.


README.md
```diff
@@ -6,31 +6,31 @@ This Obsidian plugin allows you to import notes from other apps and file formats
...
-- [x] Evernote `.enex`
-- [x] HTML, folder of files
-- [x] Notion, `.zip` of HTML files
-- [x] Bear `.bear2bk`
-- [x] Google Keep
-- [ ] Apple Notes
-- [ ] Microsoft OneNote
-- [ ] Roam Research
-- [ ] Other Markdown flavors
+-   [x] Evernote `.enex`
```
lishid, line 9: I don't think this should be reformatted.

src/formats/roam/roam_utils.ts
lishid, line 1: Should rename this to just `utils.ts`

src/formats/roam/models/roam-json.d.ts
```diff
@@ -0,0 +1,45 @@
+import { NodePickedFile } from 'filesystem';
+
+export interface RoamImportOptions {
+  saveAttachments: Boolean,
```
lishid, line 4: Should these use tabs instead?

lishid: Also inconsistency line termination between comma, semicolon, and nothing.

src/formats/roam-json.ts
```diff
@@ -0,0 +1,415 @@
...
+			const totalCount = markdownPages.size;
+			let index = 1;
+			for (const [filename, markdownOutput] of markdownPages.entries()) {
+				if (progress.isCancelled()) {
+					return;
+				}
+
+				try {
+					//create folders for nested pages [[some/nested/subfolder/page]]
+					await this.createFolders(path.dirname(filename));
```
lishid, line 120: Let's not use path.dirname since it doesn't work on mobile. Instead there's a parsing util function in filesystem.ts.

lishid (commented):
Initial quick pass, will look through the importing logic more thoroughly next.

src/formats/evernote-enex.ts
```diff
@@ -12,6 +12,7 @@ export class EvernoteEnexImporter extends FormatImporter {
 
 	async import(ctx: ImportContext) {
 		let { files } = this;
+		console.log(files)
```
tgrosinger, line 15: Oh that shouldn't be there.

README.md
tgrosinger: Oops

lishid (commented):
LGTM. If we can add progress tracking (for the progress bar) that would be great, but we can always do that later too.

lishid (approved):

Issue #83, opened 2023-08-30
Notion import error "Folder already exists"


Sinled:
Hi, there is an error with import plugin trying create folder that already exist, without any meaningful stack trace.

```
app.js:1 Uncaught (in promise) Error: Folder already exists.
    at t.<anonymous> (app.js:1:1024123)
    at app.js:1:236930
    at Object.next (app.js:1:237035)
    at a (app.js:1:235753)
```

I've downloaded and built plugin from code, and it seems problem is here in `format-importer.ts`
```ts
async createFolders(path: string): Promise<TFolder> {
    // ...
    let folder = this.vault.getAbstractFileByPath(normalizedPath);
    // ...
    await this.vault.createFolder(normalizedPath);
```

`getAbstractFileByPath` returns `null` if folder exist but have different case in name.
and when `createFolder` is trying to create folder but with different case - obsidian throws error.
I've replaced locally `getAbstractFileByPath` with `getAbstractFileByPathInsensitive` and after that it seems import works fine, but i'm not sure if this might break in some other place.

Also there is another bug - import won't do anything if notion export archive has nested archive with `*-Part-1.zip` in it.
In such case only one attachement folder with "part" archive in it created

lishid:
> Also there is another bug - import won't do anything if notion export archive has nested archive with *-Part-1.zip in it.
In such case only one attachement folder with "part" archive in it created

I've removed the requirement for detecting the notion ID from the zip file - that should fix this part.

joshuatazrein:
Would it be possible to message me a `.zip` of your data so I can work on debugging? You can share via Discord @joshua.palindrome or share with `joshuatreinier@gmail.com`

Sinled:
@joshuatazrein There are a lot of personal data, so i can't send it. I've tried to make smaller reproduction but for smaller export notion won't make nested archive. After that i've tried to make smaller archive with test data, but i cant find way to make archive nested directly in archive - mac os don't allow it, running `zip` from terminal on zip file also gives not result.

joshuatazrein:
Understandable! Yeah I'm having trouble finding nested zips because I'm on a Mac.

joshuatazrein:
I think the `createFolders` bug should be fixed in `main`, since that would potentially affect other plugins.

joshuatazrein:
[...]

Issue #82, opened 2023-08-29
Notion code blocks are imported without line breaks


smpb:
Greetings.
When using the importer on a Notion backup, code blocks that span multiple lines are erroneously recreated as a single line of text.

For example, this is a page with a couple of code blocks in Notion:
<img width="720" alt="Notion Screenshot" src="https://github.com/obsidianmd/obsidian-importer/assets/168636/062df8ab-7d0d-45c1-9e7d-6ee81a8ac659">

And this is the result of importing that same page to Obsidian:
<img width="720" alt="Obsidian Screenshot" src="https://github.com/obsidianmd/obsidian-importer/assets/168636/a73411aa-d4b7-405a-9ce4-7c1bf15f4a0d">

I believe this to be an incorrect behaviour since it invalidates the formatting and indentation of every single code block in a vault, and would greatly appreciate a way to have it behave the expected way.

wysstobi:
I reviewed the code two days ago since I ran into the same issue. 
The importer plugin replaces all newlines (\n) with <br />. This seems not to be a problem at first sight, but somehow <br /> tags inside of <code> tags are ignored by the function `htmltoMarkdown` which uses the Turndown Service. 

if you clone this repository and remove the function call on line 39 (`encodeNewlinesToBr(body);`) in the file `convert-to-md.ts` you will see this is not broken anymore. 

It might even be a work around for you, since I did not see a difference in the result by just removing this function call. 

To completely fix it, a more sophisticated version of this function could be implemented which doesn not replace new lines in code blocks. But I was not able to figure out a regular expression which works in all cases (especially when code blocks are nested in other elements as for examples in lists).

smpb:
> To completely fix it, a more sophisticated version of this function could be implemented which doesn not replace new lines in code blocks. But I was not able to figure out a regular expression which works in all cases (especially when code blocks are nested in other elements as for examples in lists).

Thanks for doing the groundwork and pointing me in the right direction.
I changed the `encodeNewLinestoBr` function to the following:

```diff
diff --git a/src/formats/notion/convert-to-md.ts b/src/formats/notion/convert-to-md.ts
index 3e7163e..4af7609 100644
--- a/src/formats/notion/convert-to-md.ts
+++ b/src/formats/notion/convert-to-md.ts
@@ -267,7 +267,7 @@ function replaceTableOfContents(body: HTMLElement) {
 }

 function encodeNewlinesToBr(body: HTMLElement) {
[...]

Issue #81, opened 2023-08-27
Equations are not imported correctly from Notion


grintroy:
When importing from Notion:

- Equations are not enclosed in `$...$` or `$$...$$`
- Overall, equations are duplicated in a following way: [visual representation][actual LaTeX][visual representation again].

For example, take a look at the following equation (screenshot from Notion)

<img width="333" alt="Screenshot 2023-08-27 at 13 30 17" src="https://github.com/obsidianmd/obsidian-importer/assets/70978482/edbaea5b-ac1a-462f-a5cf-03450e368e9a">

It is imported as

```latex
f(t)≈12A0+∑n=1∞Ancos⁡(nω0t+θn)f(t) \approx {1\over2} A_0 + \sum_{n=1}^{\infty} A_n \cos(n \omega_0 t + \theta _n)f(t)≈21​A0​+n=1∑∞​An​cos(nω0​t+θn​)​​
```

Note that the Importer placed the actual LaTeX `f(t) \approx {1\over2} A_0 + \sum_{n=1}^{\infty} A_n \cos(n \omega_0 t + \theta _n)` in between the visual representation of the formula in plaintext `f(t)≈12A0+∑n=1∞Ancos⁡(nω0t+θn)`.

This happens to all formulas imported to Obsidian.

wysstobi:
A possible approach could be to replace  all '.katex' elements with just the content of the corresponding `<annotation>` tag. This content could then be wrapped with `$` signs.

Issue #79, opened 2023-08-24
Importer not correctly handling Notion Teamspace update


pjbakker:
With Teamspaces everything in your export has an extra folder layer.

When importing an 'HTML export' the importer will only import 1 file with no actual data appearing after import.

To solve this manually:

* extract the zip file
* Compress all the folders in the extracted zip separately and import those (import work well on the teamspaces themselves)

Preferred solution: The Notion importer plugin correctly handles the additional teamspace layer that was introduced by Notion

joshuatazrein:
In order to debug this I need a Teamspace export. Can you attach one to this issue if possible?

pjbakker:
@joshuatazrein I didn't have time yet to create a custom Notion account export for debugging purposes for you.

The export has over a 1000 pages. When I import the ZIP returned from Notion export, I expect a full import, but I get this as a result:

<img width="550" alt="image" src="https://github.com/obsidianmd/obsidian-importer/assets/1267780/afd1e856-ac36-46d0-9d4c-97116cc37665">

When I extract the ZIP I get this folder structure:

<img width="512" alt="image" src="https://github.com/obsidianmd/obsidian-importer/assets/1267780/7c682c15-8b3e-467d-b9d6-507235d9b9c7">

And if I re-zip each of these folders:

<img width="521" alt="image" src="https://github.com/obsidianmd/obsidian-importer/assets/1267780/00d55b0f-f225-47e5-8b93-e8bd939a70c7">

If I import those ZIP files individually, it works as expected. All my 1000+ pages are there. There are some failures but I'm not sure if that's always the case.

<img width="546" alt="image" src="https://github.com/obsidianmd/obsidian-importer/assets/1267780/7346d6a3-467d-476b-9435-147cd169c9ad">

joshuatazrein:
Thanks for these screenshots. The issue seems to be that the `Export-` etc. file is somehow inside the `.zip` your are importing, which shouldn't be the case in any Notion export. Try importing the `Export` file directly and see if it works. I've never experienced an issue with Notion importing the root `Export` as an attachment—it might be because I'm on MacOS? Not sure.

pjbakker:
W.r.t. the failed files, I cannot see the full names (they are abbreviated with '...').

I think that's caused because I import into a new folder, but attachments get 'imported' to the root, so the already exist as I now re-import the same pages to a subfolder I can delete.

joshuatazrein:
Ok, sounds good.

[...]

PR #90, opened 2023-09-06
Notion Bug Squashing

joshuatazrein:
- [ ]  [Notion import error "Folder already exists" #83](https://github.com/obsidianmd/obsidian-importer/issues/83)
    - It seems like the change should be incorporated by the Obsidian team in `main`. Let me know how that sounds.
- [x] [Equations are not imported correctly from Notion #81](https://github.com/obsidianmd/obsidian-importer/issues/81)
- [x] [Importer not correctly handling Notion Teamspace update #79](https://github.com/obsidianmd/obsidian-importer/issues/79)
    - I tried this with a Teamspace export and it worked perfectly. Let me know if there is an edge case I missed.
- [x] [Notion code blocks are imported without line breaks #82](https://github.com/obsidianmd/obsidian-importer/issues/82)
- [x] [Incorrect formatting of Notion bookmark #76](https://github.com/obsidianmd/obsidian-importer/issues/76)
- [x] [Notion import corrupted when page includes "Toggle Heading" #75](https://github.com/obsidianmd/obsidian-importer/issues/75)


src/formats/notion/convert-to-md.ts
```diff
@@ -283,16 +315,51 @@ function fixNotionDates(body: HTMLElement) {
...
+					return
+				}
+			}
+		}
+		
+	})
+}
+
+function replaceElementsWithChildren(body: HTMLElement, selector: string) {
+	let element = body.querySelector(selector)
```
lishid, line 344: Why not use findAll?

src/formats/notion/convert-to-md.ts
```diff
@@ -193,6 +200,28 @@ function fixDoubleBackslash(markdownBody: string) {
 	return markdownBody;
 }
 
+function fixEquations(body: HTMLElement) {
+	const katexEls = body.findAll('.katex')
+	katexEls.forEach(katex => {
```
lishid, line 205: Can we convert all of these forEach to for-of loops instead?

src/formats/notion/convert-to-md.ts
```diff
@@ -193,6 +200,28 @@ function fixDoubleBackslash(markdownBody: string) {
...
+		const annotation = katex.querySelector('annotation')
```
lishid, line 206: Prefer to use `.find` over `.querySelector`.

lishid (commented):
Can you run the auto formatter? There's some missing semicolons.

src/formats/notion/convert-to-md.ts
joshuatazrein: Searching for the first element in a `while` loop ensured that the element hasn't been removed in a previous iteration (i.e. nested children etc). Was having trouble with the `findAll` function

lishid: Hmmm it seems that we are only removing the exact element being found while all nested children are reattached to the parent, so there shouldn't be any elements that gets removed from a previous iteration, no?

joshuatazrein: I can test using `findAll`.

src/formats/notion/convert-to-md.ts
```diff
@@ -193,6 +200,28 @@ function fixDoubleBackslash(markdownBody: string) {
 	return markdownBody;
 }
 
+function fixEquations(body: HTMLElement) {
+	const katexEls = body.findAll('.katex');
+	for (let katex of katexEls) {
+		const annotation = katex.find('annotation');
+		if (!annotation) return;
```
lishid, line 207: This should be `continue`.

src/formats/notion/convert-to-md.ts
```diff
@@ -283,16 +315,51 @@ function fixNotionDates(body: HTMLElement) {
...
+		}
+		
+	});
+}
+
+function replaceElementsWithChildren(body: HTMLElement, selector: string) {
+	let els = body.findAll(selector);
+	for (let el of els) {
+		const children = el.children;
+		const childNodes: Element[] = [];
```
lishid, line 348: Would `Array.from(el.children)` here simplify things?

lishid (approved):
There's a couple of forEach to change (and be careful of `return` statements) but otherwise this looks great!

PR #91, opened 2023-09-06
OneNote importer improvements

tgrosinger:
This change moves where the OneNote protocol handler is registered and then runs the formatter.


tgrosinger:
Pushed some changes so that I can pull them down in my VM and test.

- [ ] Test changes

lishid (approved):

PR #108, opened 2023-09-16
Notion: Fix codeblocks being imported without line breaks

### LINKED ISSUE #107: Notion code blocks still being imported without line breaks
Hey there! Notion import has been the main roadblock for me to switch to Obsidian, so thanks for adding this functionality! I tried to import my workspace today, however I'm still facing the issue first reported in #82. I confirmed that the version of the plugin being used is 1.3.1, which should have the fix, however all my code blocks are still ending up without any line breaks (And my notion is absolutely full of them). Let me know if I can provide anything to help debug this further!

image

_Originally posted by @aksh1618 in https://github.com/obsidianmd/obsidian-importer/issues/82#issuecomment-1717428382_

aksh1618:
Fixes #107. (See https://github.com/obsidianmd/obsidian-importer/issues/107#issuecomment-1722272675)

### DIFF
--- a/src/formats/notion/convert-to-md.ts
+++ b/src/formats/notion/convert-to-md.ts
@@ -287,10 +287,14 @@ function replaceTableOfContents(body: HTMLElement) {
 }
 
 function encodeNewlinesToBr(body: HTMLElement) {
-	body.innerHTML = body.innerHTML.replace(/\n/g, '<br />');
-	// Since <br /> is ignored in codeblocks, we replace with newlines
+	// Since <br /> is ignored in codeblocks, we preserve newlines inside codeblocks with a placeholder.
 	for (const block of body.findAll('code')) {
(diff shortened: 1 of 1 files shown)


lishid:
Can we instead make the regex a bit smarter?

Instead of `/<br \/>/g` we can make it `/<br ?\/?>/g`

aksh1618:
Yes, we could do that but I went for a placeholder, as a code block could already contain `<br />`, for instance some simple html code block, which would then be replaced along with the ones that we add. So we'll end up changing the actual contents in that case.

lishid:
You're right, but I don't suppose there's a chance `<` and `>` could be introduced into innerHTML without being properly converted to entities format. Either way though, it seems that there's an even better fix from joshua, so I'll close this in favor of that one: https://github.com/obsidianmd/obsidian-importer/pull/118

PR #121, opened 2023-09-25
notion file time 

sayzard:
When importing Notion, it extracts the time from the HTML and uses it as ctime and mtime of the MD file.

### DIFF
--- a/src/formats/notion.ts
+++ b/src/formats/notion.ts
@@ -1,4 +1,4 @@
-import { normalizePath, Notice, Setting } from 'obsidian';
+import { normalizePath, Notice, Setting, DataWriteOptions } from 'obsidian';
 import { PickedFile } from '../filesystem';
 import { FormatImporter } from '../format-importer';
 import { ImportContext } from '../main';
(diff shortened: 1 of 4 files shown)


src/formats/notion.ts
```diff
@@ -102,9 +102,13 @@ export class NotionImporter extends FormatImporter {
 					ctx.status(`Importing note ${fileInfo.title}`);
 
 					const markdownBody = await readToMarkdown(info, file);
+					const writeOptions: DataWriteOptions = {
+						ctime: fileInfo.ctime ? fileInfo.ctime.getTime() : Date.now(),
+						mtime: fileInfo.ctime ? fileInfo.ctime.getTime() : Date.now(),
```
lishid, line 107: How come `mtime` is also set to `ctime`?

src/formats/notion.ts
```diff
@@ -102,9 +102,13 @@ export class NotionImporter extends FormatImporter {
 					ctx.status(`Importing note ${fileInfo.title}`);
 
 					const markdownBody = await readToMarkdown(info, file);
+					const writeOptions: DataWriteOptions = {
+						ctime: fileInfo.ctime ? fileInfo.ctime.getTime() : Date.now(),
```
lishid, line 106: I would prefer to not set this at all if we don't have this information (instead of forcefully setting it to `Date.now()`).

src/formats/notion.ts
sayzard: If I import Notion zip file with the current version, both ctime and mtime are the current time. So it will not be able to sort in the Dataview. Just in case, even if fs uses ctime as the current time, I can sort with mtime.

src/formats/notion.ts
sayzard: The result of the current version of the behavior is that ctime and mtime will be the current time.

src/formats/notion.ts
lishid: I'm hesitant to set the last modified time to be the file's original creation time. I am assuming this information is not available anywhere in the export?

sayzard: Typically, most users don't use the last edited time property, but I've changed the code to use the last edited time if it exists.
<tr class="property-row property-row-created_time"> <tr class="property-row property-row-created_time
<tr class="property-row property-row-last_edited_time"> <tr class="property-row property-row-last_edited_time

PR #125, opened 2023-10-08
Add Apple Notes support

### LINKED ISSUE #15: Import from Apple Notes - $2,500
Please upvote this issue if you would like to see Apple Notes as an import format.

## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $2,500 USD
**Timeframe:** 21 days

Please only apply if you have taken time to explore the Importer codebase, as well as existing options such as [Apple Notes Liberator](https://github.com/HamburgChimps/apple-notes-liberator) and [Apple Cloud Notes Parser](https://github.com/threeplanetssoftware/apple_cloud_notes_parser). Provide detail about how you would approach solving it within the constraints of this Obsidian plugin.

## Requirements

- [ ] Must work on macOS — iOS is nice to have but not required for initial version
- [ ] Properly converts files to [Obsidian-flavored Markdown](https://help.obsidian.md/Editing+and+formatting/Obsidian+Flavored+Markdown), including tables, to-do lists, etc
- [ ] Support for images and attachments. Embed links converted to Markdown format `!()[image.png]` and placed in the user's defined attachment location (Settings → File & links)
- [ ] Supports internal links added in iOS17

## Current solutions

- [Official guide](https://help.obsidian.md/import/apple-notes)
- https://github.com/threeplanetssoftware/apple_cloud_notes_parser
- https://github.com/HamburgChimps/apple-notes-liberator

mirnovov:
This PR adds initial Apple Notes support, which closes #15.

Apple Notes stores its files in an SQLite database. This importer reads the database, and copies the data to a physical folder on disk, starting with the folders, and then the files and any attachments associated with them. 

- The method of authorising access to the database is a bit different from what was initially discussed. Instead of asking for Full Disk Access, I have found a way to ask for access to a specific folder using the file picker, as documented [here](https://developer.apple.com/videos/play/wwdc2019/701/). This is both more secure and more intuitive from the initial concept.
- As discussed in the issue, currently the plugin only works on macOS. On other platforms, the "Apple Notes" option is accessible, but it shows some explanatory text instead of allowing the import. I feel this is preferable as it makes the user aware that the capability exists (i.e. if they find it on iOS but sync the same vault to Mac) but if it's not desirable it should be easy to simply make it only appear on Mac.

### Requirements

- [x] Must work on macOS — iOS is nice to have but not required for initial version
- [x] Properly converts files to [Obsidian-flavored Markdown](https://help.obsidian.md/Editing+and+formatting/Obsidian+Flavored+Markdown), including tables, to-do lists, etc
- [x] Support for images and attachments. Embed links converted to Markdown format !()[image.png] and placed in the user's defined attachment location (Settings → File & links)
- [x] Supports internal links added in iOS17

### Features

- Basic formatting: strikethrough, italic, bold, underline, sub/superscript, lists, tables, etc.
- More advanced formatting: colour, font, font size, text alignment, blockquote, checklist, etc. 
- Support for both external links and internal links (new in Sonoma and iOS 17).
- By default, removes the title as Apple Notes uses this as the note name, but this can be disabled.
- Supports embedding files: images, video, audio, PDF (also new in Sonoma and iOS 17), drawings (saved as PNG), etc
- Probably more than I forgot, but in general this supports a lot more stuff than most extant solutions.

### Dependencies

The production version of `main.js` comes in at ~380 KB when compiled.

This includes a number of new dependencies (though an effort has been made to keep them minimal):
- `protobufjs` and `pako` for decoding Apple Notes' native protocol-buffer-encoded note format
- `sqlite-tag-spawned` for reading the Notes database. This avoids bundling SQLite in WASM, which saves a lot of size.

The protobuf definition is based off the [impressive work](https://github.com/threeplanetssoftware/apple_cloud_notes_parser) done by Ciofeca Forensics.

### Testing

This does not currently include any tests, as the notes which I have used to manually test it contain personal information. However, it should be easy for me to create these later in the PR process.

**I have read the CLA agreement and I hereby sign the CLA.**


kepano:
Exciting! We will review and share feedback soon!

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,385 @@
...
+				
+				const destIdentifier = row.ZTOKENCONTENTIDENTIFIER.match(NOTE_URI)[1];
+				
+				row = await this.importer.database.get`
+					SELECT z_pk FROM ziccloudsyncingobject 
+					WHERE zidentifier = ${destIdentifier.toUpperCase()}`;
+					
+				if (!(row.Z_PK in this.importer.resolvedFiles)) {
+					//if we don't have the note yet, we fetch it early to get its path
+					this.importer.resolveNote(row.Z_PK);
```
tgrosinger, line 285: Does this work as expected if two notes reference each other forming a cycle?

tgrosinger (changes requested):
Didn't quite finish reviewing, but I wanted to send off what I have so far.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+		}
+		else if (!this.importTrashed && folder.ZFOLDERTYPE == ANFolderType.Trash) {
+			this.trashedFolder = id;
+			return;	
+		}
+		else if (folder.ZPARENT !== null) {
+			this.resolveFolder(folder.ZPARENT);
+			prefix = this.resolvedFiles[folder.ZPARENT].path + '/';
+		}
+		else if (this.multiAccount) {
```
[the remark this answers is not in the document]
mirnovov: If the folder has a parent, the account prefix is already added to the parent.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+			FROM ziccloudsyncingobject
+			WHERE z_ent = 14 AND z_pk = ${id}
+		`;
+		let prefix;
+		
+		if (folder.ZFOLDERTYPE == ANFolderType.Smart) {
+			return;	
+		}
+		else if (!this.importTrashed && folder.ZFOLDERTYPE == ANFolderType.Trash) {
+			this.trashedFolder = id;
```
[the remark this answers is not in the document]
mirnovov: The former. I've changed the variable name to be clearer.

.gitignore
```diff
@@ -5,6 +5,9 @@
 *.iml
 .idea
 
+# Panic Nova
+.nova
```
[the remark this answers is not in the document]
mirnovov: I use Nova as my editor; if there is VSCode and other editors in the gitignore, it seems fair to add this as well. However, I can use a local gitignore instead if you want.

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 Obsidian
```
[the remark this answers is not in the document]
mirnovov: The protobuf definition (in `src/formats/apple-notes/descriptor.ts`) is based of prior work.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+			}
+		}
+	}
+	
+	async resolveAccount(id: number): Promise<void> {
+		if (!this.multiAccount && Object.keys(this.resolvedAccounts).length) {
+			this.multiAccount = true;
+		}
+		
+		const account = await this.database.get`
```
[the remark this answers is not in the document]
mirnovov: Sometimes, the folder may be in the database before its parent (I would guess this occurs if the parent is created later and then the folder is made a child of the parent afterwards). In that case we only have the number in the `ZPARENT` column of the child folder's row to generate it.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+				WHERE
+					a.z_ent = 10 
+					AND a.z_pk = ${id} 
+					AND a.z_pk = b.zmedia
+			`;
+			
+			const account = row.ZACCOUNT6 || row.ZACCOUNT5;
+			sourcePath = path.join(
+				os.homedir(), NOTE_FOLDER_PATH, 'Accounts', this.resolvedAccounts[account].uuid, 
+				'Media', row.ZIDENTIFIER, row.ZGENERATION1 || '', row.ZFILENAME
```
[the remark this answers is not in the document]
mirnovov: Incorrect, I tested it. It does do `identifier/filename` as intended.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+			return;
+		}
+		
+		const value = this.decodeData<ANDocument>(row.zhexdata).note;		
+		const folder = this.resolvedFiles[row.ZFOLDER] as TFolder;
+		const title = `${row.ZTITLE1}.md`;
+		
+		this.resolvedFiles[id] = await this.saveAsMarkdownFile(folder, title, ''); 
+		
+		//notes may reference other notes, so we want them in resolvedFiles before we parse to avoid cycles
```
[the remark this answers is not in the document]
mirnovov: They are added to resolvedFiles before they are parsed (by creating a `NoteConverter`). What I mean here is that since note references are checked against resolvedFiles, a note that references itself (or a note that references another note, which references the first note) won't create a cycle

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+				WHERE
+					z_ent = 4 
+					AND z_pk = ${id} 
+			`;
+			
+			sourcePath = path.join(
+				os.homedir(), NOTE_FOLDER_PATH, 'Accounts', this.resolvedAccounts[row.ZACCOUNT1].uuid, 
+				'FallbackImages', row.ZIDENTIFIER, row.ZFALLBACKIMAGEGENERATION || '', 'FallbackImage.png'
+			);
+			outName = 'Drawing', outExt = 'png';
```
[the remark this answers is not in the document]
mirnovov: They don't, correct.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,385 @@
...
+				const val = this.convertAlign(attr?.paragraphStyle?.alignment);
+				prefix += `\n<p style="text-align:${val};margin:0">`;
+			}
+		}
+		
+		return prefix;
+	}
+	
+	formatHtmlAttr(attr: ANAttributeRun): string {
+		if (attr.strikethrough) attr.fragment = `<s>${attr.fragment}</s>`;
```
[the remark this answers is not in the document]
mirnovov: In Live Preview, doing something like `a<sub>~~b~~</sub>` doesn't work (at least on my end). So if there are things that can only be done in HTML tags what's inside them is done in HTML tags too.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,385 @@
...
+			let nextIsSame = true;
+			
+			/* First, merge tokens with the same attributes */
+			while (nextIsSame) {
+				attr = this.note.attributeRun[i];
+				offsetEnd = offsetEnd + attr.length;
+				attrText += this.note.noteText.substring(offsetStart, offsetEnd);
+				
+				offsetStart = offsetEnd;
+				nextIsSame = attrEquals(attr, this.note.attributeRun[i+1]);
```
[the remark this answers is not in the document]
mirnovov: If it runs past the end of `attributeRun` it's `undefined` which is accounted for.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,385 @@
...
+		for (let j = 0; j < fragments.length; j++) {		
+			let {attr, fragment} = fragments[j];
+			
+			if (firstLineSkip) {
+				if (fragment.contains('\n') || attr.attachmentInfo) firstLineSkip = false;
+				else continue;
+			}
+
+			attr.fragment = fragment;		
+			attr.atLineStart = fragments[j - 1]?.fragment.contains('\n') || j == 0;
```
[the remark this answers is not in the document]
mirnovov: see above.

src/formats/apple-notes/convert-note.ts
mirnovov: Yes, it has been designed to avoid that (per my earlier comments, the note is intentionally added to `resolvedFiles` before its content is parsed)

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,385 @@
...
+		this.app = importer.app;
+	}
+
+	*parseTokens(): IterableIterator<ANFragmentPair> {	
+		let i = 0;
+		let offsetStart = 0;
+		let offsetEnd = 0;
+		
+		while (i < this.note.attributeRun.length) {
+			let attr!: ANAttributeRun;
```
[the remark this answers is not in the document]
mirnovov: Yes, TS isn't smart enough to know that `attr` will inevitably be assigned before it is used later on.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,303 @@
...
+		this.resolvedAccounts[id] = {
+			name: account.ZNAME,
+			uuid: account.ZIDENTIFIER
+		};
+	}
+	
+	async resolveFolder(id: number): Promise<void> {
+		if (id in this.resolvedFiles) return;
+		
+		const folder = await this.database.get`
```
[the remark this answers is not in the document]
mirnovov: While accounts aren't referenced elsewhere like notes and folders this may change, so IMO makes sense to keep it consistent.

src/formats/apple-notes/convert-note.ts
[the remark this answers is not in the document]
tgrosinger: Could that be resolved if the loop below is changed to a do..while loop?

src/formats/apple-notes/convert-note.ts
```diff
@@ -47,7 +47,10 @@ export class NoteConverter {
 				attrText += this.note.noteText.substring(offsetStart, offsetEnd);
 				
 				offsetStart = offsetEnd;
-				nextIsSame = attrEquals(attr, this.note.attributeRun[i+1]);
+				
+				if (i == this.note.attributeRun.length - 1) nextIsSame = false;
```
tgrosinger, line 51: Thanks for changing this. What do you think about stylistically tweaking it like this?

```
nextIsSame = (i == this.note.attributeRun.length - 1)
    ? false
    : attrEquals(attr, this.note.attributeRun[i+1]);
```

src/formats/apple-notes.ts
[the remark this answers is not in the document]
tgrosinger: Hmm okay, that makes sense. Thank you.

src/formats/apple-notes/convert-table.ts
```diff
@@ -0,0 +1,125 @@
...
+					cellColumns = object;
+					break;
+			}
+		}
+		
+		if (this.columnCount < 1 || this.rowCount < 1 || !cellColumns) return null;
+		return await this.computeCells(cellColumns!);
+	}
+	
+	findIndices(object: ANMergeableObjectEntry, rows: boolean): void {
```
tgrosinger, line 61: I'm having trouble following what's happening in this function without an understanding of the data model. Would it be possible to add a brief description to a couple of the methods in this class?

LICENSE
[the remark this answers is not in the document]
tgrosinger: Let's put the license that came with the protobuf definition in `descriptor.ts` instead please.

src/formats/apple-notes/convert-table.ts
mirnovov: I've tried to make it easier to follow by restructuring a lot of the code and adding a bunch of comments, but it's inherently not a very easy thing to understand.

src/formats/apple-notes/models.ts
```diff
@@ -0,0 +1,176 @@
+import { Message } from 'protobufjs';
+export { descriptor } from './descriptor';
```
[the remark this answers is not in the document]
mirnovov: I think that's cleaner (and makes sense since it's the definition that a lot of the models are based off), but I've removed it.

src/formats/apple-notes/convert-note.ts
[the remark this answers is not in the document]
mirnovov: Ofc it's your guys' call, but I would recommend against this. A bunch of my CS lecture notes that I used to test used subscript, superscript etc. for formatting; if that was stripped, it could change the meaning of things without my realising if I were to look at them later.

Outside the specialised cases like this, the vast majority of formatting for the notes I used to test ended up being in Markdown, so it's not like it is obtrusive. If it led to the notes being littered with HTML tags I wouldn't have done it.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,307 @@
...
+		
+		if (folder.ZFOLDERTYPE == ANFolderType.Smart) {
+			return;	
+		}
+		else if (!this.importTrashed && folder.ZFOLDERTYPE == ANFolderType.Trash) {
+			this.trashFolder = id;
+			return;	
+		}
+		else if (folder.ZPARENT !== null) {
+			this.resolveFolder(folder.ZPARENT);
```
liamcain, line 174: Missing `await` here. This caused the next line to fail in some cases when testing.

liamcain:
I also noticed that an error is thrown for some Apple Notes tables. Here's an example of a table that fails for me:

image

src/main.ts
```diff
@@ -234,6 +235,12 @@ export default class ImporterPlugin extends Plugin {
 				importer: OneNoteImporter,
 				helpPermalink: 'import/onenote',
 			},
+			'apple-notes': {
```
kepano, line 238: Move this to the top of the list (it should be alphabetical)

kepano:
Looking great so far! Import was really smooth.

Screenshot 2023-10-11 at 2 48 25 PM

Issues I came across:

- [ ] Apple Notes should be listed first in the Import modal, since the list is alphabetical
- [ ] If `Omit first line` is turned on there are two unnecessary line breaks at the top of every file that should be removed
- [ ] All "Scans" type notes failed — I see instead: `**(unknown attachment: com.apple.notes.gallery)**`
- [ ] Attachments should be placed in the user's attachment folder defined in the Obsidian Settings (see other importers), currently it is placed inside the Apple Notes folder

src/main.ts
mirnovov: Not sure about this, since it only works on Mac, and the one at the top is the first one everyone sees. Maybe we could move it to the bottom, so the ones that work on every platform are alphabetised nicely.

kepano: Bear is also macOS/iOS only, so I don't see it as a problem. Maybe in the future we can add a generic way for importers to list which platforms they work on, and make that a conditional display based on the user's current platform. Opened #128 for this. In the meantime we can keep the alphabetical order here.

mirnovov:
> I also noticed that an error is thrown for some Apple Notes tables. Here's an example of a table that fails for me:
> 
> image

The table that you posted works fine for me. Would you be able to give me what error you've getting?

mirnovov:
@kepano I've added support for scans (and fixed the other things you mentioned). Only issue is that unedited scans are stored as images, while edited ones are stored as PDFs. Would it be worth converting the unedited ones into PDF as well? Not sure how feasible this is given what Obsidian includes; an additional dependency definitely wouldn't be worth it IMO.

liamcain:
@mirnovov the error I get for that table I posted is "Cannot set properties of undefined (setting '2')"

kepano:
> @kepano I've added support for scans (and fixed the other things you mentioned). Only issue is that unedited scans are stored as images, while edited ones are stored as PDFs. Would it be worth converting the unedited ones into PDF as well?

Awesome! It doesn't seem necessary to convert files to PDF — we should minimally process things, so importing unedited scans as images seems best to me

liamcain:
@mirnovov Okay, even duplicating the note with the bad table doesn't cause the error to appear. But it happens every time that I follow these steps: 

https://github.com/obsidianmd/obsidian-importer/assets/693981/05371641-6604-4760-9518-e3e2f2652131

mirnovov:
Still can't reproduce this - even when testing on other computers with older versions of macOS. I think I've figured out what's happening and added something that should theoretically stop it from causing an error.

If it still doesn't work a possible solution could be to send me the encoded table directly, since that will be directly reproducible. If you apply [this patch](https://github.com/obsidianmd/obsidian-importer/files/12901448/table-debug.patch) and build the plugin it will print out the protobuf of every table in the dev tools console.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+		return prefix;
+	}
+	
+	/** Since putting markdown inside inline html tags is currentlyproblematic in Live Preview, this is a separate
+	 parser for those that is activated when HTML-only stuff (eg underline, font size) is needed */
+	formatHtmlAttr(attr: ANAttributeRun): string {
+		if (attr.strikethrough) attr.fragment = `<s>${attr.fragment}</s>`;
+		if (attr.underlined) attr.fragment = `<u>${attr.fragment}</u>`;
+		
+		if (attr?.superscript == ANBaseline.Super) attr.fragment = `<sup>${attr.fragment}</sup>`;
```
lishid, line 166: The `?` in `attr?` is not needed here and below.

src/formats/apple-notes/convert-note.ts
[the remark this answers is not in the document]
lishid: I'm ok leaving it like this for now. It doesn't impact users who are not using super/subscripts/undeliner/color/etc.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+		if (attr.strikethrough) attr.fragment = `~~${attr.fragment}~~`;
+		if (attr.link && attr.link != attr.fragment) attr.fragment = `[${attr.fragment}](${attr.link})`;
+		
+		if (attr.atLineStart) return this.formatParagraph(attr);
+		else return attr.fragment;
+	}
+	
+	formatParagraph(attr: ANAttributeRun): string {
+		const indent = '\t'.repeat(attr.paragraphStyle?.indentAmount || 0);
+		const styleType = attr?.paragraphStyle?.styleType;
```
lishid, line 227:
```suggestion
		const styleType = attr.paragraphStyle?.styleType;
```

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+		switch (alignment) {
+			default: return 'left';
+			case ANAlignment.Centre: return 'center';
+			case ANAlignment.Right: return 'right';
+			case ANAlignment.Justify: return 'justify';
+		}
+	}
+}
+
+function isBlockAttachment(attr: ANAttributeRun) {
```
lishid, line 377: `attr` is never null/undefined here, right? Which means we probably don't need the `?`s down below

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,346 @@
...
+const ROOT_DOC = Root.fromJSON(descriptor);
+
+export class AppleNotesImporter extends FormatImporter {
+	rootFolder: TFolder;
+	ctx: ImportContext;
+	attachmentPath: string;
+	
+	database: SQLiteTagSpawned;
+	resolvedAccounts: Record<number, ANAccount> = {};
+	resolvedFiles: Record<number, TAbstractFile | TFolder> = {};
```
lishid, line 26: I suggest breaking this up into one for TFile and one for TFolder so we aren't blindly casting to TFile all the time.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+					SELECT z_pk FROM ziccloudsyncingobject 
+					WHERE zidentifier = ${destIdentifier.toUpperCase()}`;
+					
+				if (!(row.Z_PK in this.importer.resolvedFiles)) {
+					//if we don't have the note yet, we fetch it early to get its path
+					this.importer.resolveNote(row.Z_PK);
+				}
+				
+				return this.app.fileManager.generateMarkdownLink(
+					this.importer.resolvedFiles[row.Z_PK] as TFile,
```
lishid, line 297: Should we check the result from indexing `resolvedFiles`?

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+		}
+		
+		if (!id) {
+			//doesn't have an associated file, so unknown
+			return ` **(unknown attachment: ${attr.attachmentInfo.typeUti})** `;
+		}
+		
+		await this.importer.resolveAttachment(id, attr.attachmentInfo.typeUti);
+		const link = this.app.fileManager.generateMarkdownLink(
+			this.importer.resolvedFiles[id] as TFile, '/'
```
lishid, line 351: Should we check the result from indexing `resolvedFiles`?

src/formats/apple-notes/convert-scan.ts
```diff
@@ -0,0 +1,40 @@
...
+			const imageUuid = object.customMap.mapEntry[0].value.stringValue;
+			
+			const row = await this.importer.database.get`
+				SELECT z_pk FROM ziccloudsyncingobject 
+				WHERE zidentifier = ${imageUuid}`;
+			
+			await this.importer.resolveAttachment(row.Z_PK, ANAttachment.Scan);
+			
+			links.push(this.importer.app.fileManager.generateMarkdownLink(
+				this.importer.resolvedFiles[row.Z_PK] as TFile, '/'
```
lishid, line 34: Also should check the result of indexing `resolvedFiles`.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+				id = row?.ZMEDIA;
+				break;
+		}
+		
+		if (!id) {
+			//doesn't have an associated file, so unknown
+			return ` **(unknown attachment: ${attr.attachmentInfo.typeUti})** `;
+		}
+		
+		await this.importer.resolveAttachment(id, attr.attachmentInfo.typeUti);
```
lishid, line 349: We should make `resolveAttachment` return the TFile, so we don't have to look it up again after.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,346 @@
...
+import { ungzip } from 'pako';
+import { Root } from 'protobufjs';
+import SQLiteTagSpawned from 'sqlite-tag-spawned';
+
+const NOTE_FOLDER_PATH = 'Library/Group Containers/group.com.apple.notes';
+const NOTE_DB = 'NoteStore.sqlite';
+/** Additional amount of seconds that Apple CoreTime datatypes start at, to convert them into Unix timestamps. */
+const CORETIME_OFFSET = 978307200; 
+
+const ROOT_DOC = Root.fromJSON(descriptor);
```
lishid, line 17: Can we move this to inside the importer, so we aren't initializing it for every other importer type?

lishid (commented):
Style nit: We prefer to have comments first letter capitalized and with a space before, like this:

```js
// This is a comment
```

src/formats/apple-notes/convert-note.ts
mirnovov: That's incorrect, attr is null at the end. But it makes sense to check for that explicitly, so I've changed it so it does that.

src/formats/apple-notes.ts
```diff
@@ -0,0 +1,346 @@
...
+				);
+				outName = 'Drawing';
+				outExt = 'png';
+				break;
+		}
+		
+		this.resolvedFiles[id] = await this.vault.createBinary(
+			//@ts-ignore
+			this.app.vault.getAvailablePath(`${this.attachmentPath}/${outName}`, outExt), 
+			fs.readFileSync(sourcePath),
```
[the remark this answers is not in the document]
mirnovov: Also changed it for file copying as well.

package.json
```diff
@@ -28,6 +29,9 @@
 	"dependencies": {
 		"@zip.js/zip.js": "2.7.20",
 		"joplin-turndown-plugin-gfm": "1.0.12",
+		"pako": "^2.1.0",
```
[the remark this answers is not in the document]
mirnovov: Thanks, this shaves quite a bit off the bundle size and makes it a bit faster (since it's doing the gunzipping in native code not JS I presume). Also it's good to have less dependencies.

src/formats/apple-notes/convert-note.ts
lishid: If it can be null then the typescript definition should have been `attr: ANAttributeRun | null`.

src/formats/apple-notes/convert-note.ts
```diff
@@ -0,0 +1,398 @@
...
+	formatMultiRun(attr: ANAttributeRun | null): string {
+		const styleType = attr?.paragraphStyle?.styleType;
+		let prefix = '';
+		
+		switch (this.multiRun) {
+			case ANMultiRun.List:
+				if (
+					(attr?.paragraphStyle?.indentAmount == 0 && 
+					!LIST_STYLES.includes(styleType!)) ||
+					isBlockAttachment(attr!)
```
lishid, line 116: Ah it seems like this is asserted to be not null here - it probably shouldn't?

lishid (commented):
Hmm is it possible to check and shortcut (return early) when attr is null at the beginning of a function? Especially for a function called early on.

Alternatively we can test for null before calling helper functions. That helps us explicitly deal with null values, rather than leaving everything to the `?` operators.

src/formats/apple-notes/convert-note.ts
mirnovov: thanks for catching this - iirc it was initially never going to be null and then I changed that later on but forgot to update here

mirnovov:
> Hmm is it possible to check and shortcut (return early) when attr is null at the beginning of a function? Especially for a function called early on.

In the case of `formatMultiRun` no since the logic still applies when attr is null. It's basically going through what to close (eg the end of a code block). Attr is null at the end of a file, so it still has to close stuff there.

lishid (approved):

lishid:
Merged, let's have some people test it!

lishid:
Quick follow-up - given that we are granted permission to read the file, is there any chance we can make it work without making a temporary copy of the database? I.e. directly read from the DB file.

mirnovov:
@lishid theoretically yes, but then Apple Notes could possibly modify the database while it's being read (SQLite allows multiple readers but one writer, and the plugin's connection is read-only).

lishid:
Ah ok, that makes sense, thanks for clarifying.

dbdennis:
This is phenomenal--kudos to all who are involved!  One question/suggestions.  Could there be a way to import only new Apple notes, after an initial pull?   If this has already been asked and discussed, apologies.

Issue #71, opened 2023-08-15
When using Bear import, header will be imported twice


jmrauen:
In Bear textbundle-format the first line will be used as the filename (correct) but it will be inserted into the obsidian note (wrong). 
Please cut the first line, because I get the header twice. Thanx!

![image](https://github.com/obsidianmd/obsidian-importer/assets/23209835/9b063b33-a53f-4557-b57e-8ecdaf0bdeda)

lishid:
Just to make sure I understand, the notes themselves have a heading that's identical from the note name, right? You're suggesting we should cut off the first heading from the note since we already show the file name in the UI.

jmrauen:
Yes, that's right. You only to cut off the first heading from the note.

timsearle:
+1 on this! Would agree that this is a bug rather than enhancement. The top line of a Bear note is always the title of a note, whereas the Importer is treating the top-line as the title and then repeating the top line in the note contents.

PR #127, opened 2023-10-11
Fix bear2bk imported headings

puffo:
This PR ensures that the first heading of markdown files imported from a bear2bk backup file is not duplicated in the obsidian .md content. Additionally, this checks-in a sample bear2bk to verify the functionality. 

The sample file also contains a few other use cases which are not yet explicitly handled, but should be easy to cover in another pull request. Examples include trashed, locked, and archived files from Bear.

Resolves https://github.com/obsidianmd/obsidian-importer/issues/71


src/formats/bear-bear2bk.ts
```diff
@@ -73,4 +74,15 @@ export class Bear2bkImporter extends FormatImporter {
 			});
 		}
 	}
+
+	private removeMarkdownHeader(mdContent: string): string {
+		const lines = mdContent.split('\n');
```
tgrosinger, line 79: I don't think we need to split the whole file into lines just to see if the first line starts with a `#`. What if we do something like this?

```js
if (mdContent.startsWith('# ')) {
  const idx = mdContent.indexOf('\n')
  mdContent = idx > 0
    ? mdContent.substring(idx + 1)
    : ''
}
```

We should also be careful to compare against `# ` instead of `#` so that we don't pick up a tag by accident.

tgrosinger (changes requested):

mirnovov:
IMO this should be an on-by-default setting (which is what I'm doing in #125), as some users don't have automatic note titles turned on.

src/formats/bear-bear2bk.ts
puffo: Really great feedback, thank you. I've applied the changes and verified them on the checked in test file.

tgrosinger (approved):

PR #136, opened 2023-10-17
Apple Notes: Support inline internal links

### LINKED ISSUE #133: Apple Notes: Internal links not matched to relevant note
When importing Apple Notes, internal links (introduced in iOS 17) appear as external links with the format

`[Note name](applenotes:note/abc?ownerIdentifier=_xyz)`

These should instead appear as

`[[Note name]]`

mirnovov:
Closes #133 

Apple Notes has two kinds of internal links, one of which is stored as an attachment, and which has a little note icon. Its text always matches the name of the note. 

This PR adds support for the other, which is stored using the same mechanism as external links. In the screenshot below, the first link is the new type and the second one is the attachment type.  
Screenshot 2023-10-18 at 11 39 10 AM

### DIFF
--- a/src/formats/apple-notes/convert-note.ts
+++ b/src/formats/apple-notes/convert-note.ts
@@ -7,7 +7,7 @@ import { TableConverter } from './convert-table';
 import { AppleNotesImporter } from '../apple-notes';
 
 const FRAGMENT_SPLIT = /(^\s+|(?:\s+)?\n(?:\s+)?|\s+$)/;
-const NOTE_URI = /applenotes:note\/(.*)\?/;
+const NOTE_URI = /applenotes:note\/([-0-9a-f]+)(?:\?ownerIdentifier=.*)?/;
 
 const DEFAULT_EMOJI = '.AppleColorEmojiUI';
 const LIST_STYLES = [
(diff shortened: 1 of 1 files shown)


src/formats/apple-notes/convert-note.ts
```diff
@@ -346,6 +342,23 @@ export class NoteConverter extends ANConverter {
 		return `\n${this.app.fileManager.generateMarkdownLink(attachment, '/')}\n`;	
 	}
 	
+	async getInternalLinkFromUri(uri: string): Promise<string> {
+		return this.getInternalLink(uri);
+	}
+	
+	async getInternalLink(uri: string): Promise<string> {
+		const identifier = uri.match(NOTE_URI)[1];
```
kepano, line 350: This throws a build error `Object is possibly 'null'.`

src/formats/apple-notes/convert-note.ts
```diff
@@ -346,6 +342,23 @@ export class NoteConverter extends ANConverter {
 		return `\n${this.app.fileManager.generateMarkdownLink(attachment, '/')}\n`;	
 	}
 	
+	async getInternalLinkFromUri(uri: string): Promise<string> {
```
tgrosinger, line 345: Is this being used?

kepano:
Thanks! Looks good!

PR #137, opened 2023-10-17
Apple Notes: Fix older drawings, scans, and a few other things

### LINKED ISSUE #134: Apple Notes: Drawings not imported
Drawings in imported Apple Notes show the following error

`**(error reading attachment)**`

We should try to import/convert the drawing, or display a more useful error

### LINKED ISSUE #135: Apple Notes: Scans not imported
Scans do not appear to be imported from Apple Notes. In a previous version of #125 I was seeing an error but now the notes are empty and the scan files are not stored in Attachments

mirnovov:
Possibly fixes #134 and possibly fixes #135

Fixes the importer being unable to find the location of the drawing fallback image under certain circumstances (definitely pre-iOS 17 but a `ZGENERATION` may not be generated under other circumstances as well).

Fixes scans not being generated when the nicely cropped version of the image can't be found, by finding the camera photo and using that. (The cropped and flattened version may be stored somewhere else for older note versions but I never used the scan functionality before testing this, and the other Apple Notes importers just always use the camera photo so I can't see what they do. So short of me downgrading to an old macOS and iOS version I won't be able to determine that)

This also includes a few minor patches that didn't quite make it in #125

### DIFF
--- a/src/formats/apple-notes.ts
+++ b/src/formats/apple-notes.ts
@@ -136,6 +136,8 @@ export class AppleNotesImporter extends FormatImporter {
 				this.ctx.reportFailed(n.ZTITLE1, e?.message); 
 			}
 		}
+		
+		this.database.close();
 	}
 	
 	async resolveAccount(id: number): Promise<void> {
--- a/src/formats/apple-notes.ts
+++ b/src/formats/apple-notes.ts
@@ -290,11 +290,18 @@ export class AppleNotesImporter extends FormatImporter {
 						AND z_pk = ${id} 
 				`;
 				
-				const filename = row.ZFALLBACKIMAGEGENERATION ? 'FallbackImage.png' : `${row.ZIDENTIFIER}.jpg`;
-				sourcePath = path.join(
-					this.resolvedAccounts[row.ZACCOUNT1].path, 
-					'FallbackImages', row.ZIDENTIFIER, row.ZFALLBACKIMAGEGENERATION || '', filename
-				);
+				if (row.ZFALLBACKIMAGEGENERATION) {
+					// macOS 14/iOS 17 and above
+					sourcePath = path.join(
+						this.resolvedAccounts[row.ZACCOUNT1].path, 
+						'FallbackImages', row.ZIDENTIFIER, row.ZFALLBACKIMAGEGENERATION, 'FallbackImage.png'
+					);
+				}
+				else {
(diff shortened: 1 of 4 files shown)


kepano:
I can confirm that #135 is resolved — importing scans works for me now. 

However #134 is not solved under the conditions I described in the issue

> iOS 17.0.3 to create the drawing, then synced it to my Mac running macOS 13.0.1

src/formats/apple-notes/convert-scan.ts
```diff
@@ -24,11 +24,14 @@ export class ScanConverter extends ANConverter {
...
+				SELECT z_pk, zmedia, ztypeuti FROM ziccloudsyncingobject 
 				WHERE zidentifier = ${imageUuid}`;
 			
-			const file = await this.importer.resolveAttachment(row.Z_PK, ANAttachment.Scan);
-			if (file) links.push(this.importer.app.fileManager.generateMarkdownLink(file, '/'));
+			// Try to get the nicely cropped version, but fallback to the raw image if that fails
+			let file = await this.importer.resolveAttachment(row.Z_PK, ANAttachment.Scan);
+			if (!file) file = await this.importer.resolveAttachment(row.ZMEDIA, row.ZTYPEUTI);
+			
+			links.push(this.importer.app.fileManager.generateMarkdownLink(file!, '/'));
```
tgrosinger, line 34: Looks like `resolveAttachment` can still return null. This should probably still be guarded by a check if file is valid.

src/formats/apple-notes.ts
```diff
@@ -287,10 +287,12 @@ export class AppleNotesImporter extends FormatImporter {
 						AND z_pk = ${id} 
 				`;
 				
+				const filename = row.ZFALLBACKIMAGEGENERATION ? 'FallbackImage.png' : `${row.ZIDENTIFIER}.jpg`;
```
tgrosinger, line 293: I'm not sure when `ZFALLBACKIMAGEGENERATION` is valid, but I just wanted to double check this condition is not backwards?

mirnovov: It's not.

mirnovov:
Should be fixed now, I've set up a test vault on an older version of macOS and managed to reproduce the issue.

kepano:
Beautiful! Drawings are working for me now.

PR #140, opened 2023-10-19
Fix failure to load on mobile

### LINKED ISSUE #138: Unable to enable obsidian importer plugin on obsidian mobile.
I'm able to install the plugin, but when I try to enable it, this happens.

C23046F4-59F4-4D18-A23E-68B10F493146_4_5005_c

Device: Iphone 15 pro max
OS: IOS 17.03

mirnovov:
Closes #138 

`sqlite-tag-spawned` imports a bunch of `node` modules in the global scope. To avoid this, bundle `sqlite-tag-spawned` (similar to what is done with `yarle`) and only load said modules on desktop.


mirnovov:
Btw, `Platform.isMacOS` is true on iOS as well. Is this intentional?

lishid:
> Btw, `Platform.isMacOS` is true on iOS as well. Is this intentional?

Mostly; a lot of the behaviors between macOS and iOS are shared - things like the command modifier key (instead of ctrl), and plus iOS browsers "pretend" to be macOS. If you want to test for desktop mac then you'd want to check `Platform.isDesktopApp` as well.

src/main.ts
```diff
@@ -204,6 +204,7 @@ export default class ImporterPlugin extends Plugin {
 	authCallback: AuthCallback | undefined;
 
 	async onload() {
+		console.log('a');
```
lishid, line 207: This was probably not intentional?

lishid (approved):

src/main.ts
lishid, line 207:
```suggestion
```

PR #155, opened 2023-10-25
Apple Notes: Don't skip first line when there is no second line

### LINKED ISSUE #153: [Apple Notes] Long titles truncated + omit first line
Notes with a really long "title" get truncated with an ellipsis during the import. This combined with the option to "omit first line" means some content from the original note is lost. 

I noticed a few suspicious notes with long truncated titles but no body content. I likely pasted them into Apple Notes and they appear as long, single-line, word-wrapped, paragraphs. (not proud)

Awkward, and low priority, but Apple does permit this behavior. :)

mirnovov:
This is a very simple patch that closes #153.


tgrosinger (approved):
Thank you @mirnovov

Issue #150, opened 2023-10-21
Add option to only use single line breaks when converting


jgscherber:
I've only experienced this when converting from Notion.

There's an option in Obsidian to not use strict line breaks in reader view, it'd be nice if this option was also added to the importer. Without this option in the importer, there ends up being lots of extra new lines in Obsidian after a conversion that aren't needed with "Use strict line breaks" disabled 

![image](https://github.com/obsidianmd/obsidian-importer/assets/6207661/6345db6b-c0c2-40f7-898a-0451158f16d0)

joshuatazrein:
In my upcoming pull request I've added a "single line breaks" option in the Notion importer.

kepano:
Closed via [1.5.5](https://github.com/obsidianmd/obsidian-importer/releases/tag/1.5.5)

Issue #92, opened 2023-09-07
Notion callouts are not handled properly


grintroy:
When importing from Notion, the callouts are not translated into [Obsidian callouts](https://help.obsidian.md/Editing+and+formatting/Callouts). Instead, just the icon and the text are imported.

For example, consider this callout (screenshot from Notion):

![Notion callout](https://github.com/obsidianmd/obsidian-importer/assets/70978482/85f755b8-e2be-4f3d-9501-c853c3eb77cd)

When imported to Obsidian, it is translated to

```
💡
    
PCIe can use multiple (up to 32) lanes per card reaching the max rate of 31.51 GB/s for v4.0.
```

Instead, it should be

```
> [!info]
> PCIe can use multiple (up to 32) lanes per card reaching the max rate of 31.51 GB/s for v4.0.
```

It should be unindented and might use a different
type identifier.

kepano:
Closed via [1.5.5](https://github.com/obsidianmd/obsidian-importer/releases/tag/1.5.5)

PR #158, opened 2023-10-26
Notion bug squash bounty

joshuatazrein:
Fixed all issues in the bounty as well as these feature requests:
- [Single line breaks](https://github.com/obsidianmd/obsidian-importer/issues/150)
- [Callouts](https://github.com/obsidianmd/obsidian-importer/issues/92)

### DIFF
--- a/src/format-importer.ts
+++ b/src/format-importer.ts
@@ -148,7 +148,7 @@ export abstract class FormatImporter {
 	 */
 	async createFolders(path: string): Promise<TFolder> {
 		// can't create folders starting with a dot
-		const sanitizedPath = path.replace(/^\./g, '').replace(/\/\./g, '/');
+		const sanitizedPath = path.split('/').map(segment => segment.replace(/^\.+/, '')).join('/');
 		let normalizedPath = normalizePath(sanitizedPath);
 		let folder = this.vault.getAbstractFileByPathInsensitive(normalizedPath);
 		if (folder && folder instanceof TFolder) {
--- a/src/format-importer.ts
+++ b/src/format-importer.ts
@@ -147,7 +147,9 @@ export abstract class FormatImporter {
 	 * Recursively create folders, if they don't exist.
 	 */
 	async createFolders(path: string): Promise<TFolder> {
-		let normalizedPath = normalizePath(path);
+		// can't create folders starting with a dot
+		const sanitizedPath = path.replace(/^\./g, '').replace(/\/\./g, '/');
+		let normalizedPath = normalizePath(sanitizedPath);
 		let folder = this.vault.getAbstractFileByPathInsensitive(normalizedPath);
 		if (folder && folder instanceof TFolder) {
 			return folder;
--- a/src/formats/notion.ts
+++ b/src/formats/notion.ts
@@ -27,11 +27,13 @@ export class NotionImporter extends FormatImporter {
 				.onChange((value) => (this.parentsInSubfolders = value)));
 
 		new Setting(this.modal.contentEl)
-		.setName('Single line breaks')
-		.setDesc('Separate Notion blocks with only one line break (default is 2).')
-		.addToggle((toggle) => toggle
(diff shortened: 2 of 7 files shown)


src/main.ts
```diff
@@ -9,6 +9,7 @@ import { NotionImporter } from './formats/notion';
 import { OneNoteImporter } from './formats/onenote';
 import { RoamJSONImporter } from './formats/roam-json';
 import { truncateText } from './util';
+import { NodePickedFile } from 'filesystem';
```
lishid, line 12: Can we revert the changes to this file?

src/format-importer.ts
```diff
@@ -147,7 +147,9 @@ export abstract class FormatImporter {
 	 * Recursively create folders, if they don't exist.
 	 */
 	async createFolders(path: string): Promise<TFolder> {
-		let normalizedPath = normalizePath(path);
+		// can't create folders starting with a dot
+		const sanitizedPath = path.replace(/^\./g, '').replace(/\/\./g, '/');
```
lishid, line 151:
```suggestion
		const sanitizedPath = path.split('/').map(n => n.replace(/^\.+/, '')).join('/');
```
Suggest changing to a split/join and processing each part separately, and also adding `+` to the regex in case there were multiple dots.

src/formats/notion.ts
```diff
@@ -22,6 +25,13 @@ export class NotionImporter extends FormatImporter {
 			.addToggle((toggle) => toggle
 				.setValue(this.parentsInSubfolders)
 				.onChange((value) => (this.parentsInSubfolders = value)));
+
+		new Setting(this.modal.contentEl)
+		.setName('Single line breaks')
+		.setDesc('Separate Notion blocks with only one line break (default is 2).')
+		.addToggle((toggle) => toggle
+			.setValue(this.singleLineBreaks)
+			.onChange((value) => (this.singleLineBreaks = value)));
```
tgrosinger, line 34:
```suggestion
			.onChange((value) => {
				this.singleLineBreaks = value;
			}));
```

src/formats/notion/convert-to-md.ts
```diff
@@ -95,6 +108,10 @@ const typesMap: Record<NotionProperty['type'], NotionPropertyType[]> = {
 	],
 };
 
+function singleLineBreaks(markdownBody: string) {
```
tgrosinger, line 111: This probably doesn't need to be a separate function.

src/formats/notion/convert-to-md.ts
```diff
@@ -217,13 +234,40 @@ function fixEquations(body: HTMLElement) {
 	}
 }
 
+function stripToSentence(paragraph: string) {
+	const firstSentence = paragraph.match(/^[^\.\?\!\n]*[\.\?\!]?/)?.[0]
+	return firstSentence ?? ''
+}
+
+function isCallout(element: Element) {
```
tgrosinger, line 242: Can this function be simplified to something like

```ts
return !!(/callout|bookmark/.test(element.getAttribute('class'))
```

I'm not seeing what the nullish is adding since both nullish values and an empty string are going to going to skip that clause.

src/main.ts
joshuatazrein: The current option for debugging that's commented out is outdated—I'll revert these changes but I'd recommend updating the commented-out debugging option for developers

src/formats/notion/convert-to-md.ts
joshuatazrein: getAttribute('class') might return undefined, which breaks the regex.

joshuatazrein:
I've resolved comments, let me know if there's anything else to address before merging.

joshuatazrein:
Just checking if this is ready to merge!

tgrosinger (approved):

tgrosinger:
Looks great, thank you.

Issue #181, opened 2023-11-25
Import data from Bear (.bear2bk) skips half of my files?


infernokodiak:
Is there someway to figure out why or something I can do differently? 

Bear Version 2.0.17 (12140) latest
Obsedian Version 1.4.16 (Installer 1.4.16) latest

<img width="1508" alt="Screenshot 2023-11-25 at 3 51 27 PM" src="https://github.com/obsidianmd/obsidian-importer/assets/254689/1fb1058b-a6ea-489c-966b-57328e2cca50">

infernokodiak:
Any idea as to how to get more debugging information?

eware-godaddy:
I had the same thing. I think the issue is that if it detect that there is an asset (eg. image) that has the same name that already exists then it outputs that message. It actually does import the note, but the issue is that it will link the first version of the file. 

This is incorrect behaviour. As you may have *many* images with image1.png etc. So all your notes will link to the wrong image.

nicolashohm:
Looking into the logs, I found that most of the skipped files are the `info.json` file in the textbundle. In my PR #205 I ignored them instead of reporting them as skipped, this should limit the output to the real skipped files. In my case they were duplicated across notes (see #67)

PR #205, opened 2024-01-27
Improve Bear import

nicolashohm:
Recently, I tried to import my Bear notes. I faced a few issues which this PR addresses:
* File timestamps are set to the import time, not the actual creation or modification of the note
* Roughly half of the files are marked as skipped (#181) without any explanation
* Notes from archive and trash are imported like normal notes

### DIFF
--- a/src/formats/bear-bear2bk.ts
+++ b/src/formats/bear-bear2bk.ts
@@ -5,8 +5,8 @@ import { ImportContext } from '../main';
 import { readZip, ZipEntryFile } from '../zip';
 
 type Metadata = {
-	ctime: number,
-	mtime: number,
+	ctime?: number,
+	mtime?: number,
 	archived: boolean,
 	trashed: boolean,
 }
@@ -41,7 +41,7 @@ export class Bear2bkImporter extends FormatImporter {
 			if (ctx.isCancelled()) return;
 			ctx.status('Processing ' + file.name);
 			await readZip(file, async (zip, entries) => {
-				const metadata = await this.colllectMetadata(entries);
+				const metadataLookup = await this.collectMetadata(entries);
 				for (let entry of entries) {
 					if (ctx.isCancelled()) return;
 					let { fullpath, filepath, parent, name, extension } = entry;
@@ -60,9 +60,17 @@ export class Bear2bkImporter extends FormatImporter {
 								mdContent = mdContent.replace(assetMatcher, `![](${attachmentsFolder.path}/`);
 							}
 							const filePath = normalizePath(mdFilename);
-							const targetFolder = metadata[parent].archived ? archiveFolder : metadata[parent].trashed ? trashFolder : outputFolder;
+							const metadata = metadataLookup[parent];
+							let targetFolder = outputFolder;
+							if (metadata?.archived) {
+								targetFolder = archiveFolder;
+							} else if (metadata?.trashed) {
+								targetFolder = trashFolder;
+							}
 							const file = await this.saveAsMarkdownFile(targetFolder, filePath, mdContent);
-							await this.modifFileTimestamps(metadata[parent], file);
+							if (metadata?.ctime && metadata?.mtime) {
+								await this.modifFileTimestamps(metadata, file);
+							}
 							ctx.reportNoteSuccess(mdFilename);
 						}
 						else if (filepath.match(/\/assets\//g)) {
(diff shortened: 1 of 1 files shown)


src/formats/bear-bear2bk.ts
```diff
@@ -53,16 +57,17 @@ export class Bear2bkImporter extends FormatImporter {
 							mdContent = this.removeMarkdownHeader(mdFilename, mdContent);
 							if (mdContent.match(assetMatcher)) {
 								// Replace asset paths with new asset folder path.
-								mdContent = mdContent.replace(assetMatcher, `![](${attachmentsFolderPath.path}/`);
+								mdContent = mdContent.replace(assetMatcher, `![](${attachmentsFolder.path}/`);
 							}
 							const filePath = normalizePath(mdFilename);
-							const file = await this.saveAsMarkdownFile(outputFolder, filePath, mdContent);
+							const targetFolder = metadata[parent].archived ? archiveFolder : metadata[parent].trashed ? trashFolder : outputFolder;
```
tgrosinger, line 63: I think this would be easier to read if the nested ternary operators were avoided:

```typescript
    let targetFolder = outputFolder;
    const parentMetadata = metadata[parent];
    if (parentMetadata?.archived) {
        targetFolder = archiveFolder;
    } else if (parentMetadata?.trashed) {
        targetFolder = trashFolder;
    }
```

tgrosinger (changes requested):
Thank you for these improvements. Just a couple of changes please.

src/formats/bear-bear2bk.ts
```diff
@@ -50,7 +56,13 @@ export class Bear2bkImporter extends FormatImporter {
...
 							}
 							let filePath = normalizePath(mdFilename);
-							await this.saveAsMarkdownFile(outputFolder, filePath, mdContent);
+							const file = await this.saveAsMarkdownFile(outputFolder, filePath, mdContent);
+
+							const writeOptions: DataWriteOptions = {
+								ctime: metaData[parent].ctime,
+								mtime: metaData[parent].mtime,
+							};
+							await this.vault.append(file, '', writeOptions);
```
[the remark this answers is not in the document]
lishid: Yeah I don't think there's a better way for now given the APIs exposed.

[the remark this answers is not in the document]
tgrosinger: Sounds good, thanks.

src/formats/bear-bear2bk.ts
```diff
@@ -78,6 +90,22 @@ export class Bear2bkImporter extends FormatImporter {
 		}
 	}
 
+	private async colllectMetadata(entries: ZipEntryFile[]): Promise<{ [key: string]: Metadata; }> {
+		let metaData: { [key: string]: Metadata; } = {};
+		const infoEntries = entries.filter((entry) => entry.name === 'info.json');
```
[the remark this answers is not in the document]
nicolashohm: I tried that first, but unfortunately the iteration first hits the markdown file and then the json file. That's why I made a dedicated iteration before

src/formats/bear-bear2bk.ts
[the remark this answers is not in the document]
nicolashohm: I think I copied that approach from the `onenote.ts` format. 

I just checked the `notion.ts` format, there it is a single call `vault.create(path, markdownBody, writeOptions)` instead of `this.saveAsMarkdownFile(outputFolder, filePath, mdContent)`. I don't know what's the difference? I can give it a try!

nicolashohm:
@tgrosinger Thanks for your feedback. I pushed another commit c67292c addressing your comments.

src/formats/bear-bear2bk.ts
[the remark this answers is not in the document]
nicolashohm: Sorry, mixed it up. This loop can be merged with the next one. Done in c67292c

src/formats/bear-bear2bk.ts
```diff
@@ -100,15 +108,19 @@ export class Bear2bkImporter extends FormatImporter {
 
 	private async colllectMetadata(entries: ZipEntryFile[]): Promise<{ [key: string]: Metadata; }> {
 		let metaData: { [key: string]: Metadata; } = {};
-		const infoEntries = entries.filter((entry) => entry.name === 'info.json');
-		for (let entry of infoEntries) {
+		for (let entry of entries) {
+			if (entry.name === 'info.json') {
```
tgrosinger, line 112: Is this backwards? I think we should be continuing if the file is _not_ `info.json`.

tgrosinger (changes requested):
Thank you, looks great except for one question.

src/formats/bear-bear2bk.ts
arthurtyukayev: Continue in this context will just skip this file, it sort of means "Continue with the loop, skipping this file"

src/formats/bear-bear2bk.ts
```diff
@@ -75,6 +98,36 @@ export class Bear2bkImporter extends FormatImporter {
...
 
+	private async modifFileTimestamps(metaData: Metadata, file: TFile) {
+		const writeOptions: DataWriteOptions = {
+			ctime: metaData.ctime,
+			mtime: metaData.mtime,
+		};
+		await this.vault.append(file, '', writeOptions);
+	}
+
+	private async colllectMetadata(entries: ZipEntryFile[]): Promise<{ [key: string]: Metadata; }> {
```
tgrosinger, line 109: typo, should be `collectMetadata`

src/formats/bear-bear2bk.ts
tgrosinger: Right, but in this loop we're specifically looking for `info.json` files and should be skipping other files.

Actually looking at this again, there's only ever going to be one `info.json` in `entries` so we could instead use `let infoJsonEntry = entries.find((entry) => entry.name === 'info.json');`

src/formats/bear-bear2bk.ts
parkerbxyz:
```suggestion
	private async collectMetadata(entries: ZipEntryFile[]): Promise<{ [key: string]: Metadata; }> {
```

src/formats/bear-bear2bk.ts
nicolashohm: @tgrosinger Thanks for catching that. I tested my changes but without running npm build so it was the old version I tested with 🤦‍♂️  I fixed it now in 84bd00f.

Regarding `find()`, the `entries` array contains a flat list of all files in the backup. There are a bunch of `info.json` and related `text.md` files. The parent contains the name of the textbundle they belong to. Here an extract of the  `entries` array:
Bildschirmfoto 2024-02-12 um 21 36 18

tgrosinger: Ah, this is iterating through all the textbundles at once. Got it.

tgrosinger (approved):

tgrosinger:
Thank you @nicolashohm!

Issue #204, opened 2024-01-26
Evernote bug bounty - $300


kepano:
## Bounty

See the [Contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md) for how to claim this bounty.

**Bounty:** $300 USD
**Timeframe:** 10 days

## Requirements

Bug bounty including the following Evernote bugs. Most of them are related to improper formatting, or missing conditions in file names.

- [ ] #24
- [ ] #187
- [ ] #26
- [x] #184
- [ ] #185
- [ ] #190
- [ ] #194
- [ ] #200
- [ ] #201

kepano:
@akosbalasko let us know if you're interested in picking this up!

akosbalasko:
Hi @kepano ,
Thanks for tagging me!
Sure, I'm definitely interested, let me check the issues and get back to you tomorrow! (It's almost midnight here)

akosbalasko:
Hi @kepano ,

According to the contribution description, here is my _formal_ application to the bounty: 
**Why would I be the best fit to the job:** 
Because I developed and actively maintain Yarle, I know what is where and how it is structured logically.
**How I would approach the requirements:**
This batch contains different type of issues: features(#200, #26) or bugs( #184, #185, #190), and  detailed in various level. 
- Some of them can be achieved as they are. 
- Some others (like #24  or #187 ) phisically cannot (Evernote does not export the required information into the enex). Therefore these latter issues would require workarounds involving manual pre-steps from the users at Evernote-side.
- Some others are more general (#194 #201), I'd need more info about the problem. For instance I tested #201 , but it importing DFX works for me totally, so I think the problem is not related to the extension of the attachment. 

Thanks again!

kepano:
I have assigned the bounty to you. Please submit your initial PR by end of day February 6th.

Let's start with the easiest and most obvious ones first. If you have any questions about requirements on specific issues please post them here.

It sounds like we should consider ignoring #24 and #187, unless you can think of a way to support those?

akosbalasko:
Hi @kepano ,
I have some questions about some of the issues: 
#184 : I cannot reproduce the problem, $ sign within note title works (it contain an image as well), linked. 
#201 : same, it looks like that the issue description (e.g. the DFX extention causes problems) doesn't point the real problem. 

What is your expectation, may I contact with the reporters and ask for more details? 

[...]

Issue #190, opened 2023-12-14
Evernote: File name of a resource is cut off at the first dot


muhme:
I have a lot of PDFs with file names like `ct.21.23.072-074.pdf` in my Evernote notes. They are all imported as `ct.pdf` or, if there are multiple PDFs in one note as `ct.1.pdf` for the second and `ct.2.pdf` for the third and so on. The file name is truncated at the first dot in the file name. For me, this means valuable information is lost (article is from issue 23 in year 2021 and has the three pages 72 ... 74).

I would like the file name `ct.21.23.072-074.pdf` to remain the same in Obsidian after the import. If this is not possible, how about replacing the dots with underscores. For this example, the file name `ct_21_23_072-074.pdf` would be valuable.

ENEX lines are:
```xml
      <mime>application/pdf</mime>
      <resource-attributes>
        <file-name>ct.21.23.072-074.pdf</file-name>
```

Found:
- on macOS Sonoma 14.1.2
- export with Evernote 10.67.4-mac-mas-public (20231204112355), Editor: V 176.32.0, Service: V 1.80.3
- import with Obsidian Version 1.4.16

muhme:
As a workaround, I wrote the small script [sfnfoi.sh](https://github.com/muhme/scripts/blob/main/sfnfoi.sh), which replaces all dots in file name in the Evernote export for the Obsidian import with underlines, with the exception of the last .pdf. This solves my import problem 😄

Issue #187, opened 2023-12-08
[Evernote] Tag hierarchy/structure is not imported


grm9762:
Hey there! 

I have created an .enex file with Evernote Legacy v. 7.14.1 on my Mac.

I work exclusively with tags in Evernote and have built up a very extensive tag structure.

Unfortunately, this structure (nested tags) is lost when importing. Only the tags that are actually assigned to notes appear in Obsidian. This means that the tag headings, which are only for the overview and are not assigned to any notes, are not present in Obsidian. 
 
Is there a workaround for this?

Thanks for your support!

kepano:
Supported in Importer 1.6.0 but requires some work in Evernote to prepare for import, see documentation
https://help.obsidian.md/import/evernote

Issue #184, opened 2023-11-29
Evernote: Notes that include $ symbol have broken links


OM1720:
If evernote-notes title include "$" all images in obsidian-note have broken links to its.

akosbalasko:
Hi @OM1720 ,
unfortunately I cannot reproduce the problem, could you please give us more details about the notes in where the problem occurred? 
If I understand you well, you have a dollar sign in the title of the note in Evernote. This note contains images. And after the conversion the images are not  linked correctly to the converted note, is this the case? 
Do you see some other special characters in the file/folder or even in the name of the image?
Thanks a lot!

OM1720:
HI!

Sorry, I cannot reproduce the problem too. Please, close the ticket.

With best regards, Mike

PR #207, opened 2024-02-02
#204 - Evernote bug bounty

### LINKED ISSUE #24: Importing .enex do not save folders
Is there a way to respect my evernote folder structure when importing the .enex file?

### LINKED ISSUE #185: Evernote: Titles with quotation marks are imported as Untitled Note
If evernote-notes title surrounded by quotation marks in obsidian-note doesn't have title (Untitled Note).

### LINKED ISSUE #194: [Evernote] Fails to set some notes' dates
I have just used 1.5.6 to import ENEX notes. Some of the 394 notes were not set dates according to the ones stored in the ENEX file. I end up with writing Python code to fix this. Interestingly os.utime() failed to set those dates also and I was forced to use win32file.SetFileTime().

### LINKED ISSUE #200: Evernote: imported notes do not include source links embedded in Evernote original
When material is pasted into Evernote from the web, a URL for the source page is also saved. This URL doesn't appear as a tag or within body copy, but as a separate field in the Evernote interface.

This URL isn't preserved in Obsidian pages produced by the importer.

When the same ENEX file is imported to Joplin, these URLs do appear in Note Properties.

Could the Obsidian importer be modified to include the source URL, perhaps as the first line of body copy? It's a rather important piece of data.

### LINKED ISSUE #201: [Evernote] Import loses files with the .dxf extension
Importing files from Evernote, for example with the .dxf extension, 
export works fine if the exported note is imported back into Evernote with all files in place.

akosbalasko:
Evernote improvements and fixes: https://github.com/obsidianmd/obsidian-importer/issues/204

#190  - Fixed
#200 - Implemented
#26 - Fixed
#201  - Fixed
#24 - Implemented, "@@@" used as delimiter between notebookstack's and notebook's name in the enex file
#187 - Cannot be implemented phisically, possible solution would be to ask the users to flat nested tags by '/' before the conversion within Evernote
#194  - Cannot reproduce, possible reason patched
#184 - Cannot reproduce, reporter asked to close
#185 - Cannot  reproduce, no hints from reporter

+1 -  Title removed from the converted note's body


kepano:
@akosbalasko thanks! Will review shortly. Two more things:

1. Can you submit a PR to the [obsidian-help](https://github.com/obsidianmd/obsidian-help) repo with additional detail on the [Evernote import instructions](https://github.com/obsidianmd/obsidian-help/blob/master/en/Import%20notes/Import%20from%20Evernote.md) for handling tags/folders?
2. Can you reply here stating `I have read the CLA agreement and I hereby sign the CLA.` per our [contribution guidelines](https://github.com/obsidianmd/obsidian-importer/blob/master/CONTRIBUTING.md)?

src/formats/yarle/utils/filename-utils.ts
```diff
@@ -37,7 +37,7 @@ export const getResourceFileProperties = (workDir: string, resource: any): Resou
 
 	if (resource['resource-attributes'] && resource['resource-attributes']['file-name']) {
 		const fileNamePrefix = resource['resource-attributes']['file-name'].substr(0, 50);
-		fileName = fileNamePrefix.split('.')[0];
+		fileName = fileNamePrefix.includes('.') ? fileNamePrefix.split('.').slice(0, -1).join('.') : fileNamePrefix;
```
tgrosinger, line 40: I think we could make this a little more efficient and avoid searching through the string multiple times:

```
const lastIdx = fileNamePrefix.lastIndexOf('.');
fileName = lastIdx !== -1
    ? fileNamePrefix.slice(0, lastIdx)
    : fileNamePrefix;
```

src/formats/yarle/utils/templates/default-template.ts
```diff
@@ -1,5 +1,7 @@
-const tagBlock = '{tags-array-block}---\ntags: {tags-array}\n---\n{end-tags-array-block}';
+const frontmatterDelimiter = '---\n';
+const sourceBlock = '{source-url-block}\nsource: {source-url}\n\n{end-source-url-block}';
```
tgrosinger: Will this cause the `source` property to be present even for files that do not have a source specified?

src/formats/yarle/utils/templates/checker-functions.ts
```diff
@@ -40,7 +41,8 @@ export const hasSourceURLInTemplate = (templateContent: string): boolean => {
 };
 export const hasAnyTagsInTemplate = (templateContent: string): boolean => {
 	return (hasItemInTemplate(TAGS, templateContent)
-		|| hasItemInTemplate(YAMLARRAYTAGS, templateContent));
```
tgrosinger, line 43: Should we remove the various bits of code for array tags now? It looks like they are unused?

src/formats/yarle/utils/templates/default-template.ts
```diff
@@ -1,7 +1,6 @@
 const frontmatterDelimiter = '---\n';
 const sourceBlock = '{source-url-block}source: {source-url}{end-source-url-block}\n';
 const tagBlock = '{tags-yaml-list-block}\ntags: {tags-yaml-list}\n\n{end-tags-yaml-list-block}';
-const titleBlock = '{title-block}# {title}\n\n{end-title-block}';
```
tgrosinger, line 2: Is there an issue corresponding to this change? I'm not sure if this is change we wanted to make.

src/formats/yarle/process-resources.ts
```diff
@@ -77,9 +77,13 @@ const processResource = (workDir: string, resource: any): any => {
 	// Skip unknown type as we don't know how to handle
 	// Source: https://dev.evernote.com/doc/articles/data_structure.php
 	// "The default type "application/octet-stream" should be used if a more specific type is not known."
-	if (resource.mime === 'application/octet-stream') {
+	// Update:
```
tgrosinger, line 80: @lishid or @ericaxu, does this seem reasonable to you? I'm not too familiar with Yarle.

src/formats/yarle/utils/folder-utils.ts
```diff
@@ -119,7 +120,20 @@ export const clearResourceDir = (note: any): void => {
...
+export const handleNotebookstacks = (enex: PickedFile, options: YarleOptions): void => {
+	const notebookFolderNames = enex.basename.split('@@@');
+	let notebookName = notebookFolderNames.pop();
+	if (!notebookName){
+		notebookName = enex.basename;
+	}
+	fs.mkdirSync(path.join(options.outputDir, ...notebookFolderNames), { recursive: true });
+	enex.fullpath = enex.fullpath.replace(enex.basename, notebookName || enex.basename);
+	enex.name = enex.name.replace(enex.basename, notebookName || enex.basename);
+	options.outputDir = [options.outputDir, ...notebookFolderNames].join(options.pathSeparator);
```
tgrosinger, line 132: Will this behave as expected if there are multiple files being imported?

It looks like if we modify the outputDir here it will affect subsequent files being imported.

tgrosinger (changes requested):

src/formats/yarle/utils/folder-utils.ts
akosbalasko: Nope, because I save the original outputDir  at line 187 in yarle.ts and restore after the conversion (line 201).

src/formats/yarle/utils/templates/checker-functions.ts
akosbalasko: well, yeah, I was thinking on it, as you wish, I think the whole array tags in yaml together with its templates is currently not used. @lishid any suggestion?

src/formats/yarle/utils/templates/default-template.ts
akosbalasko: no dedicated issue for this, I discussed it with @kepano here: https://github.com/obsidianmd/obsidian-importer/issues/204#issuecomment-1915558518

src/formats/yarle/utils/filename-utils.ts
```diff
@@ -37,7 +37,10 @@ export const getResourceFileProperties = (workDir: string, resource: any): Resou
 
 	if (resource['resource-attributes'] && resource['resource-attributes']['file-name']) {
 		const fileNamePrefix = resource['resource-attributes']['file-name'].substr(0, 50);
-		fileName = fileNamePrefix.split('.')[0];
+		const lastIdx = fileNamePrefix.lastIndexOf('.');
```
akosbalasko, line 40: thx @tgrosinger , I applied your solution.

src/formats/yarle/utils/templates/default-template.ts
akosbalasko: Nope, it is removed from the template automatically if no source property specified in the note data structure

src/formats/yarle/utils/templates/default-template.ts
tgrosinger: Ah great, thank you.

src/formats/yarle/utils/folder-utils.ts
tgrosinger: I believe that's just saving a reference to the object, but that's the same object that will be modified here.

tgrosinger: Ah, sorry I misread and thought you were storing the full options object.
Thank you for clarifying for me!

tgrosinger (approved):
Looks good to me, pending the remaining comments for @lishid.

akosbalasko:
Any updates on this @kepano @lishid @ericaxu ?

src/formats/yarle/utils/templates/checker-functions.ts
lishid: I think removing it would be best, thanks!

lishid:
Sorry I am late!

src/formats/yarle/utils/folder-utils.ts
```diff
@@ -119,7 +120,20 @@ export const clearResourceDir = (note: any): void => {
...
 
-export const setPaths = (enexSource: PickedFile): void => {
+export const handleNotebookstacks = (enex: PickedFile, options: YarleOptions): void => {
+	const notebookFolderNames = enex.basename.split('@@@');
+	let notebookName = notebookFolderNames.pop();
+	if (!notebookName){
+		notebookName = enex.basename;
+	}
+	fs.mkdirSync(path.join(options.outputDir, ...notebookFolderNames), { recursive: true });
+	enex.fullpath = enex.fullpath.replace(enex.basename, notebookName || enex.basename);
```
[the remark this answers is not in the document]
akosbalasko: > These aren't meant to be modified. PickedFile represents a file picked from the operating system. If we need to work on a different set of names over the original file then we should use a proper data structure to make this clear.

@lishid But in this case, it would mean just to clone the current PickedFile, changing its props and use it instead of that we got from outside, or not? 

> Also, this wouldn't work if the full path contains a piece that is identical to the enex basename. For example, C:/Evernote/Files/Evernote.enex would only get the first "Evernote" replaced instead of the intended second.

Good point I'll resolve it.

src/formats/yarle/utils/folder-utils.ts
akosbalasko: @lishid ok, let me try to improve it.

src/formats/yarle/utils/folder-utils.ts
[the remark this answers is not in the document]
lishid: > @lishid But in this case, it would mean just to clone the current PickedFile, changing its props and use it instead of that we got from outside, or not?

Can you explain the use case here? If I'm understanding it correctly you'd like to specify a different title/output path for the same input file, right? In that case I would suggest passing this additional information over to the code that uses it (through an object or function parameters) to clearly indicate that it's the supposed output path.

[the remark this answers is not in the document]
akosbalasko: @lishid yeah, that would be the use case, and I used to use options object would to store these kind of info. Now I cleaned a code a bit around notebook stacks, hope it looks better for you too.

src/formats/yarle/utils/templates/checker-functions.ts
akosbalasko: @lishid removed

src/formats/yarle/utils/folder-utils.ts
```diff
@@ -118,8 +120,42 @@ export const clearResourceDir = (note: any): void => {
...
+};
+
+
+export const getNotebookStackedEnex = (baseEnex: PickedFile ): NodePickedFile => {
+	if (!(baseEnex instanceof NodePickedFile)) throw new Error('Evernote import currently only works on desktop');
+
+	const { notebookName } =  getNotebookNameAndFolderNames(baseEnex.basename);
+
+	const stackedPickedFile = new NodePickedFile((baseEnex as NodePickedFile).filepath);
+	stackedPickedFile.fullpath = replaceLastOccurrenceInString(baseEnex.fullpath, baseEnex.basename, notebookName || baseEnex.basename);
```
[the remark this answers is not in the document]
akosbalasko: @lishid Ohh, okay, now I get it, thanks for the clarification! Changes made to make it work with untouchable NodePickedFiles. :)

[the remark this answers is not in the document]
lishid: As I understand it, what's happening here is that we are finding some code down the line that uses these fields, and we'd like to change their behavior to use a different names/paths. Doing it this way is a "shortcut"; while it results in minimal amount of changes, it also makes the code much more confusing. Similar to what I said in the other comment, when you have a field that is defined to mean one thing, and you use it for a slightly different purpose (and thus modify the contents to fit your alternative needs), it creates a chance of confusion, or at the very least, forces the reader to "keep in mind" these extra concerns. This is often where bugs can slip in, especially if there are many different little things one needed to keep in mind when working with the code.

Usually the best way to do this is being more explicit with the code to say what you intend with the variables and fields. Passing additional vars with names that indicate what they do (or is intended for) helps a lot.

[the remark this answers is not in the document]
lishid: Yup the new changes works for me.

lishid (approved):
Code looks ok to me. Might be worth doing one final pass of the auto formatter.

src/formats/yarle/utils/folder-utils.ts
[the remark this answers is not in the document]
akosbalasko: @lishid , sorry, I had a bad feeling about this, so I tried to improve the code further, introduced a couple of new functions to decrease the level of the possible confusion. Hope it looks better for you too.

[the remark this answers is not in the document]
lishid: Thanks. I would say don't worry too much about it. Yarle is unfortunately written in a very different paradigm of programming than the rest of the importer. It uses global singleton objects and naked functions with context being passed sometimes, set in the global object at other times.

I decided not to invest too much effort to rewrite yarle as it would not be worth the time for a functionally equivalent package.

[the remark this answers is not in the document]
akosbalasko: okay then we can move forward with it, right?

tgrosinger:
Looks good to me.

Issue #198, opened 2024-01-08
[OneNote] Attachment imports fail for duplicate file names


SSzretter:
If the same filename exists with note(s) the attachments are failing 'because Error: File already exists'.  
It should probably rename the attachment (1, 2, 3, etc.) in a loop so those attachments are not lost during import.

krizman:
Probably a duplicate of #126. I also have the same issue where I can't import a single notebook from OneNote. Copy-pasting content is also a futile task since for example the checkboxes get copied over as inline base64 encoded images ☑️ 😅

lniass:
I stopped migrating to Obsidian because I had too many files considered as "existing" that I did not want to lose.

Hoping there is soon a solution.

tgrosinger:
This should be resolved in the latest release - [1.6.1](https://github.com/obsidianmd/obsidian-importer/releases/tag/1.6.1)

Issue #177, opened 2023-11-14
OneNote bug bounty - $300


ericaxu:
This bug bounty including the following bugs:

- [x] #109
- [x] #120 
- [x] #126 
- [x] #162 
- [x] #165 
- [x] #166 
- [x] #198

p3rid0t:
I'd like to take on this bounty as I am the developer of the OneNote importer. Thanks!

adamhill:
@p3rid0t If you need a tester with large OneNote notebooks from he**. Please reach out, I have plenty with no personal data, just ideas with lots a various kinds of data and embeds.

kepano:
@p3rid0t checking in — do you have an ETA for your PR?

p3rid0t:
Sorry for the long wait time, was quite sick for the previous week, hoping to get it done by the end of this week (hopefully even quicker)

ericaxu:
Update: the bounty has been unassigned due to inactivity. If you're interested in this bounty please comment to let me know!

p3rid0t:
I would like to apply for this bounty again if it is still available. Sorry for the previous inactivity.

kepano:
@p3rid0t Great! Please move forward and submit your PR by February 2. Thank you!

kepano:
@p3rid0t I am adding #198 to this, if that's okay

kepano:
@p3rid0t are you on track to submit your PR today?

p3rid0t:
yes, hoping to do so today! if i'll be unable to submit today, i'll submit it tomorrow ASAP

p3rid0t:
Quick update: bounty tasks are done, just waiting with the PR until I fix issue #164 (too many line breaks), as I found it pretty annoying as well 😅

kepano:
@p3rid0t can you open up a PR for the fixes you have so far? That way we can start reviewing those changes.

Issue #166, opened 2023-11-03
[OneNote] Import of note will be capped at the point where a YouTube embedding is  


fatplanediaries:
Tried to import a note from OneNote that contained many YouTube links, as well as embedded players for those links. The importer will only import the note up to the point where the first embedding is, then fail to import the rest of the note. 

Below is an example of one of these OneNote embeddings. 

![image](https://github.com/obsidianmd/obsidian-importer/assets/127848260/d4490b52-e841-453c-9b34-f22677e6f147)

tgrosinger:
This should be resolved in the latest release - [1.6.1](https://github.com/obsidianmd/obsidian-importer/releases/tag/1.6.1)

Issue #165, opened 2023-11-02
[OneNote] import while maintaining the hierarchy of notes


vladik-pwnz:
does not save the hierarchy of notes, but simply saves in one row. example from OneNote
![Untitled1](https://github.com/obsidianmd/obsidian-importer/assets/61415206/1a98543c-79ac-4976-899e-f1faa34fe861)

legolasdimir:
This was restated in an issue I created #145 referencing it here for context

sol3uk:
I'm also getting this issue, it's preventing me from making the switch as I used Onedrive quite heavily.
This could be an issue with the fact that Onenote treats these "parent" sections as pages by themselves with actual content and sub pages.
I notice that Obsidian doesn't do this at all and instead opts for treating these as folders. I think the importing needs to be rethought to accommodate this so we get the option to keep our hierarchy whilst also keeping the contents of the parent pages.
Maybe the best way would be to create a folder when a page contains sub pages, then create a markdown file in the root of that folder with the same name (or a name that makes it obvious that it was the parent page's content)
Here's my before and after importing for an example:
OneDrive:
![image](https://github.com/obsidianmd/obsidian-importer/assets/26093813/cbd37d6f-5862-40c8-97b9-32cdea0cd114)
Obsidian Imported:
![image](https://github.com/obsidianmd/obsidian-importer/assets/26093813/327753fa-b4f5-41cd-bdb1-833e54322ee0)
_Lost all structure and hierarchy, still maintained the parent page content, but it's no longer obvious what is related and where pages should be. N.B. The grey box is just redacted content for this post._

glinkot:
Yep, I just found the same.  I thought hierarchy would be preserved (perhaps too optimistically) but I ended up with an alphabetical list of the notes regardless of where they are in the onenote hierarchy.  Would be amazing to preserve it.

tgrosinger:
This should be resolved in the latest release - [1.6.1](https://github.com/obsidianmd/obsidian-importer/releases/tag/1.6.1)

PR #209, opened 2024-02-07
OneNote Bug Bounty - #177

p3rid0t:
This PR fixes issues mentioned in the bounty's issue - #177, aka:
> * [x]  [Importer gets OneNote section names but does not import the notes #109](https://github.com/obsidianmd/obsidian-importer/issues/109)
> * [x]  [[OneNote] Importer does not back off, leading to rate limiting #120](https://github.com/obsidianmd/obsidian-importer/issues/120)
> * [x]  [[OneNote] Importer not working properly - "File already exists" #126](https://github.com/obsidianmd/obsidian-importer/issues/126)
> * [x]  [[OneNote] Nested Section Groups Under Nested Sections Groups Not Discovered #162](https://github.com/obsidianmd/obsidian-importer/issues/162)
> * [x]  [[OneNote] import while maintaining the hierarchy of notes #165](https://github.com/obsidianmd/obsidian-importer/issues/165)
> * [x]  [[OneNote] Import of note will be capped at the point where a YouTube embedding is   #166](https://github.com/obsidianmd/obsidian-importer/issues/166)
> * [x]  [[OneNote] Attachment imports fail for duplicate file names #198](https://github.com/obsidianmd/obsidian-importer/issues/198)

As well as some general code improvements, and issue #164 related to line breaks and whitespace.

Looking forward to what needs to be improved or fixed. Thanks!


src/formats/onenote.ts
```diff
@@ -142,150 +147,174 @@ export class OneNoteImporter extends FormatImporter {
...
+						const index = this.selectedIds.findIndex((sec) => sec === section.id);
+						if (index !== -1) {
+							this.selectedIds.splice(index, 1);
+						}
 					}
-				}
-			});
+				});
+
+				this.renderHierarchy(section, parentEl);
```
tgrosinger, line 215: Is this needed? A `OnenoteSection` does not have `sections` or `sectionGroups`.

src/formats/onenote.ts
```diff
@@ -142,150 +147,174 @@ export class OneNoteImporter extends FormatImporter {
...
 	}
 
 	async import(progress: ImportContext): Promise<void> {
-		// Remove possible duplicates, eg. when the user selects "Select all" with existing selections.
-		this.selectedSections = this.selectedSections.filter((item, index, array) => array.indexOf(item) === index);
-
-		let outputFolder = await this.getOutputFolder();
-		let remainingSections = this.selectedSections.length;
+		// Remove possible duplicate IDs, eg. when the user selects "Select all" with existing selections
+		this.selectedIds =  [...new Set(this.selectedIds)];
```
tgrosinger, line 222: Can we prevent duplicates from happening? Otherwise if there's a duplicate the checkbox won't do anything since it will only add/remove one of the entries for that ID.

src/formats/onenote.ts
```diff
@@ -142,150 +147,174 @@ export class OneNoteImporter extends FormatImporter {
...
+		
+		let remainingSections = this.selectedIds.length;
 
-		if (!outputFolder) {
+		if (!await this.getOutputFolder()) {
 			new Notice('Please select a location to export to.');
 			return;
 		}
+		else this.outputFolder = (await this.getOutputFolder())!;
```
tgrosinger, line 230: Why are we calling `getOutputFolder` twice?

src/formats/onenote.ts
```diff
@@ -142,150 +147,174 @@ export class OneNoteImporter extends FormatImporter {
...
 				}
 			}
 		}
 	}
 
-	async processFile(progress: ImportContext, sectionFolder: TFolder, outputFolder: TFolder, content: string, page: OnenotePage) {
+	async processFile(progress: ImportContext, content: string, page: OnenotePage) {
 		try {
+			// Removes multiple paragraphs in a row to replace them with a basic <br>
+			const paragraphRegex: RegExp = /<\/p>\s*<p[^>]*>/g;
```
tgrosinger, line 282: These constants can be declared outside of this function. At the top of the file would probably be a good place for them.

src/formats/onenote.ts
```diff
@@ -357,17 +385,108 @@ export class OneNoteImporter extends FormatImporter {
...
+			for (const notebook of this.notebooks) {
+				const path = this.getEntityPath(entityID, `${currentPath}/${notebook.displayName}`, notebook);
+				if (path) {
+					returnPath = path;
+					break;
+				}
+			}
+		}
+		else {
+			if (parentEntity.id === this.currentSection) {
```
tgrosinger, line 405: This doesn't seem like the right place for this to be happening. Retrieving the entity path should not have a side-effect of updating the object state.

src/formats/onenote.ts
```diff
@@ -142,150 +147,174 @@ export class OneNoteImporter extends FormatImporter {
...
+			  $select: 'id,title,createdDateTime,lastModifiedDateTime,level,order,contentUrl',
+			  $orderby: 'order',
+			  pagelevel: 'true'
+			});
 
-			const pagesUrl = `https://graph.microsoft.com/v1.0/me/onenote/sections/${section.id}/pages?$select=id,title,createdDateTime,lastModifiedDateTime`;
-			let pages: OnenotePage[] = (await this.fetchResource(pagesUrl, 'json')).value;
+			const pagesUrl = `${baseUrl}?${params.toString()}`;
+	
+			let pages: OnenotePage[] = ((await this.fetchResource(pagesUrl, 'json')).value);
```
tgrosinger, line 249: Can we put these pages on the section when we retrieve them rather than doing it in `getEntityPath`?

src/formats/onenote.ts
```diff
@@ -357,17 +385,108 @@ export class OneNoteImporter extends FormatImporter {
 		return pageElement;
 	}
 
+	/**
+	 * Returns a filesystem path for any OneNote entity like sections or notes
+	 * Paths are returned in the following format:
+	 * (Export folder)/Notebook/(possible section groups)/Section/(possible pages with a higher level) 
+	 */
+	getEntityPath(entityID: string, currentPath: string, parentEntity?: OnenoteEntityHierarchyModel | undefined): string | null {
```
tgrosinger, line 393: I think this function is trying to do too much and has too much casting. Maybe it would be simpler and safer if there were different helper functions for getting a path depending on the parent entity type.

src/formats/onenote.ts
```diff
@@ -439,27 +558,26 @@ export class OneNoteImporter extends FormatImporter {
...
+				const attachment = attachmentQueue[i];
 				try {
 					// Every 7 attachments, do a few second break to prevent rate limiting
 					if (i !== 0 && i % 7 === 0) {
 						await new Promise(resolve => setTimeout(resolve, 7500));
 					}
-					const data = (await this.fetchResource(attachment.contentLocation!, 'file')) as ArrayBuffer;
-					await this.app.vault.createBinary(attachmentPath + '/' + attachment.name, data);
+
+					if (!(await this.vault.adapter.exists(`${attachmentPath}/${attachment.name}`))) {
```
tgrosinger, line 569: Should we report the attachment as skipped if it already exists?

src/formats/onenote.ts
```diff
@@ -492,21 +610,14 @@ export class OneNoteImporter extends FormatImporter {
...
 					// Append the content and add fences in case there's no next element
 					codeElement.innerHTML = codeElement.innerHTML.slice(0, -3) + element.innerHTML + '\n```';
 				}
 			}
 			else if (element.nodeName === 'BR' && inCodeBlock) {
 				codeElement.innerHTML = codeElement.innerHTML.slice(0, -3) + '\n```';
-				element.remove();
 			}
 			else {
-				if (inCodeBlock) {
```
tgrosinger, line 505: Why are these lines being removed?

tgrosinger (changes requested):
Thanks for the PR @p3rid0t. Let me know if you have questions about any of these comments.

src/formats/onenote.ts
p3rid0t: Hmm, but if those are only used in this one function, are you sure they should be at the top of the file?

tgrosinger: I mainly meant it would be good to pull them out of this function since they are being re-instantiated every time the function is called.

p3rid0t: Thanks for noticing, that's a neat optimization.

src/formats/onenote.ts
```diff
@@ -213,10 +213,10 @@ export class OneNoteImporter extends FormatImporter {
 
 			let sectionFolder: TFolder;
 			if (section.parentSectionGroup) {
-				let sectionGroupFolder: TFolder = await this.createFolders(outputFolder.path + '/' + section.parentSectionGroup.displayName);
-				sectionFolder = await this.createFolders(sectionGroupFolder.path + '/' + section.displayName);
+				let sectionGroupFolder: TFolder = await this.createFolders(outputFolder.path + '/' + sanitizeFileName(section.parentSectionGroup.displayName!));
```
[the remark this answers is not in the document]
p3rid0t: FYI, this comment was (possibly accidentally) left on an older version of the file.

The current version doesn't use `sanitizeFileName` anymore as it's automatically being done anyway in the `createFolders` function

src/formats/onenote.ts
p3rid0t: That part was responsible for fixing line breaks in code blocks, but now line breaks are fixed in the entire file, so it's no longer necessary.

src/formats/onenote.ts
p3rid0t: Yes, that's right

src/formats/onenote.ts
p3rid0t: Hmm, your concern about this function is definitely true, however the OneNote importer is already pretty big (more functions than the most other importers), and I'm worried that splitting into multiple functions make the code harder to understand. Or perhaps, we could split some functions off into into multiple files like the Apple Notes importer. Please let me know what you think.

tgrosinger: I don't think splitting into functions is necessarily going to make the code bigger. And removing the cases where we are stepping around type safety will make the code easier to understand and safer.

p3rid0t:
I just pushed a commit with fixes to the comments mentioned here. Sorry it took a while, hope you don't mind.

src/formats/onenote.ts
```diff
@@ -142,150 +147,197 @@ export class OneNoteImporter extends FormatImporter {
...
+			for (let i = 0; i < parentGroup.sectionGroups.length; i++) {
+				await this.fetchNestedSectionGroups(parentGroup.sectionGroups[i]);
+			}
+		}
+	}
+	
+	// Renders a HTML list of all section groups and sections
+	renderHierarchy(entity: OnenoteEntityHierarchyModel, parentEl: HTMLElement) {
+		if ('sectionGroups' in entity) {
+			for (const sectionGroup of entity.sectionGroups as SectionGroup[]) {
```
tgrosinger, line 170: This assertion can be avoided if we change the method signature:

```
	renderHierarchy(entity: SectionGroup | Notebook, parentEl: HTMLElement) {
		if (entity.sectionGroups) {
			for (const sectionGroup of entity.sectionGroups) {
```

tgrosinger: The same can be done for `sections` below on line 186.

src/formats/onenote.ts
```diff
@@ -142,150 +147,197 @@ export class OneNoteImporter extends FormatImporter {
...
 		}
 	}
 
-	async processFile(progress: ImportContext, sectionFolder: TFolder, outputFolder: TFolder, content: string, page: OnenotePage) {
-		try {
-			const splitContent = this.convertFormat(content);
-
-			if (this.importDrawingsOnly) {
-				// TODO, when InkML support is added
+	insertPagesToSection(pages: OnenotePage[], sectionId: string, parentEntity?: OnenoteEntityHierarchyModel | undefined) {
```
tgrosinger, line 275: Similar to the previous comment, please update the signature of this method and remove the type assertions from this method.

```
insertPagesToSection(pages: OnenotePage[], sectionId: string, parentEntity?: Notebook | SectionGroup) {
```

src/formats/onenote.ts
tgrosinger: Looks like these still need to be moved.

src/formats/onenote.ts
tgrosinger: I like how this has been split up into separate functions, thank you! I'm wondering now if this function is actually needed at all. There are 4 places where we call it, each time with a different parentEntity type. Let's just call the different helper functions directly.

src/formats/onenote.ts
```diff
@@ -357,17 +408,121 @@ export class OneNoteImporter extends FormatImporter {
...
+						}
+					}
+				}
+				break;
+			}
+		}
+		return returnPath;
+	}
+	
+	private searchSectionGroups(entityID: string, currentPath: string, sectionGroups: SectionGroup[]): string | null {
```
tgrosinger, line 484: If you change this method signature to this:

```
private searchSectionGroups(entityID: string, currentPath: string, sectionGroups: SectionGroup[] | OnenoteSection[]): string | null {
```

Then you can remove `searchSections` and replace the places where it was called with this function, and change the last parameter to `parentEntity.sections!`.

p3rid0t:
Pushed another commit with the small type tweaks mentioned in the comments :)

tgrosinger:
Thanks @p3rid0t. I added another commit that applies some formatting fixes and a couple other tweaks to the method types that we were discussing.

I think this looks good to me now. Thank you.

Issue #229, opened 2024-04-07
[Bear] Assets are not imported correctly


tanob:
If you're importing a bear2bk file with notes that reference assets with the same base name then only the first asset is imported and all notes will reference the same asset.

For example if you're importing the following bear2bk file with two different assets called `image_1.png`, only one will be imported because the importer is putting all the assets into the same `assets` folder:
```shell
$ unzip -l Bear\ Notes\ 2024-03-31\ at\ 22.29.bear2bk
Archive:  Bear Notes 2024-03-31 at 22.29.bear2bk
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  04-07-2024 12:55   Bear Notes 2024-03-31 at 22.29.bear2bk/
        0  04-07-2024 12:07   Bear Notes 2024-03-31 at 22.29.bear2bk/BB duvida.textbundle/
      452  03-31-2024 22:29   Bear Notes 2024-03-31 at 22.29.bear2bk/BB duvida.textbundle/text.md
      514  03-31-2024 22:29   Bear Notes 2024-03-31 at 22.29.bear2bk/BB duvida.textbundle/info.json
        0  04-07-2024 12:07   Bear Notes 2024-03-31 at 22.29.bear2bk/BB duvida.textbundle/assets/
    23699  06-10-2020 17:52   Bear Notes 2024-03-31 at 22.29.bear2bk/BB duvida.textbundle/assets/image_1.png
        0  04-07-2024 12:07   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/
      840  03-31-2024 22:29   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/text.md
      514  03-31-2024 22:29   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/info.json
        0  04-07-2024 12:07   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/assets/
    27147  06-10-2020 18:01   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/assets/image_2.png
   135510  06-10-2020 18:01   Bear Notes 2024-03-31 at 22.29.bear2bk/Boleto mockups.textbundle/assets/image_1.png
[...]

PR #231, opened 2024-04-07
Fix Bear asset import when they have the same base name (#229)

tanob:


tgrosinger:
I've taken this idea and run with it, modifying it so that it uses `getAvailablePathForAttachment` instead ([commit](https://github.com/obsidianmd/obsidian-importer/commit/cb6537af46c768572558e5910ca12f6f756f6e89)). I'm going to close this PR now. Thanks.

Issue #236, opened 2024-04-13
[Roam] import settings/feature requests


dropcase:
Hi, I'm working on importing my Roam graph (~7200 notes) and looking for a few updates to help with the migration.

- Option to replace "invalid" characters in file names with a default (or user-choice option), possibly with adding the original naming as the `title` value in YAML frontmatter
- Option for removing one level of indented bullets, tested as working with this sed string: `s/  \* /* /g`
- Option for removing bullets from headers, using these for H3 and H2 as an example: `s/\* ###/###/g` and `s/\* ##/##/g`
- Option to use the `create-time` and `edit-time` from the [Roam JSON Schema](https://roamresearch.com/#/app/help/page/Nxz8u0vXU) in YAML frontmatter as `created` and `updated` dates

I'll share updates if I'm able to test anything on my local system. Thanks!

dropcase:
Updated the PR with a new commit, added TZ and corrected title YAML setup

tgrosinger:
PR merged!

PR #242, opened 2024-04-24
Importer feature updates for Roam, #236

dropcase:
First attempt at adding features to the importer, related to Roam JSON import.

1. Added YAML frontmatter options:
- title option: adds the imported title with all characters
- created/edited timestamps option: checks the page- and block-level timestamps, uses the oldest for `created` and newest for `edited`
2. Extended the `jsonToMarkdown` function to add more values, could use some work (maybe pass an array instead?)
3. Changed the default replacement for illegal filename characters from `''` to `'_'`

I tested with my Roam graph of 7100 notes, and didn't notice an increase in time to complete.

Still TODO: the timestamps are UTC, need some help sorting out time zone stuff.

Let me know if there are any issues or edits I need to make, thanks!


dropcase:
I think the TZ updates are working, they look like they're matching my local time when I import the JSON from Roam

src/formats/roam-json.ts
```diff
@@ -255,8 +315,39 @@ export class RoamJSONImporter extends FormatImporter {
 		return blockText;
 	};
 
-	private async jsonToMarkdown(graphFolder: string, attachmentsFolder: string, json: RoamPage | RoamBlock, indent: string = '', isChild: boolean = false): Promise<string> {
+	// setup to hold the newest and oldest timestamp value from a given page
+	newestTimestamp: number = 0;
+	oldestTimestamp: number = 0;
+
+	private async jsonToMarkdown(graphFolder: string, attachmentsFolder: string, json: RoamPage | RoamBlock, indent: string = '', isChild: boolean = false, checkYAMLoptions : boolean = false, runYAMLoptions: boolean = true, setTitleProperty: string, createdTimestamp: number, updatedTimestamp: number): Promise<string> {
```
tgrosinger, line 322: The `checkYAMLoptions` and `runYAMLoptions` parameters are confusing and seem redundant.

`runYAMLoptions` seems to just be `!isChild`, so I think we can remove that one.

I'm not clear on why `checkYAMLoptions` is necessary since this function seems to be changing it's value based on global config anyway.

Can we remove both of these?

src/formats/roam-json.ts
```diff
@@ -86,7 +112,41 @@ export class RoamJSONImporter extends FormatImporter {
 				}
 				const filename = `${graphFolder}/${pageName}.md`;
 
-				const markdownOutput = await this.jsonToMarkdown(graphFolder, attachmentsFolder, pageData);
+				// if title option is enabled
+				let YAMLtitle: string;
```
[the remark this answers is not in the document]
tgrosinger: We're already passing pageData into `jsonToMarkdown`. Can we extract this value in that function so we don't need to pass it as a parameter?

src/formats/roam-json.ts
```diff
@@ -86,7 +112,41 @@ export class RoamJSONImporter extends FormatImporter {
...
+				if (this.titleYAML === true) {
+					YAMLtitle = pageData.title;
+				} else {
+					YAMLtitle = '';
+				}
+				// if timestamp option is enabled
+				// set up numbers to pass
+				let pageCreateTimestamp: number;
```
tgrosinger, line 124: Can we extract these timestamps in `jsonToMarkdown` so we don't need to pass them as parameters?

src/formats/roam-json.ts
```diff
@@ -266,10 +357,56 @@ export class RoamJSONImporter extends FormatImporter {
...
+		// add YAML frontmatter, if enabled
+		if (checkYAMLoptions === true && runYAMLoptions === true) {
+
+			let timeCreated = this.oldestTimestamp;
+
+			frontMatterYAML.push('---');
+
+			// if "add title" option enabled, quotes added to prevent errors in frontmatter
+			if (this.titleYAML === true) {
+				frontMatterYAML.push('title: "' + setTitleProperty +'"');
```
tgrosinger, line 383: I think the quotes would be much easier to read if we rewrote it like this:

```suggestion
				frontMatterYAML.push(`title: "${setTitleProperty}"`);
```

src/formats/roam-json.ts
```diff
@@ -266,10 +357,56 @@ export class RoamJSONImporter extends FormatImporter {
...
+				frontMatterYAML.push('updated: ' + formatUpdateDate);
 			}
+
+			frontMatterYAML.push('---');
+			runYAMLoptions = false;
 		}
 
+		// Add frontmatter YAML to the top of the markdown array
+		// reverse it before pushing
+		frontMatterYAML.reverse();
```
tgrosinger, line 406: It would be much more efficient to join our string and just add one item to the markdown.

```
markdown.unshift(frontMatterYAML.join('\n'));
```

src/formats/roam/utils.ts
```diff
@@ -8,10 +8,11 @@ const windowsTrailingRe = /[\. ]+$/;
 const startsWithDotRe = /^\./; // Regular expression to match filenames starting with "."
 const squareBracketOpenRe = /\[/g; // Regular expression to match "["
 const squareBracketCloseRe = /\]/g; // Regular expression to match "]"
+const ReplaceNoDir = '_';
```
tgrosinger, line 11: Can you please explain what this change is doing?

package.json
```diff
@@ -26,11 +26,13 @@
 		"typescript": "4.7.4"
 	},
 	"dependencies": {
+		"@types/moment-timezone": "^0.5.30",
 		"@zip.js/zip.js": "2.7.20",
 		"joplin-turndown-plugin-gfm": "1.0.12",
+		"moment-timezone": "^0.5.45",
```
tgrosinger, line 32: Can this change be accomplished without adding this library? Obsidian already ships with moment, so we'd like to avoid adding another version if possible.

tgrosinger (changes requested):
Please be sure to run the eslint formatter on your changes.

package.json
dropcase: I wasn't sure how to get a TZ-aware function in place since `moment-timezone` is separate (and requires `moment` on its own). I'd also rather not add it, but couldn't sort out how to get the Unix timestamp into the local time (based on Daylight Saving rules too). Open to options if you know of anything that would work through moment itself - that's what I started with

src/formats/roam/utils.ts
dropcase: I forgot to remove it - I was going to set it up to have a replacement character in a dropdown list in the options for importing, so any illegal filename characters would be replaced by it. For now I'll take it out (maybe in a future version)

src/formats/roam-json.ts
dropcase: agreed, thanks!

src/formats/roam-json.ts
```diff
@@ -266,10 +357,56 @@ export class RoamJSONImporter extends FormatImporter {
...
+		switch (!checkYAMLoptions) {
+			case this.fileDateYAML:
+			case this.titleYAML:
+			// add other options here
+			// case this.{option-name}:
+				checkYAMLoptions = true;
+		}
+
+		// add YAML frontmatter, if enabled
+		if (checkYAMLoptions === true && runYAMLoptions === true) {
```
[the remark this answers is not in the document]
dropcase: thanks, overabundance of caution with types. cleaned up

dropcase:
@tgrosinger thanks for all the feedback! Working on updates and corrections, as you can tell I'm getting through the "make it work" phase - now on to "make it right" and will resubmit when done

src/formats/roam-json.ts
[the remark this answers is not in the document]
dropcase: I wasn't sure how to access it through the function, when I tried to pull the `title` it didn't show available unless I got it directly from `pageData` beforehand. Is it because `jsonToMarkdown` can get either a page _or_ a block?

Trying to use `json.title` shows "Property 'title' does not exist on type 'RoamPage | RoamBlock'." Agreed, if there's another way to access it then I'd rather do it through `jsonToMarkdown`

src/formats/roam-json.ts
dropcase: I wasn't able to, that's how I initially tried to get the dates. It didn't allow for the page dates to influence the timestamps, just each block-level. That ended up causing an issue where the `created` date could show as the oldest block's date, even if the page was created previous to that without a block-level beforehand (for example, the Daily Note being created automatically).

If there's a way to get that, please let me know. I'd rather try to do everything under `jsonToMarkdown` without needing to add parameters

src/formats/roam-json.ts
dropcase: Probably (at least getting passed as parameters), and it's mostly due to my inexperience. The idea was to set `runYAMLoptions` up to run once per note/page (or like you said, `!isChild`), and check against the user picking one or more YAML options (currently for `title`, and the `created`/`updated` choices). That's what the switch statement (near line ~365ish) is doing:

```
switch (!checkYAMLoptions) {
	case this.fileDateYAML:
	case this.titleYAML:
	// add other options here
	// case this.{option-name}:
		checkYAMLoptions = true;
}
```

I'll see if I can rework all of that under `jsonToMarkdown` and make it clearer (or remove as much as possible)

dropcase: Ok, I think I got it working better now. I rewrote the switch and removed both `checkYAMLoptions` and `runYAMLoptions` from the `jsonToMarkdown` function - is it better as an if (since the default is set to `false`) or a ternary? I have it as an `if` at the moment:

```
if (this.fileDateYAML || this.titleYAML) {
	checkYAMLoptions = true;
}
```
vs
```
checkYAMLoptions = (this.fileDateYAML || this.titleYAML) 
	? true 
	: false;
```

and then updated this statement to remove `runYAMLoptions`:

```
if (checkYAMLoptions && !isChild) {
```

package.json
dropcase: Switched back to using just `moment` (and imported from `obsidian`), the timestamps showing in Roam seem to be consistent with my TZ so importing _should_ keep it correct. The more I was digging through it, the less consistent Roam was with dates - this should be more than good enough for the import, and can always be manually adjusted for outliers

dropcase:
@tgrosinger I just pushed the most recent changes, let me know how this looks for progress. Tested as working on my system (tried with two different exports). 

I also have a sample Roam export in json that I can share (anonymized as much as possible) that should work if someone needs it

package.json
tgrosinger: Thank you, can you remove the package-lock and package.json changes from this PR then?

src/formats/roam-json.ts
```diff
@@ -255,8 +307,45 @@ export class RoamJSONImporter extends FormatImporter {
...
+		// if undefined, set newestTimestamp to the value of updatedTimestamp
+		this.newestTimestamp = (!jsonEditTime || updatedTimestamp > jsonEditTime)
+    		? updatedTimestamp
+    		: jsonEditTime;
+
+		// if the create time is defined, set oldestTimestamp to the lower of the createdTimestamp value or jsonCreateTime
+		// else, set oldestTimestamp to the value of createdTimestamp
+		if (jsonCreateTime !== undefined) {
+			if (createdTimestamp > 10) { // passed as a 0
+				this.oldestTimestamp = (createdTimestamp < jsonCreateTime)
```
tgrosinger, line 338: Oh actually I think we could make this even simpler.

```
this.oldestTimestamp = Math.min(createdTimestamp, jsonCreateTime);
```

src/formats/roam-json.ts
```diff
@@ -86,7 +111,34 @@ export class RoamJSONImporter extends FormatImporter {
...
+					// type check both for numbers, set to 0 if there's a type mismatch
+					if (typeof pageCreateTime === 'number') {
+						pageCreateTimestamp = pageCreateTime;
+					}
+
+					if (typeof pageEditTime === 'number') {
+						pageEditTimestamp = pageEditTime;
+					}
+				} 
+				else {
```
tgrosinger, line 135: This else case can be removed. The vars already hold zeros.

src/formats/roam-json.ts
```diff
@@ -266,10 +355,48 @@ export class RoamJSONImporter extends FormatImporter {
...
+
+		// once processing children is completed, add the YAML to the top
+		// check if any YAML options are set
+		if (this.fileDateYAML || this.titleYAML) {
+			checkYAMLoptions = true;
+		}
+
+		// add YAML frontmatter, if enabled
+		// only run on the initial set, skip if child 
+		if (checkYAMLoptions && !isChild) {
```
tgrosinger, line 370: Can we remove `checkYAMLoptions` entirely?

```suggestion
		if ((this.fileDataYAML || this.titleYAML) && !isChild) {
```

src/formats/roam-json.ts
```diff
@@ -266,10 +355,48 @@ export class RoamJSONImporter extends FormatImporter {
...
+
+				frontMatterYAML.push('created: ' + formatCreateDate);
+				frontMatterYAML.push('updated: ' + formatUpdateDate);
 			}
+
+			frontMatterYAML.push('---');
 		}
 
+		// Add frontmatter YAML to the top of the markdown array
+		markdown.unshift(frontMatterYAML.join('\n'));
```
tgrosinger, line 398: Can this be moved into the if-block above? Just below `frontMatterYAML.push('---');`?

src/formats/roam/utils.ts
```diff
@@ -11,7 +11,7 @@ const squareBracketCloseRe = /\]/g; // Regular expression to match "]"
 
 export function sanitizeFileNameKeepPath(name: string) {
 	return name
-		.replace(illegalReNoDir, '')
+		.replace(illegalReNoDir, '_')
```
tgrosinger, line 14: I'm not sure why this is changing. Can you explain why this is needed?

tgrosinger (changes requested):
Thanks for the changes @dropcase. Just a few more mostly stylistic changes.

src/formats/roam-json.ts
tgrosinger: To directly answer your question, you don't need an if or ternary. It can just be:

```
checkYAMLoptions = this.fileDataYAML || this.titleYAML
```

However, see my comment below. I think we can remove `checkYAMLoptions` completely and use these two dependent vars directly.

src/formats/roam-json.ts
dropcase: Nice, thanks! Yeah, as long as there aren't too many options added (potentially) that would make that confusing, this is a better approach

src/formats/roam-json.ts
dropcase: Done, better to have it there so it's not doing anything when the options aren't set for the YAML header

src/formats/roam/utils.ts
dropcase: It's not needed (and can remove it), I thought it was good to have a replacement rather than just deleting the characters. I'll take it out, and might add a feature later to optionally allow choosing a replacement character

dropcase:
Changes pushed, eslint was clean - just tested and it's looping the YAML header, need to check what's happening there

dropcase:
My fault, when I pasted in the change it took out the "extra" parentheses. Corrected, pushed, and tested as working!

src/formats/roam-json.ts
dropcase: That's much cleaner, much appreciated!

package.json
dropcase: Done, cleaned up any references to `moment-timezone` (and the types package that went with it)

PR #270, opened 2024-06-14
OneNote notebook dates

darthwalsh:
Fix two issues in the Import screen:
Without OData $select the createdDateTime is always undefined.
Show last modified date, and sort so the most recently changed is on top.

Tested manually on my account with a dozen notebooks.


darthwalsh:
BTW, let me know if I should have created issues for these first!

I'm interested to a couple other issues I ran into, but I figured I could get my feet wet with a trivial bug-fix.

tgrosinger:
This looks great, thank you @darthwalsh!

PR #277, opened 2024-07-07
OneNote page creation time

darthwalsh:
File ctime was being set to the lastModified.
The typo was hard to see, so try to make the code more readable.

Tested manually on a page I manually set the creation date far in the past.


tgrosinger:
Merged with a slight modification to add back the `page?`. Thank you!

PR #301, opened 2024-09-27
Extracted and integrated creation date from Google Keep's HTML.

sayzard:
When importing notes from Google Keep, the creation date is not included. 
To solve this issue, after parsing the JSON file, 
I implemented a method to extract the creation date from the corresponding HTML file with the same name.

### DIFF
--- a/src/formats/keep-json.ts
+++ b/src/formats/keep-json.ts
@@ -6,6 +6,7 @@ import { serializeFrontMatter } from '../util';
 import { readZip, ZipEntryFile } from '../zip';
 import { KeepJson } from './keep/models';
 import { sanitizeTag, sanitizeTags, toSentenceCase } from './keep/util';
+import { extractDateFromHtml } from './keep/html-parser';
 
 
 const BUNDLE_EXTS = ['zip'];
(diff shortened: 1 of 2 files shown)


tgrosinger:
Hello, thanks for providing this pull request.

I don't think I understand why we need to parse the HTML file to get this information. It seems like the information that we need is in the json already, matches what is in the HTML, and is set correctly when importing the files.

Can you please let us know what operating system you are on where this issue is occurring?

tgrosinger (commented):
Hello, thanks for providing this pull request.

I don't think I understand why we need to parse the HTML file to get this information. It seems like the information that we need is in the json already, matches what is in the HTML, and is set correctly when importing the files.

Can you please let us know what operating system you are on where this issue is occurring?

sayzard:
> Hello, thanks for providing this pull request.
> 
> I don't think I understand why we need to parse the HTML file to get this information. It seems like the information that we need is in the json already, matches what is in the HTML, and is set correctly when importing the files.
> 
> Can you please let us know what operating system you are on where this issue is occurring?

The last time I tried, everything imported with the current date and time, so I thought the feature wasn’t working. After trying again today following the update, it’s working fine. I apologize for the confusion.

tgrosinger:
Oh good, glad to hear it's working. Thank you.

Issue #304, opened 2024-10-02
Bear importer only imports attachments


jezper:
Any issue with the Bear importer or is this a user error on my behalf?

Trying to import a .bear2bk into a fresh Vault and only getting the attachments imported, no notes. I get two empty folders for trash and archive, but nothing in there either.

rtud:
Seeing the same behavior on my end.

tgrosinger:
Hello, thanks for reporting this issue. I've got a couple exports from Bear that we use for testing, and both are importing the notes as expected. Would one of you be able to share an export that reproduces this issue that I can use for fixing this issue?

tgrosinger:
Could you please also let us know what operating system you are using?

HybridRbt:
I am on macOS 15.0. My Bear is on v2.3.3. I will see if I can create a shareable backup of the notes.

rtud:
System 1: macOS 15.0, Bear 2.3.3, Obsidian 1.6.7, Obsidian Importer Plugin 1.6.5
System 2: Arch Linux via LTS kernel, Obsidian 1.6.7-1 via pacman, exported Bear notes in `.bear2bk` format from Bear 2.3.3

Thank you!

tgrosinger:
I have made some changes to the Bear importer to improve how we handle attachments and duplicate file paths (#305). Would any of you be willing to give it a test and report if it resolves this issue?

1. Download this zip: [main.zip](https://github.com/user-attachments/files/17247481/main.zip)
2. Unzip the file.
3. Open your Obsidian vault, replace `<your-vault>/.obsidian/plugins/obsidian-importer/main.js` with this file.
4. Close and reopen Obsidian.

Thank you!

jezper:
@tgrosinger 

That did the trick (at least for me!)

![image](https://github.com/user-attachments/assets/836749eb-b8b2-476d-812d-d0a2b2ffa4c2)

HybridRbt:
I can also confirm that the new main.js file has fixed it. 
<img width="563" alt="截屏2024-10-04 下午11 47 36" src="https://github.com/user-attachments/assets/31fb255a-819f-4e9c-b2c5-b9d31e61ca64">

tgrosinger:
This problem should be resolved in the latest release. Thanks for helping test!

PR #305, opened 2024-10-03
Bear attachments

tgrosinger:
This adds a new utility function to help with finding an available path for an attachment, with consideration for attachment paths that maybe be reserved but not yet created.

It also changes the Bear importer to make use of this new function instead of iterating through the zip files an additional time just to pull out the attachments.

If this looks good we can extend this pattern to the other importers as well.

### DIFF
--- a/src/format-importer.ts
+++ b/src/format-importer.ts
@@ -1,5 +1,5 @@
 import { App, normalizePath, Platform, Setting, TFile, TFolder, Vault } from 'obsidian';
-import { getAllFiles, NodePickedFile, NodePickedFolder, PickedFile, WebPickedFile } from './filesystem';
+import { getAllFiles, NodePickedFile, NodePickedFolder, parseFilePath, PickedFile, WebPickedFile } from './filesystem';
 import { ImporterModal, ImportContext, AuthCallback } from './main';
 import { sanitizeFileName } from './util';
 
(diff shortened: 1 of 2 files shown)


src/format-importer.ts
```diff
@@ -139,6 +139,45 @@ export abstract class FormatImporter {
...
+	 * which adds two key adjustments to aid Importer:
+	 *   - Use the provided `sourcePath` even if the file doesn't exist yet.
+	 *   - Avoid duplicating a list of provided filesnames that do not yet exist, but will in the future.
+	 *
+	 * @param filename Name of the attachment being saved
+	 * @param claimedPaths List of filepaths that may not exist yet but will in the future.
+	 * @returns Full path for where the attachment should be saved, according to the user's settings
+	 */
+	async getAvailablePathForAttachment(filename: string, claimedPaths: string[]): Promise<string> {
+		const outputFolder = await this.getOutputFolder();
```
lishid, line 157: Might be better if this is cached somehow?

src/format-importer.ts
```diff
@@ -139,6 +139,45 @@ export abstract class FormatImporter {
...
+		// Use getAvailablePathForAttachments because it can give us the configured output path.
+		//@ts-ignore
+		const prelimOutPath = await this.vault.getAvailablePathForAttachments(basename, extension, sourceFile);
+		const parsedPrelimOutPath = parseFilePath(prelimOutPath);
+
+		// Increase number until the path is unique.
+		let i = 1;
+		let outputPath = prelimOutPath;
+		while(claimedPaths.includes(outputPath) || !!this.vault.getAbstractFileByPath(outputPath)) {
+			outputPath = addExt(parsedPrelimOutPath.basename + ' ' + i, parsedPrelimOutPath.extension);
```
lishid, line 174: I don't think this is necessary. An easy alternative is to define extension earlier:
```ts
let extension = parsedPrelimOutPath.extension ? '.' + parsedPrelimOutPath.extension : '';
```

And just concatenating it directly.

tgrosinger:
Got some good reports that this change has fixed the import issue over here #304.

src/format-importer.ts
tgrosinger: Added class-level caching to `getOutputFolder`.

lishid (approved):

Issue #309, opened 2024-10-07
Attachments from Apple Notes create a folder with a space in front of the name


lorddev:
Attachments from Apple Notes create a folder with a space in front of the name: ` Attachments`.
![image](https://github.com/user-attachments/assets/ef55b6f4-1092-4de7-b1fb-3684460a1f80)

tgrosinger:
This should be resolved in 1.6.8. Please let us know if it is not.

Issue #192, opened 2023-12-21
[Apple Notes] Image file name conflict results in "(error reading attachment)"


solidgonk:
When importing from Apple Notes, "(error reading attachment)" may occur.

I suspect that this is related to a naming conflict of images when they're placed in the same attachments folder. I have a dozen or so images named "pasted graphic.png" (you can search this in the Apple Notes search field), and they are all coming up with that error.

I'm not sure if the solution here is to rename the filename (i.e. "pasted graphic-01.png") during import?

solidgonk:
This also appears to affect other attachments with duplicate names, such as html.

Johannnnes:
Same here, with .TIFF, amongst others

alexrecuenco:
Same here, 

I have a myriad of PDF documents that I just scanned, so they are called "Scanned documents" and it failed to import many of them correctly when they are on the same note.

Also, when fixing this, keep in mind that "filenames" in notes can have characters that are not valid filenames (like `/` or `\n`), and those notes work on iOS and, they don't work on macOS "sometimes"  (at least last time I tried, they might've fixed it in the last months, it was still broken around November 2023)

solidgonk:
Appears to be fixed in #310

tgrosinger:
Great, thank you for the update!

PR #310, opened 2024-10-07
OneNote, Apple Notes: Fix duplicate attachment name importing

### LINKED ISSUE #298: Notes are skipped on import due to attachment name duplicates
This has been mentioned before in https://github.com/obsidianmd/obsidian-importer/issues/229 and https://github.com/obsidianmd/obsidian-importer/pull/231

Notes are skipped on import due to attachment name duplication (this is a subset of thousands of notes I wanted to migrate):

image

On some notes that have been imported images have been swapped out with other images because filenames in attachments from Bear are the same.

tgrosinger:
Fixes #298 for OneNote.


tgrosinger:
This should also fix #309.

solidgonk:
Appears to fix #192 also

Issue #321, opened 2024-11-15
Import not Successful because of File Name


keichhorn:
Failed: "Zus: Betriebswirtschaftliche Grundlagen Zusammenfassung" because Error: File name cannot contain any of the following characters: * " \ / < > : | ?

I've already renamed lots of Notes. But some of them still slip through.
Problem: The Note in Question is not imported into Obsidian + If too many Notes failed to Convert, the Importer just stops (and does not import the remaining Notes)

I suggest, that if a filename has invalid characters, that the characters are replaced with a standard symbol (like ".").

This way, everything is converted. And the User has to check anyway, if his/her notes have converted completely and are still usable

tgrosinger:
I'm going to merge this with #291 since it looks very similar.

Issue #291, opened 2024-08-27
Improve Special Character Handling in Obsidian Note Titles


AltoRetrato:
**Obsidian note titles can't have the special characters** `* " \ / < > : | ?`  because a note title also serves as its file name. Emoji in file names might not be supported as well (e.g., when using Dropbox to sync the Obsidian vault). Currently, the Importer just removes these special characters, but better alternatives would be:

1. Allow the user to choose a custom replacement character or string (e.g., `-` `_` ` -- `) for these special characters, and, if the user wants, replace emojis as well.
2. Replace the special characters with similar-looking Unicode or ASCII characters, such as ✱ ✲ ＊ ⁎ ※ ' “ ” ⧵ ∖ ＼ ∕ ⁄ ‹ ⟨ 〈 ❮ › ⟩ 〉 ❯ ∶ ꞉ ː ։ ∣ ¦ ┃ ǀ ｜ ？ 

(FWIW, I tried importing the notes with Yarle, which has a "Global search and replace (regexp) in note title and content" feature, but it didn't work for me.)

tgrosinger:
I would advise against trying to use similar looking characters because that is likely to lead to more confusion in the long run.

I would prefer that we stick to an approach similar to what the Obsidian Clipper does, where unsafe characters are just removed from the name. (See https://github.com/obsidianmd/obsidian-clipper/blob/097dda43a0655644f053235e822aec9b1bfc70f8/src/utils/filters/safe_name.ts)

PR #324, opened 2024-11-17
Replace illegal characters in filenames with Unicode lookalikes

WetHat:
This pull request addresses issue #321 and #291 complaining about special characters in filenames.

Filenames in Obsidian are of primary importance as they are also used as note titles. Therefore, the current brute force approach of filename sanitization destroys note titles whenever illegal characters are involved.  This cause unnecessary manual cleanup effort. 

This pull request upgrades `sanitizeFileName` to replace illegal characters with legal Unicode characters that look similar.


tgrosinger:
Hi @WetHat, thank you for the contribution. I don't think we should head in this direction though as it will likely lead to increased confusion with the unexpected characters. I'll leave a comment in the ticket with some more guidance.

Issue #281, opened 2024-07-14
[Notion] Import of equations with the cases environment do not work


tooxo:
Hey, I tried to import my notion workspace into obsidian using this tool, but every equation using the cases environment do not import completely. 
I recreated a minimal example to recreate this.

This is how it looks in Notion:
![image](https://github.com/user-attachments/assets/fe0d4451-dd74-4d97-aa67-be656168722f)

With the following LaTeX code:
```latex
f(x) = \begin{cases} 
1: &\text{test} \\
0.5: &\text{else} 
\end{cases}
```

This is how it imports into Obsidian:
![image](https://github.com/user-attachments/assets/3df583cb-b9b1-4327-8d36-01a6deb10d66)

This is the notion workspace export I used.
[cf8a7e57-5af5-4e14-88db-02fbc7db48f1_Export-4dccaf41-b1e3-4ddd-9924-0447a1573730.zip](https://github.com/user-attachments/files/16216563/cf8a7e57-5af5-4e14-88db-02fbc7db48f1_Export-4dccaf41-b1e3-4ddd-9924-0447a1573730.zip)

Issue #289, opened 2024-08-27
[Notion]: breaks with additional cases inside a math formula


Jrol123:
# Before
![image](https://github.com/user-attachments/assets/73b1bae2-7835-4c2d-b682-035a1d53450d)
![image](https://github.com/user-attachments/assets/958198f3-9860-4aab-84f0-7eef10deb808)
# After
![image](https://github.com/user-attachments/assets/92c22ce0-3a19-49de-8e2e-2afc50e72388)
![image](https://github.com/user-attachments/assets/dcbbd109-81ab-4b32-a28b-7fe8ba6742fe)
![image](https://github.com/user-attachments/assets/79ef32e1-d660-4910-a486-d8138385c11a)

Sometimes it even breaks on a single char formulas, like` \maths{\mu}`

tgrosinger:
@Jrol123, could you please update to the latest version of the Importer plugin and test if this issue has been resolved? Thank you!

Jrol123:
> @Jrol123, could you please update to the latest version of the Importer plugin and test if this issue has been resolved? Thank you!

Yes, now it works almost every time!
![image](https://github.com/user-attachments/assets/a1687239-7667-421d-86e3-db0674d8298d)

Although it was kinda scary to look at the pre-render of it (img 2)
![image](https://github.com/user-attachments/assets/5201201b-2246-4218-94e6-18523e3240ff)
 At the render (img 1) it looks great!
 
 Thank you very much!
 
 ----
 
 But there's still some bugs left, like this:
 Preview:
![image](https://github.com/user-attachments/assets/45ebe31f-46a1-41df-8d75-cefe340e4f60)
Pre-render:
![image](https://github.com/user-attachments/assets/01774721-fa3b-41b2-bc9b-3720c266c5c9)

 ----
 
 And as a little idea, it would be cool, if importer changed Notion-s KaTeX symbols to the Obsidian-s version of it (like \exist -> \exists, etc)
 
![image](https://github.com/user-attachments/assets/3557e7e5-9901-4d44-9c5c-840385016327)


Again, thank you very much for your work!

Jrol123:
And one more thing:

TAB in pre-render results in non-rendering:
Pre-render:
![image](https://github.com/user-attachments/assets/1ffc3db6-8d81-4e73-9596-5f017f39e23c)
Result:
![image](https://github.com/user-attachments/assets/42f49703-0af7-471c-96b5-195c50628414)

[...]

PR #329, opened 2024-12-05
[Notion] fix #259 math inside callouts; fix cases #281; fix display math

### LINKED ISSUE #259: [Notion] Math blocks inside Callouts don't render correctly
As an example, this is a callout in Notion:

Screenshot 2024-06-04 alle 23 20 41

and after import in Obsidian:

Screenshot 2024-06-04 alle 23 22 43

This issue repeats on all files, but only on callouts. Math blocks outside callouts render correctly

felciabatta:
Fixed issue #259 where math inside callouts didn't render correctly due to order in which DOM elements are processed and issues with `<style>` content creeping into markdown output.

Also spotted and fixed 2 other issues: 
1. Inline- and display-math were always converted to inline-math. This PR adds correct procedure for converting display-math.
2. Import would fail if the Notion page contained any callouts that did not have an icon (this was because of how the child elements of callout nodes were indexed).

**Before**
before

**After**
after

### DIFF
--- a/src/formats/notion/convert-to-md.ts
+++ b/src/formats/notion/convert-to-md.ts
@@ -55,11 +55,12 @@ export async function readToMarkdown(info: NotionResolverInfo, file: ZipEntryFil
 	replaceNestedTags(body, 'strong');
 	replaceNestedTags(body, 'em');
 	fixNotionEmbeds(body);
-	fixNotionCallouts(body);
+	// fixEquations must come before fixNotionCallouts
+	fixEquations(body);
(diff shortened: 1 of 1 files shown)


felciabatta:
I just noticed that this RP has also resolves issue #281 with the **math cases environment** (and potentially fixed #289 — the OP wasn't very clear, and there seem to be several issues in their example).

felciabatta:
Thank you for reviewing this; glad to help @tgrosinger!

PR #331, opened 2024-12-06
fixes #330 lead/trailing spaces in Notion math

### LINKED ISSUE #330: [Notion] Inline equation with leading or tailing spaces do not get imported correctly
In Notion leading or tailing spaces do not change anything in the appearance of the equation. If I import an equation like this into Obsidian the inline math breaks. 

Notion:
Notion

Obsidian:
Obsidian

@felciabatta (I hope the mention is okay, as you successfully solved the problem I had before :) )

Example Project: 
[1b19d8ae-4664-40db-bd85-608867c8bb80_Export-d70d856c-b32d-4af3-a1db-44944a1654cb.zip](https://github.com/user-attachments/files/18037450/1b19d8ae-4664-40db-bd85-608867c8bb80_Export-d70d856c-b32d-4af3-a1db-44944a1654cb.zip)

felciabatta:
Fixes issue #330 where leading/trailing spaces break LaTeX equations.

**Explanation:** This is due to the difference in math syntax in Notion, which allows leading/trailing spaces, unlike Obsidian-MD.

**Implementation:** Format math strings via regex that removes leading/trailing spaces and newlines.

# Example
**Before**
before

**After**
after

### DIFF
--- a/src/formats/notion/convert-to-md.ts
+++ b/src/formats/notion/convert-to-md.ts
@@ -235,15 +235,21 @@ function fixEquations(body: HTMLElement) {
 	for (const figEqn of figEqnEls) {
 		const annotation = figEqn.find('annotation');
 		if (!annotation) continue;
-		figEqn.replaceWith(`$$${annotation.textContent}$$`);
+		figEqn.replaceWith(`$$${trimMath(annotation.textContent)}$$`);
 	}
 	// Inline Equations
 	const spanEqnEls = body.findAll('span.notion-text-equation-token');
 	for (const spanEqn of spanEqnEls) {
 		const annotation = spanEqn.find('annotation');
 		if (!annotation) continue;
-		spanEqn.replaceWith(`$${annotation.textContent}$`);
+		spanEqn.replaceWith(`$${trimMath(annotation.textContent)}$`);
 	}
+}
+
+function trimMath(math: string | null | undefined): string {
+	// Trim trailing spaces & newlines in LaTeX math experessions.
+	let regex = new RegExp(/^[\s\r\n\\]*(.*?)[\s\r\n\\]*$/, 's');
+	return math?.replace(regex, '$1') ?? '';
 }
 
 function stripToSentence(paragraph: string) {
--- a/src/formats/notion/convert-to-md.ts
+++ b/src/formats/notion/convert-to-md.ts
@@ -235,21 +235,22 @@ function fixEquations(body: HTMLElement) {
 	for (const figEqn of figEqnEls) {
 		const annotation = figEqn.find('annotation');
 		if (!annotation) continue;
-		figEqn.replaceWith(`$$${trimMath(annotation.textContent)}$$`);
+		figEqn.replaceWith(`$$${formatMath(annotation.textContent)}$$`);
 	}
 	// Inline Equations
 	const spanEqnEls = body.findAll('span.notion-text-equation-token');
 	for (const spanEqn of spanEqnEls) {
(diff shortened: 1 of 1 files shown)


src/formats/notion/convert-to-md.ts
```diff
@@ -235,17 +235,24 @@ function fixEquations(body: HTMLElement) {
...
 		if (!annotation) continue;
-		spanEqn.replaceWith(`$${annotation.textContent}$`);
+		spanEqn.replaceWith(`$${formatMath(annotation.textContent, true)}$`);
 	}
 }
 
+function formatMath(math: string | null | undefined, inline: boolean=false): string {
+	// Trim lead/trailing space in LaTeX math experessions.
+	// Removes empty lines, which can cause equations to break.
+	let regex = new RegExp(/^(?:[\s\r\n]|\\\\|\\\s)*(.*?)[\s\r\n\\]*$/, 's');
```
tgrosinger, line 252: I'm not completely familiar with math expressions, so apologies if this is a silly question.

Could we simplify this with something like `math.trim()`, and then possibly a simpler regex replacement afterwards?

What is the `\\\\` matching in this regex?

felciabatta: Originally I did use trim plus a regex for the escape sequences; I ended up combining them but I'm happy to change it back, as I can see how that might be clearer.

In LaTeX math: `\\` and `\ ` are the escapes for line-breaks and white-space, respectively.
Depending on the engine, `\\` line-breaks may or may not be parsed — Notion actually

felciabatta: In LaTeX math: `\\` and `\ ` are the escapes for line-breaks and white-space, respectively — `\\\\` and `\\\s` in the regex.
Not all LaTeX engines parse `\\` — Notion _does_, but Obsidian MD does not, hence the need to clean up some of the syntax.
It can cause issues in Obsidian (except in a supported environment, such as `cases` or `array`).

Originally I did use trim _plus_ a regex for the escape sequences; I ended up combining them to account for repetition:
The issue with `\ ` is as follows: consider `$f(x)=e^x \ $` which is ok in Notion but
1. this breaks in Obsidian so we `trim` trailing spaces to give `$f(x)=e^x \$`
2. this breaks again due to the escaped `\$`, so remove this to give `$f(x)=e^x $`
3. we're back to the trailing space issue so have to `trim` a second time leaving us with `$f(x)=e^x$`, which is now ok in Obsidian.

felciabatta: As a side tangent... arguably code like this is not particularly "good practice" in LaTeX and isn't reliably supported by all engines; however, Notion syntax is much more forgiving than Obsidian so there's no reason a user would need to be vigilant (including myself, as I found out the hard way).
The thing that got me looking into these bugs in the first place was because my Notion workspace contains a lot of math, with some pretty complex code, half of which breaks in the Importer conversion process (things get a lot worse when nested in callouts and bullets).

tgrosinger: Very helpful, thank you.

> In LaTeX math: \\ and \ are the escapes for line-breaks and white-space, respectively — \\\\ and \\\s in the regex.

This was the key I was missing. Would you mind adding this as a comment below the existing two comment lines?

tgrosinger (approved):
LGTM after the one request for a comment is added. Thank you!

src/formats/notion/convert-to-md.ts
felciabatta: Sure! All done :)

PR #340, opened 2024-12-20
fixes #337 broken ToC import in Notion

### LINKED ISSUE #337: [Notion] Table of Contents import does not work correctly
I know that there is no built-in table of content feature in obsidian, so it won't work correctly anyway, but the Notion Table of contents should be imported with links to the sections

| Notion | Obsidian |
|--------|--------|
| image | image | 

The TOC contents are just plain text while it should be of the format: `[[#<subtitle name here>]]` and they should be indented the same way that the notion one is.

felciabatta:
# Issue
As in #337, table of contents import is broken:
- there are no links
- the indentation structure is not preserved

# Solution
1. extend `convertLinksToObsidian` to format ToC heading links
2. recreate indentation using unordered lists by recursively appending each item


felciabatta:
# Before & After (Render)

toc-before
toc-after

# Before & After (Source)
toc-before-src
toc-after-src

src/formats/notion/convert-to-md.ts
```diff
@@ -434,13 +437,62 @@ function splitBrsInFormatting(body: HTMLElement, tagName: FormatTagName) {
...
-		if (link.getAttribute('href')?.startsWith('#')) {
-			link.setAttribute('href', '#' + link.textContent);
-		}
+
+function getTOCIndent(tocItem: Element | null): Number {
+	return Number(tocItem?.classList[1].slice(-1) ?? -1);
+}
+
+function appendTOCItem(itemNew: Element, itemPre: Element | null | undefined) {
+	if (!itemPre) return;
```
tgrosinger, line 446: Why does this function accept null or undefined for `itemPre` if it's going to then return without appending in those cases?

src/formats/notion/convert-to-md.ts
```diff
@@ -434,13 +437,62 @@ function splitBrsInFormatting(body: HTMLElement, tagName: FormatTagName) {
...
+	const indentNew = getTOCIndent(itemNew);
+	let indentPre = getTOCIndent(itemPre);
+	// itemPre will contain minimum 1 <span> element
+	if (indentNew > indentPre && itemPre.childElementCount == 1) {
+		const listNew = createEl('ul');
+		listNew.append(itemNew);
+		itemPre.append(listNew);
+	}
+	else if (indentNew > indentPre && itemPre.childElementCount == 2) {
```
tgrosinger, line 455: Can you add some notes about what these child elements are expected to be? Alternatively, documenting though code by destructuring the children into descriptive variable names could also work.

src/formats/notion/convert-to-md.ts
```diff
@@ -434,13 +437,62 @@ function splitBrsInFormatting(body: HTMLElement, tagName: FormatTagName) {
...
+		itemPre.lastElementChild?.append(itemNew);
+	}
+	else if (indentNew == indentPre) {
+		itemPre.parentElement?.append(itemNew);
+	}
+	else if (indentNew < indentPre) {
+		const listPre = itemPre.parentElement;
+		itemPre = listPre?.parentElement;
+		// recurse through parents until (indentNew < indentPre)
```
tgrosinger, line 464: Should this comment be "until (indentNew == indentPre)"?

tgrosinger (commented):
Thanks! Just a couple small questions.

felciabatta:
> Thanks! Just a couple small questions.

No worries, will have look through this later in the week :)

src/formats/notion/convert-to-md.ts
felciabatta: Corrected.

src/formats/notion/convert-to-md.ts
felciabatta: Good point. Originally, there was a possibility that `itemPre` could have be `undefined`, if recursing too far up the tree. However, the way `formatTableOfContents` is currently implemented, that should be impossible.

felciabatta: I've removed the `null | undefined` now, as it is redundant.

src/formats/notion/convert-to-md.ts
felciabatta: I have added some details under `@param itemPre` in the docstring.
Plus, I have assigned the children to a descriptive variable for clarity.

tgrosinger:
This is looking great. I just noticed though that "A Toggle Heading 2" and "A Toggle Heading 3" do not show up in the TOC. I'm not very familiar with Notion. Should those have been included?

felciabatta:
> This is looking great. I just noticed though that "A Toggle Heading 2" and "A Toggle Heading 3" do not show up in the TOC. I'm not very familiar with Notion. Should those have been included?

Oh, that's interesting, I did not notice!
However, this is expected: I just checked and those two toggle headings do not appear within Notion.

I think the reason is simply because they are "hidden" inside "A Toggle Heading 1", so the nested headings are interpreted as invisible to the TOC. 
That seems like a strange choice to me so I'm not sure if it's a feature it's or a bug in Notion, but it _is_ the native behaviour.

tgrosinger:
Got it, thanks for checking.

tgrosinger (approved):

Issue #341, opened 2024-12-23
OneNote embedded images with no title do not embed in Obsidian note


scooby359:
I have a OneNote notebook, with many pages with an image embedded. I'm trying to import these using Obsidian Importer.

Where the images have been imported from a file and have a name, the note imports to Obsidian as expected, with the image embedded in the file.

Where an image in the OneNote was copied and pasted, so has no file name, this does not import to Obsidian correctly. When using Obsidian Importer, the note imports but without any reference to the image. The image file does get imported separately and with a given name of the current datestamp.

Expectation is that the untitled image would be imported, given a name, and that image name would be embedded in the note.

Using Obsidian v1.7.7 for Windows, Importer v1.6.17.

tgrosinger:
Hello, can you please tell me more about how you are embedding the image in your OneNote note?

I tested this by using the OneNote webapp, and pasting an image from my clipboard into a note. When I used the Importer, an image called `Exported image 202412....png` was created and was embed in my note (`![Exported image](Exported image 202412....png)`)

Let me know if you're seeing something different, and what steps you took to get there. Thank you.

scooby359:
Hey there!

In OneNote, I've used a Windows computer, and a couple of methods to import pictures that are affected:

1. I've taken a partial screen shot using winkey + shift + s to click and drag around the area I want, then pasted that into OneNote. 
2. I've used the toolbar in OneNote to import a PDF, then chosen to print to file. This causes images of each page of the PDF to be added to the note.

In both cases, I've generally added some text, a URL, and a PDF file.

The output in OneNote for a screenshot looks something like this:
![image](https://github.com/user-attachments/assets/e5eb39ce-9af8-4671-a607-9f685bcb0f88)

When I use the Obsidian Importer, it's copied everything to the note except the image:
![image](https://github.com/user-attachments/assets/23ab3def-7c31-444f-93eb-4f54d2344316)

Looking at the raw source, there's nothing for the image:
![image](https://github.com/user-attachments/assets/a9222cb7-145b-4631-ac16-0989f77ae634)

The image has imported separately, with a timestamp as the file name:
![image](https://github.com/user-attachments/assets/848e26e8-1fbc-45fa-a188-282735a6ad2b)


Not all notes are affected - many have imported as expected. But I have hundreds of notes that haven't imported correctly, and without a relevant filename on the imported image, it's too difficult to manually correct the issue.

scooby359:
[...]

PR #345, opened 2024-12-27
#341 Missing content on OneNote import

scooby359:
Fixes bug where an object node may contain other content as child nodes, which gets lost when the node is converted to markdown. Moves child nodes to be sibling nodes to preserve them.

### DIFF
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -674,6 +674,12 @@ export class OneNoteImporter extends FormatImporter {
 		const videos: HTMLIFrameElement[] = pageElement.findAll('iframe') as HTMLIFrameElement[];
 
 		for (const object of objects) {
+			// Objects may contain child nodes which would be lost when the object is replaced by markdown.
+			// To preserve these, move any child items to be siblings of the object
+			while (object.firstChild) {
+				object.parentNode?.insertBefore(object.firstChild, object.nextSibling);
+			}
+
 			let split: string[] = object.getAttribute('data-attachment')!.split('.');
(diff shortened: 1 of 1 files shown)


src/formats/onenote.ts
```diff
@@ -674,6 +674,12 @@ export class OneNoteImporter extends FormatImporter {
 		const videos: HTMLIFrameElement[] = pageElement.findAll('iframe') as HTMLIFrameElement[];
 
 		for (const object of objects) {
+			// Objects may contain child nodes which would be lost when the object is replaced by markdown.
+			// To preserve these, move any child items to be siblings of the object
+			while (object.firstChild) {
+				object.parentNode?.insertBefore(object.firstChild, object.nextSibling);
```
tgrosinger, line 680: Could `object.nextSibling` be null? Maybe this should be `object.after(object.firstChild)` instead? ([after](https://developer.mozilla.org/en-US/docs/Web/API/Element/after))

tgrosinger (commented):
Great spotting. Thank you for working on this! Just left a quick question.

src/formats/onenote.ts
scooby359: Yes, it may be null, in which case it gets added to the end of the list

(https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore#referencenode)

tgrosinger: Ah perfect, thank you.

tgrosinger (approved):

PR #382, opened 2025-05-08
Improve Bear imports: added front matter with date and tags, preserving timestamps for assets

roelven:
This pull request enhances the Bear (.bear2bk) importer, adding new features to improve the import experience and data fidelity.

1.  **Accurate Note Timestamp Preservation:**
    *   Original creation and modification dates for notes are now extracted from Bear's `info.json`.
    *   These timestamps are added to the frontmatter of imported notes as `created:` and `modified:` fields (format: `YYYY-MM-DDTHH:mm:ss`).
    *   The file system timestamps for the note files are also updated to match these original dates.
    *   Front matter now will look like this:
        ```
            ---
            created: 2023-02-21T13:36:29
            modified: 2024-02-29T14:25:51
            tags:
            - tag1
            - tag2
            ---
        ```

2.  **Comprehensive Tag Extraction and Processing:**
    *   Tags are now extracted directly from the Markdown body of each Bear note.
    *   **Supported Bear Tag Formats:**
        *   `#enclosed/tag#` (allows spaces and slashes; content within `#...#` must be single-line).
        *   `#simpletag` or `#tag/with/hierarchy` (allows slashes and hyphens not at start/end; no spaces within the tag itself).
    *   **URL Safety:** Tag extraction logic includes safeguards (lookaheads/lookbehinds) to prevent misinterpreting URL fragments (e.g., `page.html#section`) as tags.
    *   **Tag Flattening:** Nested tags (e.g., `#personal/project#` or `#work/projectA`) are automatically flattened into a list of individual tags (e.g., `personal`, `project` and `work`, `projectA`) in the frontmatter.
    *   **Frontmatter Integration:** Unique, flattened tags are added as a standard YAML list under the `tags:` key in each note's frontmatter.

3.  **Attachment Timestamp Preservation:**
    *   The importer now attempts to preserve the original creation (`ctime`) and modification (`mtime`) timestamps for attachments.
    *   Timestamps are extracted from the zip entry for each asset and passed as `DataWriteOptions` when creating the binary file in Obsidian. (Note: The age of these timestamps may ultimately be limited by the metadata stored in the Bear `.bear2bk` zip archive itself).

4.  **Handling of Archived and Trashed Notes:**
    *   Notes marked as "archived" or "trashed" in Bear are now correctly placed into `archive/` and `trash/` subfolders respectively, within the main import output directory.

### DIFF
--- a/src/formats/bear-bear2bk.ts
+++ b/src/formats/bear-bear2bk.ts
@@ -19,6 +19,43 @@ export class Bear2bkImporter extends FormatImporter {
 		this.addOutputLocationSetting('Bear');
 	}
 
+	private extractTagsFromContent(content: string): string[] {
+		const tags = new Set<string>();
+		console.log("[Bear Importer Tag Debug] Original content for tag extraction:\n", content.substring(0, 300)); // Log start of content
+
+		// 1. Extract #enclosed/tags# (can contain spaces and slashes, but content should be single-line)
+		const enclosedTagRegex = /#([^\n#]+?)#/g; // Changed [^#] to [^\n#] to disallow newlines in tag content
+		let matchEnclosed;
+		console.log("[Bear Importer Tag Debug] Trying enclosedTagRegex");
+		while ((matchEnclosed = enclosedTagRegex.exec(content)) !== null) {
+			console.log("[Bear Importer Tag Debug] enclosedTagRegex match found: raw=", matchEnclosed[0], "group1=", matchEnclosed[1]);
+			const tag = matchEnclosed[1].trim();
+			if (tag) {
+				tags.add(tag);
+				console.log("[Bear Importer Tag Debug] Added enclosed tag:", tag);
+			}
+		}
+
+		// 2. Extract simple #tags (alphanumeric, underscore, hyphen, and slash, no spaces)
+		//    Ensures it's not part of a URL or an already processed enclosed tag.
+		//    Allows / in the middle of the tag, but not at the start or end of the simple tag.
+		const simpleTagRegex = /(?<!\S)#([a-zA-Z0-9_][a-zA-Z0-9_/\-]*[a-zA-Z0-9_]|[a-zA-Z0-9_]+)(?![#\w/])/g;
+		let matchSimple;
+		console.log("[Bear Importer Tag Debug] Trying simpleTagRegex");
+		while ((matchSimple = simpleTagRegex.exec(content)) !== null) {
+			console.log("[Bear Importer Tag Debug] simpleTagRegex match found: raw=", matchSimple[0], "group1=", matchSimple[1]);
+			const simpleTag = matchSimple[1].trim(); 
+			if (simpleTag) {
+				tags.add(simpleTag);
+				console.log("[Bear Importer Tag Debug] Added simple tag:", simpleTag);
+			}
+		}
+
+		const finalTags = Array.from(tags);
+		console.log("[Bear Importer Tag Debug] Final extracted tags:", finalTags);
+		return finalTags;
+	}
+
 	async import(ctx: ImportContext): Promise<void> {
 		let { files } = this;
 		if (files.length === 0) {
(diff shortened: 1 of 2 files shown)


tgrosinger:
Thanks for submitting this. Looks like a bunch of great improvements and I'll get started reviewing it.

Can you please clarify why you're flattening tag hierarchys? Obsidian does support tag hierarchy using forward slash, so I would guess it makes sense to import them unchanged.

src/formats/bear-bear2bk.ts
```diff
@@ -19,6 +19,47 @@ export class Bear2bkImporter extends FormatImporter {
...
+	private extractTagsFromContent(content: string): string[] {
+		const tags = new Set<string>();
+
+		// 1. Extract #enclosed/tags# (can contain spaces and slashes, but content should be single-line)
+		const enclosedTagRegex = /#([^\n#]+?)#/g; // Changed [^#] to [^\n#] to disallow newlines in tag content
+		let matchEnclosed;
+		while ((matchEnclosed = enclosedTagRegex.exec(content)) !== null) {
+			const rawEnclosedTag = matchEnclosed[1].trim();
+			if (rawEnclosedTag) {
+				const parts = rawEnclosedTag.split('/');
```
tgrosinger, line 31: As I mentioned in a comment, I don't think we need to be flattening the tag hierarchy. We can import them as written.

src/formats/bear-bear2bk.ts
```diff
@@ -111,12 +174,46 @@ export class Bear2bkImporter extends FormatImporter {
...
+			const frontmatterString = Object.entries(frontmatter)
+				.map(([key, value]) => {
+					if (Array.isArray(value)) { // Handle tags array
+						return `${key}:\n  - ${value.join('\n  - ')}`;
+					}
+					return `${key}: ${value}`;
+				})
+				.join('\n');
+			
+			contentWithFrontmatter = `---\n${frontmatterString}\n---\n\n${content}`;
```
tgrosinger, line 203: Does Bear use yaml frontmatter at all? Will this conflict with preexisting data?

tgrosinger (commented):
I haven't tested yet, but looks really good.

src/formats/bear-bear2bk.ts
roelven: I went for a flattened tag hierarchy as Obsidian's tag management offers better flexibility and scales well long term. Bear has inconsistent formatting for nested tags, e.g. `#work/operations#` and `#work/operations`. Also, Bear supports infinite nested tags, which is setting users up for problems at a later stage: https://bear.app/faq/nested-tags/ 

I know this is heavily opinionated, but when optimizing for larger Bear archives I think this will result in less user headaches.

src/formats/bear-bear2bk.ts
roelven: It's very, very unlikely that there will be a frontmatter block in a Bear note. The note metadata is stored in Bear's database and after export it's in the info.json file. We could make this more robust by adding a check here, agreed.

roelven:
I've added a check to see if there's already frontmatter present in the Bear note. If there is, it won't touch it. I tested this with frontmatter present in Bear notes and this doesn't create any issues now. Let me know if there's anything else you'd like to see here.

src/formats/bear-bear2bk.ts
tgrosinger: Hey @roelven, sorry for the delay. I'm still just hesitant on this PR because of this decision. I would be much more comfortable merging this if we are leaving the tags unchanged.

Obsidian supports arbitrarily long tags nesting as well, so I don't think a tag that works in Bear would be problematic in Obsidian.

olivierguerriat: I agree with @tgrosinger. If we didn't want the hierarchy, we wouldn't use it in Bear either. Flattening it looks like it should some kind of maintenance command after the import for those interested.

tgrosinger:
Thank you @roelven for the work you did on this so far. It's greatly appreciated!
I'm going to close this PR since your changes are being continued in #419.

Issue #359, opened 2025-03-02
[OneNote] Importer ends in half-finished state


altano:
I just took the OneNote import improvements for a test drive hoping that I could now import my notebook[^1] but I got a weird result: the import ended with 82 items remaining, and had 6 errors due to "maximum retry attempts":

<img width="588" alt="Image" src="https://github.com/user-attachments/assets/b3be9901-fc07-42cd-9e5e-c55e770a033b" />

It looks like the limit is 5 and that can be exceeded pretty easily:
https://github.com/obsidianmd/obsidian-importer/blob/b14b3ae79a278f0402f7bc5bd5f0b5152406aeb5/src/formats/onenote.ts#L16

and I guess 82 items just got lost somehow?

I'd be happy to help debug what's going wrong here if someone could give me a pointer where to look.

[^1]: which was previously hitting lots of rate-limiting errors

anelshaer:
@altano Also facing the same rate limit and a need to keep rerunning the importer every 30min to continue where it stopped. 

it would be better to make the process continue every predefined time like 30mins without the need to rerun the importer process again and maybe in background .

altano:
The OneNote api tells the importer to back off and the importer has code to back off, so there probably just shouldn’t be a max number of retries. Or it should be really high. Then retrying just wouldn’t be needed, which would be better.

The bug that’s causing some items to get stuck in “remaining”) without actually being in a queue is a separate issue. Looks like there’s still some flakiness here.

altano:
I debugged this a bit today and noticed a few issues with the onenote importing code:

Low severity:
- The retry backoff is per-request, so if the onenote API throttles us, every in-flight request keeps retrying. There should be one place the backoff happens, where ALL requests are paused while we're throttled.
- This Retry-After logic (`(+!response.headers.get('Retry-After') * 1000) || 15000;`) isn't right: if the header is not in the response you always get back 1000, not 15000: https://github.com/obsidianmd/obsidian-importer/blob/b14b3ae79a278f0402f7bc5bd5f0b5152406aeb5/src/formats/onenote.ts#L890

High severity:
- `MAX_RETRY_ATTEMPTS` is being trivially hit because 429 errors are contributing to failed requests. Either `MAX_RETRY_ATTEMPTS` should be removed or 429 backoff errors should not contribute to the max limit being hit.
[...]

PR #388, opened 2025-06-01
fix onenote importer bugs

altano:
Fixes issues in https://github.com/obsidianmd/obsidian-importer/issues/359

- Improve typing by creating a `json-wrapped` that captures the case where the return value is in the `.value` property of the response
- allow `fetchRequest` callers to specify return type
- handle http status = 429 errors the same as response.code === 20166 errors
- backoff errors (429/20166) do NOT increase the retry count, we might theoretically retry these forever
- never return undefined responses from `fetchRequest`. it succeeds or throws an error when the retry limit is hit.
- make it so that when the importer dialog is closed, we abort all network requests and stop retrying. This is particularly important now that we might infinitely retry on 429 errors.

### DIFF
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -12,7 +12,7 @@ const GRAPH_SCOPES: string[] = ['user.read', 'notes.read'];
 const SELF_CLOSING_REGEX = /<(object|iframe)([^>]*)\/>/g;
 // Regex for fixing whitespace and paragraphs
 const PARAGRAPH_REGEX = /(<\/p>)\s*(<p[^>]*>)|\n  \n/g;
-// Maximum amount of request retries, before they're marked as failed
+// Maximum amount of request retries, before they're marked as failed. Does not include 429 backoff errors.
 const MAX_RETRY_ATTEMPTS = 5;
 
 const BASE64_REGEX = new RegExp(/^data:[\w\d]+\/[\w\d]+;base64,/);
@@ -964,7 +964,14 @@ export class OneNoteImporter extends FormatImporter {
 
 					if (retryCount < MAX_RETRY_ATTEMPTS) {
 						await new Promise(resolve => setTimeout(resolve, retryTime));
-						return this.fetchResource(url, returnType as any, retryCount + 1);
+						return this.fetchResource(
+							url, 
+							returnType as any, 
+							// don't increment the retryCount because we were told
+							// to backoff, and we should infinitely retry on backoff
+							// errors.
+							retryCount
+						);
 					}
 					else throw new Error('Exceeded maximum retry attempts');
 				}
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -959,7 +959,8 @@ export class OneNoteImporter extends FormatImporter {
 
 				// We're rate-limited - let's retry after the suggested amount of time
 				if (err?.code === '20166') {
-					let retryTime = (+!response.headers.get('Retry-After') * 1000) || 15000;
+					const retryAfter = response.headers.get('Retry-After');
+					let retryTime = retryAfter ? (+retryAfter * 1_000) : 15_000;
 					console.log(`Rate limit exceeded, waiting for: ${retryTime} ms`);
 
 					if (retryCount < MAX_RETRY_ATTEMPTS) {
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -473,7 +473,7 @@ export class OneNoteImporter extends FormatImporter {
 					consecutiveFailureCount++;
 					progress.reportFailed(page.title, e.toString());
 
-					if (consecutiveFailureCount > 5) {
+					if (consecutiveFailureCount > 5 || this.modal.abortController.signal.aborted) {
 						// Likely being rate limited.
 						progress.status('Microsoft OneNote has limited how fast notes can be imported. Please try again in 30 minutes to continue importing.');
 						return;
@@ -927,9 +927,20 @@ export class OneNoteImporter extends FormatImporter {
 			throw new Error('Exceeded maximum retry attempts');
 		}
 
+		if (this.modal.abortController.signal.aborted) {
+			const abortReason = this.modal.abortController.signal.reason ?? 'no reason given';
+			throw new Error(`The import was aborted (${abortReason})`);
+		}
+
 		try {
 			// any errors that happen in the try block will be retried.
-			let response = await fetch(url, { headers: { Authorization: `Bearer ${this.graphData.accessToken}` } });
+ 			let response = await fetch(
+ 				url, 
+ 				{
+ 					headers: { Authorization: `Bearer ${this.graphData.accessToken}` },
+ 					signal: this.modal.abortController.signal,
+ 				}
+ 			);
 
 			if (response.ok) {
 				switch (returnType) {
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -960,7 +960,14 @@ export class OneNoteImporter extends FormatImporter {
 				// We're rate-limited - let's retry after the suggested amount of time
 				if (err?.code === '20166') {
 					const retryAfter = response.headers.get('Retry-After');
-					let retryTime = retryAfter ? (+retryAfter * 1_000) : 15_000;
+					// If we're rate-limited, the soonest we'll be able to make
+					// the request again is the next minute, so wait either as
+					// long as the API tells us to (with the Retry-After header)
+					// or wait 1 minute. Note: the OneNote API does not
+					// currently ever provide the Retry-After header. See
+					// https://learn.microsoft.com/en-us/graph/throttling-limits#onenote-service-limits
+					// for more info.
+					let retryTime = retryAfter ? (+retryAfter * 1_000) : 60_000;
 					console.log(`Rate limit exceeded, waiting for: ${retryTime} ms`);
 
 					if (retryCount < MAX_RETRY_ATTEMPTS) {
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -424,7 +424,7 @@ export class OneNoteImporter extends FormatImporter {
 
 			let pages: OnenotePage[] | null = null;
 			try {
-				pages = ((await this.fetchResource<OnenotePage>(pagesUrl, 'json-wrapped')).value);
+				pages = ((await this.fetchResource<OnenotePage>(pagesUrl, 'json-wrapped', progress)).value);
 			}
 			catch (e) {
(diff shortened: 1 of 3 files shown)


altano:
These changes are NOT tested. I don't know how to test my edits in Obsidian. Help testing is welcome.

tgrosinger:
Hi @altano, thanks for contributing this. Your changes are pretty comprehensive and intermingled, which makes determining changes in behavior very difficult.

It would be much more likely that we can review and merge these improvements if you could split the changes into distinct commits with a single purpose. Ideally each commit should be a functional end state so they can be reviewed, tested, and merged independently.

Thank you!

altano:
Hey @tgrosinger, I’d be happy to oblige but I’m not sure how.

You can’t stack PRs that you’re submitting from another repository in GitHub, and my changes would definitely need to be stacked.

Or are you saying one PR with multiple commits is fine, as long as the commits are _theoretically_ independently mergeable?

tgrosinger:
You can definitely just keep them in this one PR. Just separate commits. I just meant that ideally each commit leaves the plugin in a working state.

altano:
@tgrosinger I split up the commits, ~~but I still need to test these changes. How do you test your changes to the importer?~~

I figured out how to test my local plugin:
1. Disable the published importer in the Obsidian plugin settings
2. Symlink the github repo into the vault's plugin dir, e.g. `sudo ln -s ~/src/obsidian-importer/ '/Users/alan/Library/Mobile Documents/iCloud~md~obsidian/Documents/Personal/.obsidian/plugins/obsidian-importer'`
3. Optional: in the github repo's manifest.json, rename the plugin to something to make sure we're using the local-development version, e.g. `"name": "Importer (Local Dev)",`
4.  Enable the local development importer in the Obsidian plugin settings (might have to click "refresh")
5. Optional: add the hot-reload plugin and add a .hotreload file to the github repo

I'll test and report back.

altano:
I was able to make a full, error-free clone of a large notebook of mine with these changes:
image

It took 62 minutes with the current logic, which includes some more improvements (e.g. being more lenient with dumb graph api errors).

I'm going to (1) clean things up and (2) test against my entire note collection. Will report back soon.

altano:
Okay, the code is ready for review, @tgrosinger. The OneNote importer is now resilient enough to import all my notebooks:
image

**Import stats**:
Imported: 1,346 notes / 1,793 attachments
3h44m total time spent importing

**Requests over time, in 15m buckets**:
image

You can see the OneNote service limit having a rough cycle of 1 hour (there's a burst of http 200 requests every hour, until it gets throttled again)

src/formats/onenote.ts
```diff
@@ -17,6 +17,42 @@ const MAX_RETRY_ATTEMPTS = 5;
...
+
+function assertJSONWrappedResponse<T>(res: unknown): asserts res is JSONWrappedResponse<T> {
+	if (res == null) {
+		throw new Error(`response is nullish`);
+	}
+	if (typeof res !== 'object') {
+		throw new Error(`response is not an object type`);
+	}
+
+	if ('@odata.nextLink' in res) {
```
tgrosinger, line 39: Should this be marked as optional in the type?

src/formats/onenote.ts
```diff
@@ -964,7 +964,14 @@ export class OneNoteImporter extends FormatImporter {
...
 					if (retryCount < MAX_RETRY_ATTEMPTS) {
 						await new Promise(resolve => setTimeout(resolve, retryTime));
-						return this.fetchResource(url, returnType as any, retryCount + 1);
+						return this.fetchResource(
+							url, 
+							returnType as any, 
+							// don't increment the retryCount because we were told
+							// to backoff, and we should infinitely retry on backoff
+							// errors.
+							retryCount
```
tgrosinger: I'm not sure about this. If we go with this change I'm wondering if we should change the `console.log` above to instead report to the `progress` object. Otherwise this could make the UI look like it's just hanging.

src/formats/onenote.ts
```diff
@@ -72,6 +72,7 @@ export class OneNoteImporter extends FormatImporter {
 	attachmentDownloadPauseCounter = 0;
 	rememberMe = false;
 	refreshToken?: string;
+	abortController: AbortController = new AbortController();
```
tgrosinger: I'm wondering if it would be reasonable to put the `abortController` on the `ImporterModal` directly. Then we could simplify this a little bit and reduce the duplicate code that would be introduced by adopting this method in the other importers.

The `ImporterModal` is accessible from `fetchResource`, e.g. `this.modal.abortController`.

src/formats/onenote.ts
```diff
@@ -483,10 +482,25 @@ export class OneNoteImporter extends FormatImporter {
...
+						// Report progress as complete
+						progress.reportProgress(progressTotal, progressTotal);
+
 						return;
 					}
 				}
-				progress.reportProgress(progressCurrent, progressTotal);
+
+				// report progress even if page import fails or is skipped
+				progress.reportProgress(++progressCurrent, progressTotal);
```
tgrosinger, line 502: Is the `reportProgress` at the beginning of the outer for-loop still required?

tgrosinger (commented):
This is looking really good. Thanks for all your work on these improvements!

src/formats/onenote.ts
tgrosinger: Hmm, actually it seems like we probably won't have easy access to `progress` here because this is used before `import` is called.

What about leaving in the max-retries, but raising the limit a bit, and adding a very basic linear backoff. For example something like `Math.min(retryCount, 5) * 60 * 1000`

I think it's okay if the import fails due to running out of retries because we have the option to skip importing of previously imported notes. Just needs to catch most of the cases.

src/formats/onenote.ts
altano: I *think* it's either there and a string, or not there, which is what the type is. I don't think it can be in the object but have a value of `null`.

src/formats/onenote.ts
altano: > but raising the limit a bit, and adding a very basic linear backoff. For example something like Math.min(retryCount, 5) * 60 * 1000

The logic here could probably be better but, because the API has a 1 hour granularity for throttling, there isn't much point in retrying until the next hour. Normal backoff strategies (e.g. exponential or linear backoff) therefore don't make a lot of sense here. I can't think of a reason to make the retry duration contingent on the retry count.. If I were to rewrite the retry time limit logic I would probably increase it from 60s to 15 minutes or something. I'm open to brainstorming a formula that makes more sense.

> I think it's okay if the import fails due to running out of retries because we have the option to skip importing of previously imported notes. Just needs to catch most of the cases.

It's not my call and I'll let you overrule me, but I think it's a MUCH better experience to have the import eventually succeed than to have it keep failing and have the user keep running the import with "skip previously imported" turned on. The "remaining" count is a bit buggy atm (I've hopefully fixed that but I'm not sure) so it's hard to tell when it's stopped importing. And it's also hard to tell when you've imported everything successfully and can just stop (it's when an import imports 0 notes I guess?).

If we want to make sure that we give good feedback about the fact that the import is stalled on a 429 backoff, I'd be happy to implement that, but I think that would be the better solution.

src/formats/onenote.ts
altano: Good suggestion. I've implemented that and it looks good. Will be in next push.

src/formats/onenote.ts
altano: Uhhh, good question. Probably not? I can't think of a case where it does something. I'll remove it.

src/formats/onenote.ts
altano: To make it easier, we could not worry about throttling _before_ the import begins, which is possible but less likely, and only update the progress to tell the user about throttling _during_ the import. I can implement that if you think it's good enough?

altano: I went ahead and implemented that ^^^:
Untitled-cropped-split
(I cut the middle of the gif out and sped it up so you're not staring at the pause screen for 60s)

Let me know if you think it's good enough or what. Feedback welcome.

tgrosinger: Nice, I like the feedback showing that it's pausing due to the OneNote limitation.

> because the API has a 1 hour granularity for throttling, there isn't much point in retrying until the next hour

That's a very interesting point. Do we have a way of knowing for sure that has happened? Maybe this is a stretch, but what about:

- Tell the user that we've hit the hourly api limit.
- Start retrying once every 5-10 minutes.
- If we can't make progress for >90 minutes then abort?

altano: > That's a very interesting point. Do we have a way of knowing for sure that has happened?

Unfortunately no =/. And the graph sdk is bad at handling the OneNote case too, as it defaults to exponential backoff for 3 retries, which obviously doesn't make sense (see [this PR](https://github.com/microsoftgraph/msgraph-sdk-javascript/pull/1514) that's been open for 2 years), and can only be configured to go up to 10 retries/17 minutes, which [is obviously insufficient](https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/978#issuecomment-1780358315).

I think we should just be semi-smart and not retry too frequently but also not rely on any specific behavior of the API. Something in the 1 to 15 minute time range feels okay to me.

> Maybe this is a stretch, but what about:
> 
>     Tell the user that we've hit the hourly api limit.

So what I have above works for that feedback, right? Feel free to wordsmith.

>     Start retrying once every 5-10 minutes.

Yeah okay, I can def increase the 1m timer to 5-10 minutes. Any preference for how long?

>     If we can't make progress for >90 minutes then abort?

I only have one small hesitation with doing this, and it's why I didn't include it in the PR:

The http requests are SORT of serial now, so such logic would be totally fine, because there is only one out-going request at a time and it'll eventually unblock and succeed once the API lets us through again. But, if anyone adds concurrency in the future, e.g. 5 requests in parallel, they'll execute randomly and any single request can become starved for >90 minutes while we're still making progress.

Should I just ignore that hypothetical future and add a 90m hard timeout? That seems fine to me.

altano: Oh, I could actually code the 90-minute timeout to not be request specific, making that concern not even valid. We'll start counting when any 429 is hit, and clear it when any resource is successfully fetched? Yeah I can do that.

altano: Okay, just pushed a new version that does the above:

- keep retrying with no error limit
- if we hit a 429, wait 1m (can increase, let me know)
- if we don't fetch any resources successfully for >90m, give up with an error

I also made a few small changes to improve the progress reporting and clean things up.

I'm happy to tweak things more, just let me know what you think.

tgrosinger: This looks great. I did some additional testing and it seems to be working as expected. Thank you!

PR #394, opened 2025-06-09
Add option to skip duplicate notes in Apple Notes importer

jykim:
Introduced a new setting to allow users to skip importing notes that already exist in the vault. This feature helps prevent duplication during the import process.

### DIFF
--- a/src/formats/apple-notes.ts
+++ b/src/formats/apple-notes.ts
@@ -35,6 +35,7 @@ export class AppleNotesImporter extends FormatImporter {
 	omitFirstLine = true;
 	importTrashed = false;
 	includeHandwriting = false;
+	skipDuplicates = true;
 	trashFolders: number[] = [];
 
 	init(): void {
@@ -80,6 +81,16 @@ export class AppleNotesImporter extends FormatImporter {
 			.addToggle(t => t
 				.setValue(false)
 				.onChange(async v => this.includeHandwriting = v)
+			);
+
+		new Setting(this.modal.contentEl)
+			.setName('Skip duplicate notes')
+			.setDesc(
+				'Skip importing notes that already exist in the vault.'
+			)
+			.addToggle(t => t
+				.setValue(true)
+				.onChange(async v => this.skipDuplicates = v)
 			);
 	}
 
@@ -253,6 +264,12 @@ export class AppleNotesImporter extends FormatImporter {
 
 		const title = `${row.ZTITLE1}.md`;
 		const file = await this.saveAsMarkdownFile(folder, title, '');
+
+		// Check for duplicate notes if the option is enabled
+		if (this.skipDuplicates && this.vault.getAbstractFileByPath(file.path)) {
+			this.ctx.reportSkipped(row.ZTITLE1, 'note is a duplicate');
+			return null;
+		}
 
 		this.ctx.status(`Importing note ${title}`);
 		this.resolvedFiles[id] = file;


tsconfig.json
```diff
@@ -17,7 +17,8 @@
 			"DOM",
 			"ES5",
 			"ES6",
-			"ES7"
+			"ES7",
+			"ES2019"
```
tgrosinger: Please don't update this file for now. Let's find a way to accomplish the changes at the current API level please.

tgrosinger:
Thanks for making these improvements. I see that the PR is marked as a work in progress still, so let me know when it's ready for review.

jykim:
You can test the pre-release version using [BRAT](https://tfthacker.com/BRAT).

https://github.com/jykim/obsidian-importer/releases/tag/v0.2

jykim:
@tgrosinger -- I think it's ready for your review!

jykim:
Just a gentle nudge @tgrosinger :)

src/formats/apple-notes.ts
```diff
@@ -50,6 +58,18 @@ export class AppleNotesImporter extends FormatImporter {
...
 		this.addOutputLocationSetting('Apple Notes');
 
+		new Setting(this.modal.contentEl)
+			.setName('File prefix format')
+			.setDesc(
+				'Format for the date prefix in filenames. Use YYYY, MM, DD for year, month, day.' +
+				' Leave blank for no prefix.'
+			)
+			.addText(t => t
+				.setValue('YYYY-MM-DD')
```
tgrosinger: Let's please leave this empty so the default is to preserve the existing behavior.

src/formats/apple-notes.ts
```diff
@@ -251,8 +284,42 @@ export class AppleNotesImporter extends FormatImporter {
...
 		const folder = this.resolvedFolders[row.ZFOLDER] || this.rootFolder;
 
-		const title = `${row.ZTITLE1}.md`;
-		const file = await this.saveAsMarkdownFile(folder, title, '');
+		// Get creation date and format it according to user preference
+		let title = row.ZTITLE1;
+		if (this.filePrefixFormat) {
+			const creationDate = new Date(this.decodeTime(row.ZCREATIONDATE3 || row.ZCREATIONDATE2 || row.ZCREATIONDATE1));
+			const datePrefix = this.filePrefixFormat
+				.replace('YYYY', creationDate.getUTCFullYear().toString())
```
tgrosinger: This is pretty rigid right now. Let's please use Moment to output the date using the provided format string. See the other importers to see how this is being done.

src/formats/apple-notes.ts
```diff
@@ -251,8 +284,42 @@ export class AppleNotesImporter extends FormatImporter {
...
+				.replace('MM', (creationDate.getUTCMonth() + 1).toString().padStart(2, '0'))
+				.replace('DD', creationDate.getUTCDate().toString().padStart(2, '0'));
+			title = `${datePrefix} ${title}`;
+		}
+
+		const fullPath = `${folder.path}/${title}.md`;
```
tgrosinger: `path.join`. See other uses in this file.

src/formats/apple-notes.ts
```diff
@@ -251,8 +284,42 @@ export class AppleNotesImporter extends FormatImporter {
...
+
+		// Check for duplicate notes based on the selected handling option
+		const existingFile = this.vault.getAbstractFileByPath(fullPath) as TFile;
+		
+		if (existingFile) {
```
tgrosinger: Instead of asserting above, please do `if (existingFile && existingFile instanceof TFile) {`

src/formats/apple-notes.ts
```diff
@@ -352,9 +420,42 @@ export class AppleNotesImporter extends FormatImporter {
 				break;
 		}
 
+		// Apply date prefix to attachment name if configured
+		let finalAttachmentName = outName;
+		if (this.filePrefixFormat && row.ZCREATIONDATE) {
+			const creationDate = new Date(this.decodeTime(row.ZCREATIONDATE));
+			const datePrefix = this.filePrefixFormat
+				.replace('YYYY', creationDate.getUTCFullYear().toString())
```
tgrosinger: Same comment as above.

src/formats/apple-notes.ts
```diff
@@ -352,9 +420,42 @@ export class AppleNotesImporter extends FormatImporter {
...
+				.replace('MM', (creationDate.getUTCMonth() + 1).toString().padStart(2, '0'))
+				.replace('DD', creationDate.getUTCDate().toString().padStart(2, '0'));
+			finalAttachmentName = `${datePrefix} ${outName}`;
+		}
+
+		// Check for existing attachment based on the selected handling option
+		const attachmentPath = await this.getAvailablePathForAttachment(`${finalAttachmentName}.${outExt}`, []);
+		const existingAttachment = this.vault.getAbstractFileByPath(attachmentPath) as TFile;
```
tgrosinger: Same comment about assertions as above.

src/formats/apple-notes.ts
```diff
@@ -352,9 +420,42 @@ export class AppleNotesImporter extends FormatImporter {
...
+		
+		if (existingAttachment) {
+			if (this.duplicateHandling === DuplicateHandling.Skip) {
+				this.ctx.reportSkipped(finalAttachmentName, 'attachment already exists');
+				return existingAttachment;
```
tgrosinger, line 448: Should this be returning null?

src/formats/apple-notes.ts
```diff
@@ -352,9 +420,42 @@ export class AppleNotesImporter extends FormatImporter {
...
+			} else if (this.duplicateHandling === DuplicateHandling.ImportUpdated) {
+				// Check modification times for attachments
+				const appleAttachmentModTime = this.decodeTime(row.ZMODIFICATIONDATE);
+				const existingAttachmentModTime = existingAttachment.stat.mtime;
+				
+				if (appleAttachmentModTime <= existingAttachmentModTime) {
+					this.ctx.reportSkipped(finalAttachmentName, 'attachment unchanged since last import');
+					return existingAttachment;
```
tgrosinger, line 457: Should this be returning null?

src/formats/apple-notes.ts
```diff
@@ -391,4 +492,30 @@ export class AppleNotesImporter extends FormatImporter {
 			return await fsPromises.readFile(path.join(os.homedir(), NOTE_FOLDER_PATH, sourcePath));
 		}
 	}
+
+	async getAvailablePathForAttachment(filename: string, existingPaths: string[]): Promise<string> {
+		// Always use the default behavior from FormatImporter to respect Obsidian's attachment folder settings
+		// This ensures attachments go to the configured attachment folder (e.g., _files_) rather than the note folder
+		return super.getAvailablePathForAttachment(filename, existingPaths);
```
tgrosinger: If there are no behavior changes please don't override this method.

src/formats/apple-notes.ts
```diff
@@ -391,4 +492,30 @@ export class AppleNotesImporter extends FormatImporter {
...
+	}
+
+	async saveAsMarkdownFile(folder: TFolder, title: string, content: string): Promise<TFile> {
+		if (this.duplicateHandling === DuplicateHandling.Skip || this.duplicateHandling === DuplicateHandling.ImportUpdated) {
+			// For Skip and ImportUpdated, create the file directly without numeric suffix
+			const sanitizedName = sanitizeFileName(title);
+			const fullPath = `${folder.path}/${sanitizedName}`;
```
tgrosinger: path.join

src/formats/apple-notes.ts
```diff
@@ -391,4 +492,30 @@ export class AppleNotesImporter extends FormatImporter {
...
+			
+			// Check if file already exists and handle overwriting
+			const existingFile = this.vault.getAbstractFileByPath(fullPath) as TFile;
+			if (existingFile) {
+				// File exists, this means we're updating it (timestamp comparison passed)
+				return existingFile;
```
tgrosinger, line 514: Where is the content updated?

src/formats/apple-notes.ts
```diff
@@ -391,4 +492,30 @@ export class AppleNotesImporter extends FormatImporter {
...
+			} else {
+				// Create new file
+				return await this.vault.create(fullPath, content);
```
tgrosinger: If the file doesn't exist then just fall through to call the super function.

tgrosinger (changes requested):
This looks like a great start. Thank you working on this. Let me know if you have questions about any of these comments.

jykim:
Thanks for comments @tgrosinger .
I'm out of town so will work on this next week!

src/formats/apple-notes.ts
jykim: done

src/formats/apple-notes.ts
jykim: Content is updated later in resolveNote

src/formats/apple-notes.ts
jykim: removed

jykim:
All your comments addressed. @tgrosinger -- thanks for thoughtful comments. 

As for default file prefix, I still prefer to keep 'YYYY-MM-DD' as default value, although I understand this might cause disruption for existing user. Since the use case I have in mind is running the import regularly to keep the contents in sync, setting the value every time is a bit of hassle. Can we move this into plug-in setting so that it won't have to be set for every import?

src/formats/apple-notes.ts
```diff
@@ -81,6 +101,19 @@ export class AppleNotesImporter extends FormatImporter {
...
+
+		new Setting(this.modal.contentEl)
+			.setName('Handle duplicate files')
+			.setDesc(
+				'How to handle notes that already exist in the vault.'
+			)
+			.addDropdown(d => d
+				.addOption(DuplicateHandling.Skip, 'Skip import')
+				.addOption(DuplicateHandling.ImportUpdated, 'Import only updated')
+				.addOption(DuplicateHandling.CreateCopy, 'Create a copy (legacy)')
```
tgrosinger:
```suggestion
				.addOption(DuplicateHandling.CreateCopy, 'Create a copy')
```

I don't think we need to include `(legacy)`. Most people using this plugin are first time users of it, so they don't need to know what the previous default behavior was.

src/formats/apple-notes.ts
tgrosinger: Looks like this comment was missed. Please don't assert here and instead check the type.
If you remove the assertion here, then below where it's checked you can add the type check too.

```
if (existingAttachment && existingAttachment instanceof TFile) {
```

src/formats/apple-notes.ts
tgrosinger: This comment was missed.

src/formats/apple-notes.ts
```diff
@@ -251,8 +284,39 @@ export class AppleNotesImporter extends FormatImporter {
...
+		if (this.filePrefixFormat) {
+			const creationTimestamp = this.decodeTime(row.ZCREATIONDATE3 || row.ZCREATIONDATE2 || row.ZCREATIONDATE1);
+			const datePrefix = moment(creationTimestamp).format(this.filePrefixFormat);
+			title = `${datePrefix} ${title}`;
+		}
+
+		const fullPath = path.join(folder.path, `${title}.md`);
+
+		// Check for duplicate notes based on the selected handling option
+		const existingFile = this.vault.getAbstractFileByPath(fullPath) as TFile;
```
tgrosinger:
```suggestion
		const existingFile = this.vault.getAbstractFileByPath(fullPath);
```

src/formats/apple-notes.ts
```diff
@@ -391,4 +486,21 @@ export class AppleNotesImporter extends FormatImporter {
...
 	}
+
+	async saveAsMarkdownFile(folder: TFolder, title: string, content: string): Promise<TFile> {
+		if (this.duplicateHandling === DuplicateHandling.Skip || this.duplicateHandling === DuplicateHandling.ImportUpdated) {
+			// For Skip and ImportUpdated, create the file directly without numeric suffix
+			const sanitizedName = sanitizeFileName(title);
+			const fullPath = path.join(folder.path, sanitizedName);
+			
+			// Check if file already exists and handle overwriting
+			const existingFile = this.vault.getAbstractFileByPath(fullPath) as TFile;
```
tgrosinger: Another assertion. Please change this one too.

src/formats/apple-notes.ts
tgrosinger: Got it, thank you.

src/formats/apple-notes.ts
tgrosinger: Actually it seems like maybe the `return null` below should not be returning null? Because these cases are successful, they just don't need to do anything. So we don't want the caller of this function to think this was unsuccessful.

tgrosinger:
> As for default file prefix, I still prefer to keep 'YYYY-MM-DD' as default value, although I understand this might cause disruption for existing user. Since the use case I have in mind is running the import regularly to keep the contents in sync, setting the value every time is a bit of hassle. Can we move this into plug-in setting so that it won't have to be set for every import?

I don't think we should add it as a plugin setting because there aren't any other plugin settings so far. What about storing the last used value in local storage, and using that if there is one? Take a look at the OneNote importer where we are using `localStorage.getItem` and `localStorage.setItem`.

jykim:
Thanks. I applied your other comments, and I also tried to use incorporate local storage, and the stored value (YYYY-MM-DD in my case) does show up in UI once I entered, but when I run import the default value ("") is used, creating files with no prefix. I'm trying to debug this myself but any comment would be appreciated.

jykim:
I added getter/setter for filePrefixFormat with log to see the behavior, and here's my observation:
1. When I run import without touching the filePrefixFormat value in UI, no prefix is applied to import although the UI shows YYYY-MM-DD as the value
2. When I run import after touching the filePrefixFormat value in UI (e.g. YYYY-MM-DD >> YYYY-MM-D >> YYYY-MM-DD), set prefix value (YYYY-MM-DD) is applied
But the setter for filePrefixFormat indicates that the filePrefixFormat is set to YYYY-MM-DD when the importer UI initially loads, so I don't understand the behavior for case (1).

rgaufman:
Does this also update the md file created/updated time?  - or at least references the created/updated date/time in the note?

tgrosinger:
I pushed a few changes to address some problems I found in testing. Thank you @jykim!

jykim:
Thank you so much for finishing this off!

Issue #363, opened 2025-03-03
OneNote: nested bulletted lists have extra preceding space


altano:
If I have OneNote lists that look like this:

```text
* a
    * b
    * c
        * d
```

It is imported into Obsidian with Markdown like this:

```md
- a

    - b
    - c

        - d
```

There is an extra line break before every indentation level and that is incorrect.

altano:
This is because OneNote wraps the initial list item in a 0-margin paragraph:

```html
<ul>
	<li>
		<p style="margin-top:0pt;margin-bottom:0pt">First Line</p>
		<ul>
			<li style="list-style-type:circle">Nested Bullet</li>
		</ul>
	</li>
</ul>
```

And [turndown](https://mixmark-io.github.io/turndown/) converts that into an extra newline:

<img width="1099" alt="Image" src="https://github.com/user-attachments/assets/c7d1bac4-6481-4ff9-80ba-d87f12d7c0c9" />

altano:
Fixed by #399

PR #399, opened 2025-06-18
extra newlines in bulleted lists

altano:
Addresses https://github.com/obsidianmd/obsidian-importer/issues/363

OneNote wraps the initial list item in a 0-margin paragraph, and [turndown](https://mixmark-io.github.io/turndown/) (Obsidian's html-to-markdown converter) converts that into an extra newline:

image

We can fix this by removing these extra paragraphs that appear as the first child of list items:

BEFORE:
```html
	<ul>
		<li>
			<p style="margin-top:0pt;margin-bottom:0pt">List Item 1</p>
			<ul>
				<li style="list-style-type:circle">List Item 1.a</li>
			</ul>
		</li>
	</ul>
```

AFTER:
```html
	<ul>
		<li>
			List Item 1
			<ul>
				<li style="list-style-type:circle">List Item 1.a</li>
			</ul>
		</li>
	</ul>
```

This results in the correct markdown from turndown (no extra newlines):

| Before | After |
|-|-|
| image | image |


src/formats/onenote.ts
```diff
@@ -935,6 +937,44 @@ export class OneNoteImporter extends FormatImporter {
...
+	//				<li style="list-style-type:circle">List Item 1.a</li>
+	//			</ul>
+	//		</li>
+	//	</ul>
+	//
+	// https://github.com/obsidianmd/obsidian-importer/issues/363
+	removeExtraListItemParagraphs(element: HTMLElement): HTMLElement {
+		// if the first list item child is a paragraph
+		Array.from(element.querySelectorAll('li > p:first-child'))
+			.filter(isHTMLElement)
```
tgrosinger: I understand that this would work, but could this be reorganized into a single for..each loop just so we aren't looping over the list three times?

src/formats/onenote.ts
```diff
@@ -4,6 +4,7 @@ import { genUid, parseHTML } from '../util';
 import { FormatImporter } from '../format-importer';
 import { ATTACHMENT_EXTS, AUTH_REDIRECT_URI, ImportContext } from '../main';
 import { AccessTokenResponse } from './onenote/models';
+import { isHTMLElement } from './isHTMLElement';
```
tgrosinger: I think we should just move this into this file until it's needed elsewhere.

tgrosinger (changes requested):
Couple small comments, but otherwise looks great. Thank you!

tgrosinger:
Thank you!

Issue #396, opened 2025-06-13
Table html not being converted correctly


altano:
Sometimes I have tables in OneNote:

![Image](https://github.com/user-attachments/assets/7db239a0-068a-4584-ac20-10d4b0286389)

That get imported incorrectly into Obsidian:

| | |
|-|-|
| <img width="599" alt="Image" src="https://github.com/user-attachments/assets/a1512a52-382c-47f1-bc4f-8d35129f67e5" /> | <img width="599" alt="Image" src="https://github.com/user-attachments/assets/3c4e88f1-4f69-409c-bc0d-302cf775a98a" /> |

It looks like the html for the entire table is being dumped into one of the cells.

I thought maybe the response from the OneNote API was weird for this table but it looks fine to me. Here's the table html pulled out of the network response:

```html
<table style="border:1px solid;border-collapse:collapse">
	<tr>
		<td style="border:1px solid"><span style="font-weight:bold">CPU</span></td>
		<td style="border:1px solid"><span style="font-weight:bold">TDP</span></td>
		<td style="border:1px solid"><span style="font-weight:bold">Cores</span></td>
		<td style="border:1px solid"><span style="font-weight:bold">CPU Mark</span></td>
		<td style="border:1px solid"><span style="font-family:Cambria Math;font-weight:bold">⍙</span></td>
		<td style="border:1px solid"><p style="margin-top:0pt;margin-bottom:0pt"><span style="font-weight:bold">Single </span></p>
		<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-weight:bold">Thread</span></p>
		</td>
		<td style="border:1px solid"><span style="font-family:Cambria Math;font-weight:bold">⍙</span></td>
		<td style="border:1px solid"><span style="font-family:Cambria Math;font-weight:bold">$</span></td>
	</tr>
	<tr>
		<td style="border:1px solid"><span style="font-weight:bold">EPYC 3251</span></td>
		<td style="border:1px solid;text-align:right"><span style="font-family:Consolas;color:#5b9bd5">50w</span></td>
		<td style="border:1px solid;text-align:right"><span style="font-family:Consolas;color:#5b9bd5">8/16</span></td>
		<td style="border:1px solid;text-align:right"><span style="font-family:Consolas;color:#5b9bd5">13,912</span></td>
		<td style="border:1px solid"><br />
		</td>
[...]

PR #400, opened 2025-06-18
Onenote code import

altano:
Addresses https://github.com/obsidianmd/obsidian-importer/issues/396

Fix several bugs with code in imported OneNote notes:

## Bug 1: don't merge code fences

Currently, all code on the page is merged into the first code fence found on the page. That's clearly a bug.

| OneNote Content | Import - Before | Import - After |
|-|-|-|
| image | image | image |

## Bug 2 and 3: text nodes and newlines

The importer uses `querySelectorAll` which only finds elements and not text nodes, so e.g. if two code blocks have a text node between them, we erroneously merge the two disjoint code blocks and then dump the text node after the code fence.

Also, we're injecting too many newlines in the imported markdown.

| OneNote Content | Import - Before | Import - After |
|-|-|-|
| image | image | image |

## Bug 4: support inline code spans

if an element is code, is a span, and only contains text, use a `<code>` inline span to wrap it instead of a multi-line code fence.

| OneNote Content | Import - Before | Import - After |
|-|-|-|
| image | image | image |

### DIFF
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -4,7 +4,7 @@ import { genUid, parseHTML } from '../util';
 import { FormatImporter } from '../format-importer';
 import { ATTACHMENT_EXTS, AUTH_REDIRECT_URI, ImportContext } from '../main';
 import { AccessTokenResponse } from './onenote/models';
-import { getSiblingsInSameCodeBlock, isFenceCodeBlock, isInlineCodeSpan, isBRElement, isCodeBlockWrappingParagraph } from './onenote/code';
+import { getSiblingsInSameCodeBlock, isFenceCodeBlock, isInlineCodeSpan, isBRElement, isParagraphWrappingOnlyCode } from './onenote/code';
 
 const LOCAL_STORAGE_KEY = 'onenote-importer-refresh-token';
 const GRAPH_CLIENT_ID: string = '66553851-08fa-44f2-8bb1-1436f121a73d';
@@ -550,13 +550,13 @@ export class OneNoteImporter extends FormatImporter {
 			let taggedPage = this.convertTags(parseHTML(splitContent.html));
 			let html = await this.getAllAttachments(progress, taggedPage.replace(PARAGRAPH_REGEX, '<br />'));
 			this.combineCodeBlocksAsNecessary(html);
-			let parsedPage = this.styledElementToHTML(html);
-			parsedPage = this.convertInternalLinks(parsedPage);
-			parsedPage = this.convertDrawings(parsedPage);
-			parsedPage = this.removeExtraListItemParagraphs(parsedPage);
-			this.escapeTextNodes(parsedPage);
+			this.styledElementToHTML(html);
+			this.convertInternalLinks(html);
+			this.convertDrawings(html);
+			this.removeExtraListItemParagraphs(html);
+			this.escapeTextNodes(html);
 
-			let mdContent = htmlToMarkdown(parsedPage).trim().replace(PARAGRAPH_REGEX, ' ');
+			let mdContent = htmlToMarkdown(html).trim().replace(PARAGRAPH_REGEX, ' ');
 			const fileRef = await this.saveAsMarkdownFile(pageFolder, page.title!, mdContent);
 
 			// Add the last modified and creation time metadata
@@ -575,7 +575,7 @@ export class OneNoteImporter extends FormatImporter {
 	}
 
 	/** Escape characters which will cause problems after converting to markdown. */
-	escapeTextNodes(node: ChildNode) {
+	escapeTextNodes(node: ChildNode): void {
 		if (node.nodeType === Node.TEXT_NODE && node.textContent) {
 			node.textContent = node.textContent
 				.replace(/([<>])/g, '\\$1');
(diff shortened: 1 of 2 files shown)


tgrosinger:
Nice, this looks really good. I'm having trouble getting it to merge two codeblocks that are separated by a newline when that newline is still set to code formatting. Is that expected, or something we can support?

altano:
I dug in and the case you mention appears to be a teeny bit more complicated than you would expect.

This code very early on in the transforms in onenote.ts:

```ts
taggedPage.replace(PARAGRAPH_REGEX, '<br />')
```

Takes consecutive paragraphs and turns them into one paragraph with `br`s. For example, this:

```html
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
```

Is converted to this:

```html
<p style="margin-top:0pt;margin-bottom:0pt">
	<span style="font-family:Consolas">consolas</span>
	<br />
	<span style="font-family:Consolas">consolas</span>
	<br />
	<span style="font-family:Consolas">consolas</span>
	<br />
	<span style="font-family:Consolas">consolas</span>
</p>
```

With my code, those spans and `br` elements will all be processed as sibling code blocks, and combined.

But as soon as you have the newlines, which are `<br />` elements in the original html, you instead get this:

```html
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
<br>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">^^^ consolas newline</span></p>
<p style="margin-top:0pt;margin-bottom:0pt"><span style="font-family:Consolas">consolas</span></p>
```

which is converted to this:

```html
<p style="margin-top:0pt;margin-bottom:0pt">
	<span style="font-family:Consolas">consolas
	</span>
	<br />
	<span style="font-family:Consolas">consolas
	</span>
</p>
<br>
<p style="margin-top:0pt;margin-bottom:0pt">
	<span style="font-family:Consolas">^^^ consolas newline
	</span>
	<br />
	<span style="font-family:Consolas">consolas
	</span>
</p>
```

My code won't combine those spans into one code block anymore because they aren't siblings. They are in different parent paragraphs.

So the fix here is to either tweak the existing regex to combine paragraphs that are only separated by a SINGLE newline `br`, or tweak the code block code to treat paragraphs that only contain code as code blocks themselves, and then sibling paragraph codeblocks can be merged into one contiguous code block.

I don't love either of those options though. I feel like they're both making too many assumptions about the input.

Do you have any thoughts, @tgrosinger ?

altano:
There's one more issue: newlines are translated to `<br />` by OneNote's export, regardless of font type. So if we decided to "fix" this issue by combining code blocks separated by a single newline, we'd be combing code blocks regardless of the font of the newline. So semantically there's no right choice here: either combine all code blocks separated by a newline and mess up code blocks that should have been distinct, or don't combine any code blocks across newlines, and then we're missing some code blocks that should have been combined.

altano:
I'm pushing a new commit that handles `combineCodeBlocksAsNecessary` by taking any instances of `p + br + p` where both `p` only wrap code blocks, and then combining them. It has all the considerations I enumerated above, but I thought I'd add the code to the discussion.

Here's the effect it has:

| OneNote Content | Before | After |
|-|-|-|
| image | image | image |

Note that in the example above, we now combine the code blocks before and after both the Consolas and Calibri newlines. It's both or neither.

tgrosinger:
Hey, thanks for putting so much time and thought into this! I like the direction you've taken it and I think it's really close.

I'm testing again and still running into an issue with it merging code blocks though. I've got a notebook that looks like this:

image

I'm seeing that `isParagraphWrappingOnlyCode` is always returning false because the child node passed to `isCode` is a Text, which doesn't have styles. The parent paragraph does, but the child doesn't.

Let me know if you're able to reproduce that issue too.

altano:
Hey, sorry for the delay. Took a break from this but I'm back and ready to get it done.

I put your sample in a test case and the result looks fine to me:
image

If my result also looks fine to you, maybe we're seeing a different result due to the font of the newlines. All my newlines are Calibri. Can you tell me what your newline fonts are? Since a blank newline can have a font in the onenote output:
image

If my result does NOT look fine to you, let me know what you expect and aren't seeing, as I'm missing it.

Thanks!

tgrosinger:
Hi, my turn to apologize for the delay.

I rewrote my example and imported it again and it's working as expected this time. There must have been something unexpected about my text box that I couldn't see. I don't think we need to try to fix that here though.

Let me do one more quick read through to refresh my memory and then get this merged.

Thanks for your patience.

tgrosinger:
Merged. Thanks @altano!

altano:
My pleasure, thanks for the review. I’ll probably be back with more bug fixes as I try to get my notes imported.

Feel free to tag me on OneNote bugs. For the time being I’ll be active and willing to help.

PR #412, opened 2025-08-11
[Apple Notes] Fix markdown links

### LINKED ISSUE #411: [Apple Notes] The generated markdown links are invalid
If you have `Use [[Wikilinks]]` setting off, the generated markdown links are invalid.

The attachment goes to `/attachment.pdf` and the note goes to `/Apple Notes/note.md` but the link generates is ``, while it has to be ``

Technically speaking, it's an invalid markdown link, which Obsidian tolerate resolving using some heuristics.

mnaoumov:
Fixes #411

### DIFF
--- a/src/formats/apple-notes.ts
+++ b/src/formats/apple-notes.ts
@@ -261,7 +261,7 @@ export class AppleNotesImporter extends FormatImporter {
 		// Notes may reference other notes, so we want them in resolvedFiles before we parse to avoid cycles
 		const converter = this.decodeData(row.zhexdata, NoteConverter);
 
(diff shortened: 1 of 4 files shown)


tgrosinger:
This doesn't seem to be working in my testing. I'm still ending up with a file `/Apple Notes/On My Mac/Example with Attachment.md`, and then the attachment is in `/attachments/example.docx` but the link is ``.

mnaoumov:
@tgrosinger do you have `Relative path` setting enabled?

tgrosinger:
Ah, thank you. I had "shortest path when possible". Changing that setting made it work as expected. Looks good.

PR #419, opened 2025-09-09
Bear import improvements: space and diacritics in tags + note ID links

olivierguerriat:
Starting from https://github.com/obsidianmd/obsidian-importer/pull/382, I went a bit further.

### Make tag flattening optional

As it was a blocker for the source PR, I made it optional (disabled by default).

### Support spaces and diacritics in tags

Tags like `#book/L'électricité du futur#` are now imported as `#book/L_électricité_du_futur`.

### More dates in front matter

I added the `archived` and `trashed` date in front matter (if any).

### Improved handling of note identifiers

Bear gives each note a unique identifier that can be used in its URL scheme. All `bear://x-callback-url/open-note?id=NOTE_ID` URLs are updated so that they work within Obsidian.

Optionally (disabled by default), this id is also added to the front matter. This allows URLs saved elsewhere to be translated to `obsidian://search?vault=VAULT_ID&query=[id:NOTE_ID]`

### Ignore `backup.json`

It was reported as skipped and now properly ignored.

### DIFF
--- a/src/formats/bear-bear2bk.ts
+++ b/src/formats/bear-bear2bk.ts
@@ -299,7 +299,7 @@ export class Bear2bkImporter extends FormatImporter {
 		return content.replace(/bear:\/\/x-callback-url\/open-note\?id=([A-Z0-9\-]+)/g, (match, noteId) => {
 			const noteTitle = idMapping[noteId]?.filename;
 			if (noteTitle) {
-				return encodeURI(noteTitle);
+				return encodeURI(noteTitle.normalize('NFC'));
 			}
 			return match; // No change if ID not found
 		});


olivierguerriat:
I just stumbled upon [Advanced URI's support for `uid` in front matter](https://publish.obsidian.md/advanced-uri-doc/Concepts/File+identifiers#Key+in+frontmatter), so I'll use that key for Bear's identifier.

src/formats/bear-bear2bk.ts
```diff
@@ -261,6 +278,33 @@ export class Bear2bkImporter extends FormatImporter {
 		return contentWithFrontmatter;
 	}
 
+	private updateNotesLinks(idMapping: Record<string, Record<string, any>>): Promise<void> {
+		const updatePromises = Object.values(idMapping).map(async (note) => {
+			const { metadata, file, mdContent } = note;
```
tgrosinger, line 283: If the file has already been written, then we shouldn't be using the `mdContent` cached here. For safety, please instead use `Vault.process` to read, modify, and write the file content.

https://docs.obsidian.md/Reference/TypeScript+API/Vault/process

src/formats/bear-bear2bk.ts
```diff
@@ -72,6 +72,10 @@ export class Bear2bkImporter extends FormatImporter {
 	}
 
 	async import(ctx: ImportContext): Promise<void> {
+
+		// Keep track of Bear IDs to new Obsidian file names to update links based on the identifier
+		let idMapping: Record<string, Record<string, any>> = {};
```
tgrosinger, line 77: Please define a type instead of using `any`.

src/formats/bear-bear2bk.ts
```diff
@@ -261,6 +278,33 @@ export class Bear2bkImporter extends FormatImporter {
...
+					mtime: metadata?.mtime,
+				};
+				await this.vault.modify(file, updatedContent, writeOptions);
+			}
+		});
+		return Promise.all(updatePromises).then(() => {});
+	}
+
+	private updateNoteLinks(idMapping: Record<string, Record<string, any>>, content: string): string {
+		return content.replace(/bear:\/\/x-callback-url\/open-note\?id=([A-Z0-9\-]+)/g, (match, noteId) => {
```
tgrosinger, line 299: Please use a function like `generateMarkdownLink` instead of using the note title directly.

https://docs.obsidian.md/Reference/TypeScript+API/FileManager/generateMarkdownLink

src/formats/bear-bear2bk.ts
```diff
@@ -191,12 +214,21 @@ export class Bear2bkImporter extends FormatImporter {
...
+			frontmatter['id'] = metaData.id;
+		}
 		if (metaData?.ctime) {
 			frontmatter['created'] = new Date(metaData.ctime).toISOString().slice(0, 19);
 		}
 		if (metaData?.mtime) {
 			frontmatter['modified'] = new Date(metaData.mtime).toISOString().slice(0, 19);
 		}
+		if (metaData?.archivedtime) {
+			frontmatter['archived'] = new Date(metaData.archivedtime).toISOString().slice(0, 19);
```
tgrosinger, line 244: Looks like these are already parsed into a date.

src/formats/bear-bear2bk.ts
```diff
@@ -1,25 +1,81 @@
...
+			)
+			.addToggle(t => t
+				.setValue(false)
+				.onChange(async v => this.flattenTags = v)
+			);
+
+		new Setting(this.modal.contentEl)
+			.setName('Store note identifiers in front matter')
+			.setDesc(
+				'This can be useful if you refered to those identifiers elsewhere than in Bear itself.'
```
tgrosinger, line 37: What do you think about something like "Links will be automatically updated. Enable this if the note identifier is outside of linking between notes."

tgrosinger (commented):
Thanks for picking up this change and adding some improvements to it. I definitely like having the tag flattening as an optional feature.

tgrosinger:
Refactored a bit and merged as https://github.com/obsidianmd/obsidian-importer/commit/e3ced4addbd12b31f52d7c5ab327cb0d89066014. Thank you @olivierguerriat  and @roelven.

PR #442, opened 2025-10-23
CSV import with template UI and base file generation

kepano:
Adds three related capabilities:

- CSV import that creates notes with frontmatter properties, and a matching `.base` file
- Note template configuration that allows mapping CSV columns to properties, and uses a similar UI to Web Clipper. Set up so that other importers can add template configuration.
- Bases file generation that can be adopted by other importers .

### Template configuration

A reusable `TemplateConfigurator `class that other importers can use. This enables any importer to provide a template-based configuration step for mapping data fields to note properties.

### Bases module

TypeScript Interfaces

- `BaseOptions`: Top-level configuration with filters, formulas, properties, summaries, and views
- `BaseView`: View configuration with filters, sort, groupBy, limit, and summaries
- `BaseSortConfig`: Multi-level sorting with direction
- `BaseGroupByConfig`: Grouping configuration
- `BasePropertyConfig`: Property display settings

### DIFF
--- a/src/format-importer.ts
+++ b/src/format-importer.ts
@@ -29,6 +29,9 @@ export abstract class FormatImporter {
 	/**
 	 * Optional: Show template configuration UI and prepare data for import.
 	 * This will be called as a configuration step before the import progress.
+	 *
+	 * Overriding functions are responsible for displaying errors before returning false.
+	 *
 	 * @param ctx The import context
 	 * @param container The container element to show the configuration UI in
 	 * @returns true if configuration was successful, false if cancelled or failed, null if no configuration needed


src/base.ts
```diff
@@ -0,0 +1,290 @@
+import { TFolder, TFile } from 'obsidian';
+
+/**
+ * Sort configuration for a view.
+ */
+export interface BaseSortConfig {
```
tgrosinger: I'm wondering if we should export these types in the Obsidian API instead of needing to reproduce them there. If plugins are going to be using them then it would be nice to make sure they get it right and have the up-to-date information from the API.

The downside is that the types we have in the app are a little complicated because they are serialized/deserialized from more involved classes.

src/base.ts
```diff
@@ -0,0 +1,290 @@
...
+ *   views: [{
+ *     type: 'table',
+ *     name: 'Table',
+ *     order: ['file.name', 'title', 'date', 'category'],
+ *     sort: [{ property: 'file.name', direction: 'ASC' }]
+ *   }]
+ * });
+ * ```
+ */
+export function generateBaseContent(options: BaseOptions): string {
```
tgrosinger: Why does this function exist instead of using a yaml library to serialize it? That's what we're doing in the app. If we used the same library then we would know the same input produces the same result.

src/main.ts
```diff
@@ -425,10 +440,26 @@ export class ImporterModal extends Modal {
 						if (this.current) {
 							this.current.cancel();
 						}
+
```
tgrosinger: It looks like your editor is not clearing trailing whitespace. These lines with no content still have a bunch of tabs.

src/main.ts
```diff
@@ -425,10 +440,26 @@ export class ImporterModal extends Modal {
...
+						
+						// Clear content
+						contentEl.empty();
+						let configEl = contentEl.createDiv();
+						let ctx = this.current = new ImportContext(configEl);
+						
+						// Check if importer needs template configuration
+						const templateResult = await importer.prepareTemplateConfig(ctx, configEl);
+						
+						if (templateResult === false) {
```
tgrosinger, line 452: Is an error being shown anywhere when this occurs?

src/template.ts
```diff
@@ -0,0 +1,316 @@
...
+	/** Maps field IDs to property value templates in frontmatter */
+	propertyValues: Map<string, string>;
+	/** Template for generating note body content */
+	bodyTemplate: string;
+}
+
+/**
+ * Default values for template configuration.
+ */
+export interface TemplateDefaults {
```
tgrosinger: This type could be replaced with a partial. That would reduce the duplication and make sure it stays in sync with `TemplateConfig`.

```ts
export type TemplateDefaults = Partial<TemplateConfig>;
```

src/template.ts
```diff
@@ -0,0 +1,316 @@
...
+
+			buttonContainer.createEl('button', { text: 'Cancel' }, el => {
+				el.addEventListener('click', () => {
+					resolve(null);
+				});
+			});
+		});
+	}
+
+	getConfig(): TemplateConfig {
```
tgrosinger: This function appears unused.

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+	private csvHeaders: string[] = [];
+	private csvRows: CSVRow[] = [];
+	private config: TemplateConfig | null = null;
+
+	init() {
+		this.addFileChooserSetting('CSV', ['csv']);
+		this.addOutputLocationSetting('CSV import');
+	}
+
+	async prepareTemplateConfig(ctx: ImportContext, container: HTMLElement): Promise<boolean> {
```
tgrosinger: I might suggest a name that indicates that this is going to show a modal and ask for user input. I was thinking this was just rearranging data until I got to the very bottom of the function. Maybe something like `configureTemplateConfig` or `showTemplateConfiguration`.

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+
+		// Add last line if exists
+		if (currentLine.trim().length > 0) {
+			lines.push(currentLine);
+		}
+
+		return lines;
+	}
+
+	private parseCSVLine(line: string): string[] {
```
tgrosinger, line 224: Should we consider using an off the shelf library for csv parsing, or does this code do everything we need already? [This library](https://www.npmjs.com/package/csv-parse) appears to be the standard and has no dependencies.

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+		return values.map(v => v.trim());
+	}
+
+
+	private sanitizeYAMLKey(key: string): string {
+		// Remove special characters that aren't valid in YAML keys
+		return key.replace(/[^\w\s-]/g, '');
+	}
+
+	private convertToYAML(value: string): string {
```
tgrosinger: I think we can get away with something like `JSON.stringify(value)` instead of needing to implement this. It's okay if there are technically unnecessary quotes around strings in a yaml file.

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+		// Parse CSV files to extract headers
+		for (const file of files) {
+			if (ctx.isCancelled()) return false;
+
+			ctx.status('Parsing ' + file.name);
+			const csvContent = await file.readText();
+			const parsedData = this.parseCSV(csvContent);
+
+			// Store all rows for later processing
+			if (this.csvHeaders.length === 0 && parsedData.rows.length > 0) {
```
tgrosinger: What if the files have different headers?

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+		if (this.csvHeaders.length === 0 || this.csvRows.length === 0) {
+			new Notice('No data found in CSV file(s).');
+			return false;
+		}
+
+		// Prepare template fields
+		const fields: TemplateField[] = this.csvHeaders.map(header => ({
+			id: header,
+			label: header,
+			exampleValue: this.csvRows[0]?.[header] || '',
```
tgrosinger: Maybe this could search through the rows until it finds one that is valid?

src/formats/csv.ts
```diff
@@ -0,0 +1,386 @@
...
+		await this.processRows(ctx);
+	}
+
+	private parseCSV(content: string): { headers: string[], rows: CSVRow[] } {
+		const lines = this.splitCSVLines(content);
+		if (lines.length === 0) {
+			return { headers: [], rows: [] };
+		}
+
+		const headers = this.parseCSVLine(lines[0]);
```
tgrosinger: Generally CSV importers allow you to check a box indicating if your CSV file has headers. We shouldn't assume that the CSV file will always have a header because it runs the risk of skipping the first line of data when importing. Or if headers are required due to the templating then maybe we need to make that very clear when selecting the file.

src/base.ts
tgrosinger: The types are exported in the API now (might need to build locally until the next release). This file should be updated to use those types along with `stringifyYaml` from our API instead of building the string manually.

src/template.ts
```diff
@@ -0,0 +1,301 @@
...
+	const lines = ['---'];
+
+	for (const [fieldId, propertyName] of propertyNames) {
+		if (!propertyName) continue; // Skip if property name is empty
+
+		const valueTemplate = propertyValues.get(fieldId) || '';
+		if (!valueTemplate) continue; // Skip if property value is empty
+
+		// Apply the template to get the actual value
+		const value = applyTemplate(valueTemplate, data);
```
tgrosinger, line 313: We need to make sure that string values do not have newlines. Our app doesn't support that and will show the yaml frontmatter as invalid.

src/main.ts
tgrosinger: Added a note about this to the documentation on that function.

tomByrer:
Thanks  @kepano, I installed Obsidian to transfer CSV into bases!

BTW were you planning to write some docs for it, or should your test.md just be copy/pasted?
https://github.com/obsidianmd/obsidian-help/issues/988

PR #444, opened 2025-10-29
Supports importing via the Notion API

Xheldon:
The initial commits only supported a limited set of modules that are easy to convert to Markdown, such as Paragraph, Heading, List, and Quote. The focus was on handling the conversion from Database to Base, including Relation, Rollup, and Formula.

The submission process also includes refactoring and fixing previous submissions, so during the review, please review each PR individually.

https://github.com/obsidianmd/obsidian-releases/blob/master/cla.md

### DIFF
--- a/src/formats/notion-api.ts
+++ b/src/formats/notion-api.ts
@@ -48,7 +48,7 @@ export class NotionAPIImporter extends FormatImporter {
 	downloadExternalAttachments: boolean = false; // Download external attachments
 	coverPropertyName: string = 'cover'; // Custom property name for page cover
 	databasePropertyName: string = 'base'; // Property name for linking pages to their database
-	incrementalImport: boolean = true; // Incremental import: skip files with same notion-id
+	incrementalImport: boolean = false; // Incremental import: skip files with same notion-id (default: disabled)
 	private notionClient: Client | null = null;
 	private processedPages: Set<string> = new Set();
 	private requestCount: number = 0;


Xheldon:
I refactored the `src/formats/notion-api/formula-converter.ts` file in a subsequent commit, so the previous commits to this file can be ignored.

src/formats/notion-api/api-helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+ * Determine if a formula property should be added to page YAML
+ * Based on the import strategy and whether the formula can be converted
+ */
+function shouldAddFormulaToYAML(
+	propertyKey: string,
+	databaseProperties: any,
+	strategy: 'static' | 'function' | 'hybrid'
+): boolean {
+	// Import formula-converter functions
+	const { canConvertFormula, getNotionFormulaExpression } = require('./formula-converter');
```
tgrosinger: Please move these imports to the top of the file and use import instead of require.

src/formats/notion-api/api-helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+		case 'date':
+			if (!prop.date) return null;
+			// If has time component, use datetime format
+			if (prop.date.start && prop.date.start.length > 10) {
+				return prop.date.end 
+					? `${prop.date.start} to ${prop.date.end}`
+					: prop.date.start;
+			}
+			// Otherwise use date format
+			return prop.date.end
```
tgrosinger, line 512: I don't see how this case is different from the one above.

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,256 @@
...
+		if (i < lines.length - 1) {
+			const currentBlock = blocks[i];
+			const nextBlock = blocks[i + 1];
+			
+			// Check if both current and next are list items
+			const currentIsList = currentBlock.type === 'bulleted_list_item' || currentBlock.type === 'numbered_list_item';
+			const nextIsList = nextBlock.type === 'bulleted_list_item' || nextBlock.type === 'numbered_list_item';
+			
+			// Use single newline between list items, double newline otherwise
+			if (currentIsList && nextIsList) {
```
tgrosinger: Maybe instead, when adding items to lines, if the item being added or the last item is not a list, then add an empty string to lines. Then this section can be fully replaced with a `.join('\n')`

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,256 @@
...
+		
+		case 'child_database':
+			// Database blocks are handled separately in the main importer
+			// Return a placeholder that will be replaced
+			markdown = `<!-- DATABASE_PLACEHOLDER:${block.id} -->`;
+			break;
+		
+		default:
+			// Unsupported block type - skip for now
+			console.log(`Unsupported block type: ${type}`);
```
tgrosinger, line 401: Log to the context?

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,256 @@
...
+	ctx: ImportContext,
+	currentFolderPath: string,
+	client: Client,
+	indentLevel: number = 0
+): Promise<string> {
+	if (block.type !== 'numbered_list_item') return '';
+	
+	// Use 2 spaces per indent level (standard Markdown)
+	const indent = '  '.repeat(indentLevel);
+	let markdown = indent + '1. ' + convertRichText(block.numbered_list_item.rich_text);
```
tgrosinger: Does this number need to be incremented somewhere?

src/formats/notion-api.ts
```diff
@@ -0,0 +1,678 @@
...
+			await this.createFolders(pageFolderPath);
+			
+			// Fetch page blocks (content) with rate limit handling
+			const blocks = await fetchAllBlocks(this.notionClient!, pageId, ctx);
+			
+			// Convert blocks to markdown with nested children support
+			let markdownContent = await convertBlocksToMarkdown(blocks, ctx, pageFolderPath, this.notionClient!);
+			
+		// Process database placeholders
+		markdownContent = await processDatabasePlaceholders(
```
tgrosinger: Incorrect indentation

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+/**
+ * Process database placeholders in markdown content
+ * Replace <!-- DATABASE_PLACEHOLDER:id --> with actual database references
+ */
+export async function processDatabasePlaceholders(
+	markdownContent: string,
+	blocks: any[],
+	ctx: ImportContext,
+	currentPageFolderPath: string,
+	client: Client,
```
tgrosinger: It might be nice to create a config type or something that contains some of these parameters and use it here and for `convertChildDatabase`. This is too many parameters for a function to remain legible.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,678 @@
...
+	}
+	
+	/**
+	 * Replace all relation placeholders with actual links after all pages are imported
+	 */
+	private async replaceRelationPlaceholders(ctx: ImportContext): Promise<void> {
+		if (this.relationPlaceholders.length === 0) {
+			return;
+		}
+
```
tgrosinger, line 1398: Please make sure your editor is trimming trailing white space and be sure to run the linter.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	DATETIME: 'datetime',
+	LIST: 'list',
+	NUMBER: 'number',
+	TEXT: 'text',
+};
+
+/**
+ * Convert a child_database block to Markdown
+ * This creates a reference to a .base file and sets up the database structure
+ * In fact, Notion allows top-level databases, which will be addressed later.
```
tgrosinger: This has been addressed, right? Though the fake block.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+ */
+export async function convertChildDatabase(
+	block: BlockObjectResponse,
+	ctx: ImportContext,
+	currentPageFolderPath: string,
+	client: Client,
+	vault: Vault,
+	app: App,
```
tgrosinger: Unused param

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	// Import each database page recursively
+	for (const page of databasePages) {
+		if (ctx.isCancelled()) break;
+		
+		// Import the page using the callback (which handles the full page import logic)
+		await importPageCallback(page.id, databaseFolderPath);
+	}
+	
+	// Create .base file in "Notion Databases" folder
+	const baseFilePath = await createBaseFile(
```
tgrosinger: Be sure to run linter and formatter please.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+		
+		// In Notion API v2025-09-03, databases have data_sources array
+		// Get the first data source to retrieve properties
+        // It seems that the code I used in the Notion Flow plugin is not compatible with the new 2025-09-03 version of the API.
+        // FIXME: At the same time, I may need to inform the user in the "Import Notes" that the imported database only reads the first data source.
+		
+		let dataSourceProperties: Record<string, any> = {};
+		let propertyIds: string[] = [];
+		
+		if ((database as any).data_sources && (database as any).data_sources.length > 0) {
```
tgrosinger: Why is `as any` required here?

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	ctx: ImportContext,
+	currentPageFolderPath: string,
+	client: Client,
+	vault: Vault,
+	app: App,
+	outputRootPath: string,
+	formulaStrategy: FormulaImportStrategy,
+	processedDatabases: Map<string, DatabaseInfo>,
+	relationPlaceholders: RelationPlaceholder[],
+	importPageCallback: (pageId: string, parentPath: string) => Promise<void>,
```
tgrosinger: This `importPageCallback` feels like the code organization isn't right. Maybe this convertChildDatabase should be on the NotionAPIImporter?

src/formats/notion-api.ts
```diff
@@ -0,0 +1,678 @@
...
+			// Import each database page
+			for (const page of databasePages) {
+				if (ctx.isCancelled()) break;
+				
+				// Import the page
+				await this.fetchAndImportPage(ctx, page.id, databaseFolderPath);
+			}
+			
+			// Create .base file
+			const { createBaseFile } = await import('./notion-api/database-helpers');
```
tgrosinger: Move imports to the top of the file

src/formats/notion-api.ts
```diff
@@ -0,0 +1,678 @@
...
+				console.error(`Failed to replace relation placeholder for page ${placeholder.pageId}:`, error);
+			}
+		}
+	}
+	
+	/**
+	 * Import a database that was not in the original import scope
+	 * but is needed for relation links
+	 */
+	private async importUnimportedDatabase(ctx: ImportContext, databaseId: string, parentPath: string): Promise<void> {
```
tgrosinger, line 1567: Why is there a second code path for importing databases? Can this be merged into the first set of functions?

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+}
+
+/**
+ * Generate content for .base file
+ */
+function generateBaseFileContent(
+	databaseName: string,
+	databaseFolderPath: string,
+	properties: any,
+	databasePages: PageObjectResponse[],
```
tgrosinger: Unused param

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	return finalPath;
+}
+
+/**
+ * Generate content for .base file
+ */
+function generateBaseFileContent(
+	databaseName: string,
+	databaseFolderPath: string,
+	properties: any,
```
tgrosinger: Make this name consistent through all the functions the value is passed through. `notionProperties` or `datasourceProperties` would be better than `properties`.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	propertyIds: string[] = []
+): string {
+	// Basic .base file structure
+	let content = `# ${databaseName}\n\n`;
+	
+	// Add filter to show files in first-level subfolders of the database folder
+	// Logic: file.folder should start with database path, but should only have one more level
+	// Example: if database is "Root/db-1", match "Root/db-1/page-1" but not "Root/db-1/page-1/db-2"
+	// This allows user-created files in the same structure to appear in the base
+	content += `filters:\n`;
```
tgrosinger: I don't like building this yaml by hand. Let's see what happens with https://github.com/obsidianmd/obsidian-importer/pull/442 and reuse the logic or library that it uses to build a base file.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	// Obsidian properties support most characters including spaces and hyphens
+	// Return the original key to ensure consistency with YAML frontmatter
+	return key;
+}
+
+/**
+ * Convert a Notion rollup property to an Obsidian formula
+ * Rollup aggregates values from related pages
+ */
+function convertRollupToFormula(
```
tgrosinger, line 674: Multiple unused params. Is there a type that can be used for `rollupConfig` instead of any?

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+			}
+		}
+	} else {
+		// No propertyIds provided, use the order from Object.entries
+		for (const [key, config] of Object.entries(mappings)) {
+			orderedMappings.push({ key, config });
+		}
+	}
+	
+	return orderedMappings;
```
tgrosinger: Return and object with two lists instead of combining here and then separating again.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,864 @@
...
+	const relationPropertyKey = rollupConfig.relation_property_name || rollupConfig.relation_property_key;
+	const rollupPropertyKey = rollupConfig.rollup_property_name || rollupConfig.rollup_property_key;
+	const rollupFunction = rollupConfig.function;
+	
+	if (!relationPropertyKey || !rollupFunction) {
+		return null;
+	}
+	
+	// Sanitize the relation property key
+	const sanitizedRelationKey = sanitizePropertyKey(relationPropertyKey);
```
tgrosinger, line 692: I'm not following what's happening in this function. Can you give me an example of what the sanitizedRelationKey might be?

Looking at `count` for example, what is the formula `note["${sanitizedRelationKey}"].length` trying to achieve?

tgrosinger (commented):
Thanks @Xheldon, this looks like it's coming along great.

I think there could be a little refactoring to help reduce the long list of arguments being passed between functions.

I've left a list of comments, but I'm sure there will be more in a second pass. Just wanted to send what I have so far over to you.

src/formats/notion-api/api-helpers.ts
Xheldon: Yes, I’ve set up a if branch here to handle different date formats, such as dates like 2025-11-01 and ones that include time like 2025-11-01 23:23. 

However, during processing I found that both can be displayed using the same start → end format. I’ll remove the redundant content.

src/formats/notion-api/block-converter.ts
Xheldon: As I mentioned at the start of my PR, the current commit only supports a limited set of eight modules, so the logic for converting most modules to Markdown hasn't been implemented yet; for now we log it here and await further handling.

src/formats/notion-api/block-converter.ts
Xheldon: Are you referring to the number as indent? Of course, the list can contain sub‑lists, and sub‑lists are indented two spaces relative to the parent list. However… does it need to be a Tab indent? Or four spaces? I tested that two spaces does not cause any conversion errors.

If you mean the indentLevel parameter, yes, Notion’s List is a container, and the List items inside also need to be fetched via the API again, so the nested List here involves the following function calls:
convertBlocksToMarkdown -> convertBlockToMarkdown -> convertNumberedListItem -> convertBlocksToMarkdown(indexLevel + 1) 
recursive call.

CleanShot_2025_10_30_at_09_54_12@2x

CleanShot_2025_10_30_at_09_55_16@2x

src/formats/notion-api.ts
Xheldon: Sorry, my editor uses Prettier for formatting, and I’ll fix these errors later.

src/formats/notion-api.ts
Xheldon: All lint issues have been resolved in a single commit.

src/formats/notion-api/database-helpers.ts
Xheldon: Done.

src/formats/notion-api/database-helpers.ts
Xheldon: Yes, this comment is a bit outdated; I'll modify it.

src/formats/notion-api/database-helpers.ts
Xheldon: Fixed.

src/formats/notion-api/database-helpers.ts
Xheldon: This was an error; the Notion Database API does not return the same property order as the UI, and it has been corrected here.

src/formats/notion-api/database-helpers.ts
Xheldon: Although handling the Database is part of the API import, its logic is quite complex, so I pulled it into a separate file. It also internally includes Pages that need to be imported, with a page‑import callback passed as a parameter. That's why I organized the code this way. Does that convince you, or do you need more reasons?

src/formats/notion-api.ts
Xheldon: done.

src/formats/notion-api.ts
Xheldon: This involves a strategic issue: how to handle DatabaseC that isn’t under the currently imported Root PageA but is referenced by a Relation property of DatabaseB under PageA? My current strategy is to place DatabaseC in a separate folder at the same level as PageA, called “Relation Unimported Databases.”

There are two import Database functions precisely because this concerns two different database import strategies and timings. A normal Database import treats Relation/Rollup properties as placeholders, and in the final step it searches for placeholders based on the already imported Pages and replaces them. At that point, if a Relation points to a Database that does not exist in PageA, the situation described above occurs.

Therefore, I will perform multiple rounds of Database Relation lookup processing until a Database/Page is inaccessible, or there are no Relation properties.

Thus, I recommend not merging these two Database handling functions, as they represent different timings and intentions.

hexqubit:
Hi, I’m waiting for this feature to get merged. I tested the current stage of the feature.

I have a database (gallery view) in Notion, and after importing it, everything works great except that the cover images aren’t imported.

It would be nice if the cover images could also be imported.

Xheldon:
> Hi, I’m waiting for this feature to get merged. I tested the current stage of the feature.
> 
> I have a database (gallery view) in Notion, and after importing it, everything works great except that the cover images aren’t imported.
> 
> It would be nice if the cover images could also be imported.

Thanks for testing! This PR is mainly to let the author see my approach to handling Pages/Databas and to address common issues in earlier community contributions, such as code style and function processing flow. Consequently, the currently supported modules only handle common blocks like Paragraph, Heading, List, Quote, and Database (the main focus). I recommend waiting until the PR is merged into master before using this repository’s code, as unreviewed code could corrupt your Vault (even though I run a quick self‑test on each commit, I can’t guarantee it works for complex cases).

src/formats/notion-api/database-helpers.ts
Xheldon: done.

src/formats/notion-api/database-helpers.ts
Xheldon: done, add RollupConfig type

src/formats/notion-api/database-helpers.ts
Xheldon: done

src/formats/notion-api/database-helpers.ts
Xheldon: The logic here is designed to handle the Rollup property of a Notion database. In Notion, its purpose is to point to a specific attribute of the Page(s) in a related Relation column and then render it.

Within a Rollup property, you can choose how to render that attribute: count means only calculate the quantity, i.e., how many Pages the related Relation column has in that row. count_values calculates how many attribute values those Pages have; for example, if the attribute is a multi‑select with three options, then count_values would be 3, while count would be 1.

The expression `note["${sanitizedRelationKey}"].length` is meant to achieve the same effect as Notion’s count, i.e., compute how many Pages the Relation column points to.

I treat the Relation in the Base as a Pages YAML property of type Link List, because it is static and points to multiple Pages. The Rollup is a Formula type that calculates the length of this Relation.

CleanShot 2025-10-30 at 18 46 24@2x

src/formats/notion-api/block-converter.ts
Xheldon: done.

src/formats/notion-api/database-helpers.ts
tgrosinger: I'm working on adding types to the Obsidian API that can be used for this along with the stringifyYaml function from the Obsidian API.

tgrosinger:
Looks like you're making great progress continuing to add functionality and addressing the feedback I previously left. Please leave a comment mentioning me when you're ready for me to take another pass through the code. Thank you :)

Xheldon:
I think I’ve finished developing this plugin and have self‑tested the import of all modules; everything runs smoothly. 

If you need a Notion test page and token, I’ll add you on Discord—please accept. @tgrosinger

src/formats/notion-api/database-helpers.ts
tgrosinger: #442 has been merged, so this can be converted to use the new function for building base files. Please use `createBaseFile` ([source](https://github.com/obsidianmd/obsidian-importer/blob/master/src/base.ts#L25)) instead of building yaml manually.

src/formats/notion-api/attachment-helpers.ts
```diff
@@ -0,0 +1,388 @@
...
+			path: attachment.url,
+			isLocal: false
+		};
+	}
+}
+
+/**
+ * Get file extension from Content-Type header
+ */
+function getExtensionFromContentType(contentType: string): string | null {
```
tgrosinger: Please use `extensionForMime` ([source](https://github.com/obsidianmd/obsidian-importer/blob/master/src/mime.ts#L987))

src/formats/notion-api.ts
tgrosinger: I just read through `importUnimportedDatabase` and `convertChildDatabase` and they are virtually identical. There are a couple pieces of information that are retrieved in slightly different ways, but as far as I've noticed they are doing the exact same thing. Please deduplicate this code.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,861 @@
...
+		ctx.reportFailed(`Database ${block.id}`, error.message);
+		return `<!-- Failed to import database: ${error.message} -->`;
+	}
+}
+
+/**
+ * Query all pages from a database with pagination support
+ * Note: In Notion API v2025-09-03, databases are now called "data sources"
+ */
+async function queryAllDatabasePages(
```
tgrosinger: This function is identical to `queryDatabasePages`. Please deduplicate.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1210 @@
...
+				// Try as page first
+					const pageId = childId;
+					const pagePlaceholder = `[[SYNCED_CHILD_PAGE:${pageId}]]`;
+				
+					// Check if this is a page placeholder
+					if (content.includes(pagePlaceholder)) {
+						// Check if page is already imported
+						let pagePath = this.notionIdToPath.get(pageId);
+				
+						if (pagePath) {
```
tgrosinger, line 1757: Invert the order here so that if it has not been imported yet, then the import is performed. Then outside of the if-block whether it was imported or not, perform the content replacement. Shouldn't need to duplicate the `if(targetFile && targetFile instanceof TFile)` block.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1210 @@
...
+								// Import database using importTopLevelDatabase
+								await this.importTopLevelDatabase(ctx, databaseId, syncedBlocksFolder);
+								importedCount++;
+						
+								// Now get the database info
+								const newDbInfo = this.processedDatabases.get(databaseId);
+								if (newDbInfo) {
+									const baseFilePath = newDbInfo.baseFilePath.replace(/\.base$/, '');
+									const targetFile = this.vault.getAbstractFileByPath(baseFilePath + '.base');
+									if (targetFile && targetFile instanceof TFile) {
```
tgrosinger: Same comment as above. Flip the order of the if-block so that this can be deduplicated.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1210 @@
...
+				}
+			
+				// Save the file if it was modified
+				if (content !== originalContent) {
+					await this.vault.modify(file, content);
+					filesModified++;
+				}
+			}
+			catch (error) {
+				console.error(`Failed to process synced child placeholders in file ${filePath}:`, error);
```
tgrosinger: This should be reported in the context as an error instead of in the console.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1210 @@
...
+								// Failed to import (no access or error)
+								console.warn(`Failed to import synced child page ${pageId}:`, error);
+								content = content.replace(pagePlaceholder, `**Page** _(no access)_`);
+							}
+						}
+					}
+			
+					// Check if this is a database placeholder
+					const databaseId = childId;
+					const dbPlaceholder = `[[SYNCED_CHILD_DATABASE:${databaseId}]]`;
```
tgrosinger: Please make a function which takes a `databaseId` and returns this string. It should be used here and in the places where the placeholders are being created to ensure that the format stays in sync. Please do the same for the `SYNCED_CHILD_PAGE` placeholder above too.

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,1417 @@
...
+ * Convert a Notion synced_block to Obsidian wiki link
+ * Handles both original blocks and synced copies
+ */
+export async function convertSyncedBlock(
+	block: BlockObjectResponse,
+	context: BlockConversionContext
+): Promise<string> {
+	if (block.type !== 'synced_block') return '';
+	
+	const syncedBlockData = (block as any).synced_block;
```
tgrosinger: Remove the `as any`. Doesn't seem to be required here because `block` has `synced_block` in the type.

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,1417 @@
...
+			syncedBlocksMap.set(originalBlockId, filePath);
+		}
+		catch (error) {
+			console.error(`Failed to process synced block ${originalBlockId}:`, error);
+			return `<!-- Failed to import synced block: ${error.message} -->`;
+		}
+	}
+	
+	// Extract filename without extension for wiki link
+	const fileName = filePath.split('/').pop()?.replace(/\.md$/, '') || 'Synced Block';
```
tgrosinger: Please use `filesystem.parseFilePath`

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,1417 @@
...
+): Promise<string> {
+	const { client, ctx, vault, outputRootPath } = context;
+	
+	if (!outputRootPath) {
+		throw new Error('outputRootPath is required for synced blocks');
+	}
+	
+	try {
+		// Fetch the block to get its content
+		const block = await client.blocks.retrieve({ block_id: blockId }) as BlockObjectResponse;
```
tgrosinger: Why is this cast safe? Maybe should check the response instead of casting.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1210 @@
...
+				}
+		
+				let content = await this.vault.read(file);
+				const originalContent = content;
+			
+				// Process each child ID that was recorded for this file
+				for (const childId of childIds) {
+				// Try as page first
+					const pageId = childId;
+					const pagePlaceholder = `[[SYNCED_CHILD_PAGE:${pageId}]]`;
```
tgrosinger: Why are the `SYNCED_CHILD_PAGE` and `SYNCED_CHILD_DATABASE` placeholders being added to the same list if they need to be treated differently? Maybe they should be stored in two lists so we don't have to do unnecessary scans of the content for the wrong type.

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,1417 @@
...
+	
+	const indentLevel = context.indentLevel || 0;
+	const indent = '    '.repeat(indentLevel); // 4 spaces per indent level
+	
+	// Use [x] for checked items, [ ] for unchecked
+	const checkbox = block.to_do.checked ? '[x]' : '[ ]';
+	let markdown = indent + '- ' + checkbox + ' ' + convertRichText(block.to_do.rich_text, context);
+	
+	// Check if this block has children
+	if (block.has_children) {
```
tgrosinger: This pattern of checking for children, fetching children, storing, converting, and then doing something is done over and over again in these converters. Maybe we can find a higher abstraction that would allow factoring out and deduplicating a lot of this functionality? For example, maybe make a `function withBlockChildren(block: BlockObjectResponse, callback: (children) => void)` which can perform the checking, fetching, etc and then just execute the unique logic we need to perform in each converter.

src/formats/notion-api/attachment-helpers.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/a3e7a73b0a25edd88446cd3f90c69b2c30406173

src/formats/notion-api/database-helpers.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/a0acdeec62d1ed730b1015de25599b42957db1ee

src/formats/notion-api.ts
Xheldon: Done, and the following as well.. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/37ba3522e6f4ab7c63178a723cc96c87c99eed0a

src/formats/notion-api.ts
Xheldon: Yes, this is a fatal error and I should report it to the user. I've handled similar cases in a few other places in the same way. Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/08dab198827b6950d364b6fa4f84821a04f29fc0

src/formats/notion-api.ts
Xheldon: Done. All placeholder have be handled. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/51a9e94bda34386051c90555f197a9434aa9201a

src/formats/notion-api/block-converter.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/26c5fc06ec6b908d166e656589fc4d42236225bc
I added comments to the places where using `any` is necessary.

src/formats/notion-api/block-converter.ts
Xheldon: It has been corrected. At the same time, I checked the utils files and found that some filename‑normalization utility functions can be reused, so I updated them as well. Additionally, I also handled `normalizePath`. See: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/2dc170f7ac517c79ad6edb9598c4f0bb5e9e3635

src/formats/notion-api/block-converter.ts
Xheldon: Hmm, you're right, this might also return a PartialBlockObjectResponse. Fixed: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/7037955556c8945259d6b5edc1caaa3baba3bea0

src/formats/notion-api.ts
Xheldon: My reason for distinguishing the two is that the import methods for childpages and childdatabase differ, but putting them in the same Set was indeed inappropriate, which forced me to guess and then search. I have refactored this part, handling it with two Sets. Fixed: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/72fce1990272416009bd7c2f343cdc6808120c7e

src/formats/notion-api/block-converter.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/edd268dbd5477c902ac287cddd96019b61693089

 TODO: Some attachment types may require the same handling.

src/formats/notion-api.ts
tgrosinger: Just a reminder about this comment.

Xheldon: OK, I'll try to sort this out today.

src/formats/notion-api/block-converter.ts
Xheldon: Attachment has Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/ecf820f1aef908cfc95d3a1d96bad51b316c7dc0

src/formats/notion-api.ts
Xheldon: Done. Extracted a function called importDatabaseCore https://github.com/obsidianmd/obsidian-importer/pull/444/commits/05ab91e19e6256b1f12c9d5f26b2bdd9b48ab53a

src/formats/notion-api/database-helpers.ts
Xheldon: Just a reminder here so you don’t miss some messages on Discord: it seems you haven’t exported the `BasesConfigFile` interface in Obsidian’s ts type package.

I saw you used it in the recent CSV importer, but I can’t get its type declaration. Please fix this so I can continue the refactor @tgrosinger

tgrosinger: It's there now, sorry for the delay.

https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L547

src/formats/notion-api.ts
tgrosinger: Great, thank you!

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+						this.attachmentsDownloaded++;
+						ctx.attachments = this.attachmentsDownloaded;
+						ctx.attachmentCountEl.setText(this.attachmentsDownloaded.toString());
+					
+						// Extract extension from filename
+						const ext = result.filename.substring(result.filename.lastIndexOf('.'));
+						const fullPath = result.path + ext;
+						// Use wiki link syntax with double quotes: "[[path]]"
+						// The double quotes are necessary for YAML to render it as a link
+						const coverValue = `[[${fullPath}]]`;
```
tgrosinger, line 1296: It looks like the path for the cover image is incorrect. Clicking on it shows a notice "Folder already exists". This path should probably be created using `app.fileManager.generateMarkdownLink`

image

tgrosinger:
I'm not sure where these empty folders called "Untitled" came from.

image

Probably has something to do with these import errors?

image

I am not sure what I need to share with my integration to make that work though.

The owner did not import properly.

image

tgrosinger:
It seems like I cannot import a sub-page under a database? For example this one 

https://www.notion.so/e5c55b3c07ad4d53aad53cea0fe2523b?v=04ac95662356414299cecaaa8ad5066c&source=copy_link

image

Apologies if this is a user error. I'm not super familiar with Notion.

src/formats/notion-api.ts
tgrosinger: I'm noticing this problem in all places where references are made to attachments.

tgrosinger:
image

I think it would be more intuitive for the `db-2.base` to be located as a sibling of the `db-2` folder, rather than grouping all the `.base` files together.

tgrosinger:
image

I was testing some of the other rollup functions and it looks like the numbers functions aren't supported yet.

tgrosinger:
This is looking really great so far @Xheldon! It's exciting to start testing and seeing it in action. Thanks for all of your work on this so far.

src/formats/notion-api/database-helpers.ts
Xheldon: Thanks, this part of the refactoring is complete; see https://github.com/obsidianmd/obsidian-importer/pull/444/commits/51051caef84119ae46aaff08ebd453bacba8eb48

src/formats/notion-api.ts
Xheldon: The font in the screenshot suggests you’re using Windows, and your link format is `Relative to file`, right?
I haven’t tested Windows yet, and my option is `Shortest path when possible`. I’m not sure if either of these is affecting the issue, because I use the normalizePath function everywhere a path appears. I happen to have a Windows computer, so I’ll test it today to see what’s going on. The screenshot below shows my Mac attachment path:
CleanShot_2025_11_13_at_09_40_41@2x
CleanShot_2025_11_13_at_09_42_17@2x

Xheldon:
@tgrosinger Here I’ll quickly answer your question, and also clarify things for anyone else seeing this PR.

1. Currently, generating a Notion token requires you to select an authorization scope, and pages that are not within this scope will cause an error indicating that they cannot be found when you try to import (due to lack of permission; a typical scenario is that the Pages/Databases you are importing reference pages that the token does not have access to). I may have already reported this error to users as an import failure; I will double‑check today. If you are sure this is a bug, you can share the relevant database with me (please remove any sensitive information) and I’ll try to fix it. Also, the “Untitled” folder names come from me trying to use the database’s title; if you haven’t named a database, it will be called “Untitled”, and if there are duplicates it will appear as “Untitled (n)”.

2. Under normal circumstances, you should be able to import the sub‑pages under a Database; see the Notion Pages test page I shared with you. If it doesn’t work, could you share the page with me so I can see where the problem is?

3. Regarding the location of the Base file, this is a strategy matter; you can discuss it with @kepano in the group. Previously we agreed to put the Base in a unified directory for easier management, as placing it directly in a specific folder could be too scattered.

4. Yes, some functions are not supported because Obisdian's functions do not fully support the complex functions in Notion. As you can see in the log, functions such as sum, average, etc., are not implemented in Obsidian. My suggestion is to add these actual functions to Obsidian (there may be many) rather than forcibly implementing them through hacky methods, which could make the code look unsightly.

src/main.ts
```diff
@@ -293,6 +310,13 @@ export default class ImporterPlugin extends Plugin {
 				helpPermalink: 'import/notion',
 				formatDescription: 'Export your Notion workspace to HTML format.',
 			},
+			'notion-api': {
+				name: 'Notion (via API)',
+				optionText: 'Notion (via API)',
+				importer: NotionAPIImporter,
+				helpPermalink: 'import/notion-api',
```
kepano: Use `import/notion` (we won't have a separate page)

README.md
```diff
@@ -15,6 +15,7 @@ Import guides are hosted on the [official Obsidian Help site](https://help.obsid
 - [Import from Google Keep](https://help.obsidian.md/import/google-keep)
 - [Import from Microsoft OneNote](https://help.obsidian.md/import/onenote)
 - [Import from Notion](https://help.obsidian.md/import/notion)
+- [Import from Notion API(WIP)](https://help.obsidian.md/import/notion-api)
```
kepano: Use `import/notion` (we won't have a separate page)

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+			.setDesc(this.createImportDescription())
+			.setHeading();
+	}
+
+	private createTokenDescription(): DocumentFragment {
+		const frag = document.createDocumentFragment();
+		frag.appendText('Enter your Notion integration token. ');
+		frag.createEl('a', {
+			text: 'Learn how to get your token',
+			href: 'https://developers.notion.com/docs/create-a-notion-integration',
```
kepano: Link directly to the integrations page: https://www.notion.so/profile/integrations

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+	private syncedChildPagePlaceholders: Map<string, Set<string>> = new Map();
+	private syncedChildDatabasePlaceholders: Map<string, Set<string>> = new Map();
+
+	init() {
+		// No file chooser needed since we're importing via API
+		this.addOutputLocationSetting('Notion API Import');
+
+		// Notion API Token input
+		new Setting(this.modal.contentEl)
+			.setName('Notion API Token')
```
kepano: Use sentence case: `Notion API token`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+			.addText(text => text
+				.setPlaceholder('cover')
+				.setValue('cover')
+				.onChange(value => {
+					this.coverPropertyName = value.trim() || 'cover';
+				}));
+
+		// Base view type setting
+		new Setting(this.modal.contentEl)
+			.setName('Database view type')
```
kepano: Change to `Default view type`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+		frag.appendText('Leave as "cover" if you don\'t have conflicts with existing properties.');
+		return frag;
+	}
+
+	private createBaseViewTypeDescription(): DocumentFragment {
+		const frag = document.createDocumentFragment();
+		frag.appendText('Choose the default view type for database .base files:');
+		frag.createEl('br');
+		frag.createEl('br');
+		frag.appendText('• Table: Default table view, suitable for all database types');
```
kepano: Remove descriptions of each view type. It's not necessary.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+					this.notionToken = value.trim();
+				})
+				.then(textComponent => {
+					// Set as password input
+					textComponent.inputEl.type = 'password';
+				}));
+
+		// Page ID input
+		new Setting(this.modal.contentEl)
+			.setName('Page ID')
```
kepano: This field should not be required. By default we should import everything available from the token. Alternatively we can try an experience similar to the OneNote importer where users can select which pages to import (defaults to all).

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1147 @@
...
+			.setDesc(this.createPageIdDescription())
+			.addText(text => text
+				.setPlaceholder('Enter Notion page ID Or Page URL')
```
kepano: this placeholder text is too long

tgrosinger:
> Currently, generating a Notion token requires you to select an authorization scope, and pages that are not within this scope will cause an error indicating that they cannot be found when you try to import (due to lack of permission; a typical scenario is that the Pages/Databases you are importing reference pages that the token does not have access to).

I did make sure that the two databases in my Product Roadmap page were shared and it said they were shared because they were included in Product Roadmap. I'm not sure what else I'm missing.

src/formats/notion-api.ts
Xheldon: Okay, I’ll try to build a selection tree that lets users choose integrated pages they have permission to import (multiple selection allowed)

src/formats/notion-api.ts
Xheldon: As mentioned earlier, I will logically delete the input PageID

Xheldon:
> > Currently, generating a Notion token requires you to select an authorization scope, and pages that are not within this scope will cause an error indicating that they cannot be found when you try to import (due to lack of permission; a typical scenario is that the Pages/Databases you are importing reference pages that the token does not have access to).
> 
> I did make sure that the two databases in my Product Roadmap page were shared and it said they were shared because they were included in Product Roadmap. I'm not sure what else I'm missing.

As we discussed on Discord, your page contains a linked database, which is a type not supported by the Notion API. Therefore, I have submitted the relevant code to handle this situation, and now it won’t throw an error or create an empty “Untitled” folder. See: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/d60d8283408aec46ea3f6b2615d6871ddc2e6113

Regarding linked database, see https://developers.notion.com/docs/working-with-databases#additional-types-of-databases

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

README.md
Xheldon: done, just remove it. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/main.ts
Xheldon: done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/9a03fc09d94daa989236829c49d25e98b6975843

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+	private syncedBlocksMap: Map<string, string> = new Map();
+	// Track synced child placeholders (file path -> Set of child IDs)
+	// Used to efficiently replace synced child placeholders without scanning all files
+	// Separated by type to avoid unnecessary placeholder checks
+	private syncedChildPagePlaceholders: Map<string, Set<string>> = new Map();
+	private syncedChildDatabasePlaceholders: Map<string, Set<string>> = new Map();
+
+	init() {
+		// No file chooser needed since we're importing via API
+		this.addOutputLocationSetting('Notion API Import');
```
kepano: Rename to just `Notion` (like the existing zip importer)

src/main.ts
```diff
@@ -293,6 +294,13 @@ export default class ImporterPlugin extends Plugin {
 				helpPermalink: 'import/notion',
 				formatDescription: 'Export your Notion workspace to HTML format.',
 			},
+			'notion-api': {
+				name: 'Notion (via API)',
```
kepano: change to `Notion (API)`

src/main.ts
```diff
@@ -293,6 +294,13 @@ export default class ImporterPlugin extends Plugin {
...
+				optionText: 'Notion (via API)',
```
kepano: change to `Notion (API)`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+		const placeholder = this.pageTreeContainer.createDiv();
+		placeholder.addClass('notion-tree-placeholder');
+		placeholder.style.color = 'var(--text-muted)';
+		placeholder.style.textAlign = 'center';
+		placeholder.style.padding = '30px 10px';
+		placeholder.setText('Click "List importable pages" to load your Notion pages and databases.');
+
+		// Formula import strategy
+		new Setting(this.modal.contentEl)
+			.setName('Formula import strategy')
```
kepano: Rename setting name to `Convert formulas`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+			.setDesc(this.createFormulaStrategyDescription())
+			.addDropdown(dropdown => {
+				dropdown
```
kepano, line 195: Rename the options, and reduce to only two options. We should only use `hybrid` and `static`.

- `To Obsidian syntax` — set as default
- `To static values`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+		frag.createEl('a', {
+			text: 'Learn how to get your token',
+			href: 'https://www.notion.so/profile/integrations',
+		});
+		frag.appendText('.');
+		return frag;
+	}
+
+	private createFormulaStrategyDescription(): DocumentFragment {
+		const frag = document.createDocumentFragment();
```
kepano, line 250: Simplify this text. Remove the bullets and change to

```
By default Notion formulas are converted to Obsidian syntax. If any Notion syntax is not supported the static values will be saved instead. Alternatively you can import all formulas as static values.
```

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+			.setName('Import notes')
+			.setDesc(this.createImportDescription())
+			.setHeading();
+	}
+
+	private createTokenDescription(): DocumentFragment {
+		const frag = document.createDocumentFragment();
+		frag.appendText('Enter your Notion integration token. ');
+		frag.createEl('a', {
+			text: 'Learn how to get your token',
```
kepano: Change to `Get API token`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+		// Description text
+		new Setting(this.modal.contentEl)
+			.setName('Import notes')
+			.setDesc(this.createImportDescription())
+			.setHeading();
+	}
+
+	private createTokenDescription(): DocumentFragment {
+		const frag = document.createDocumentFragment();
+		frag.appendText('Enter your Notion integration token. ');
```
kepano: Change to `To get an API token create an integration in Notion and give it access to pages in your workspace.`

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+				ctx
+			);
+			
+			// Extract page title
+			const pageTitle = extractPageTitle(page);
+			// Use custom file name if provided, otherwise use page title
+			const sanitizedTitle = customFileName ? sanitizeFileName(customFileName) : sanitizeFileName(pageTitle || 'Untitled');
+	
+			// Update status with page title instead of ID
+			ctx.status(`Importing: ${sanitizedTitle}...`);
```
kepano: I believe the `...` is not necessary because it's already added?

src/main.ts
```diff
@@ -293,6 +294,13 @@ export default class ImporterPlugin extends Plugin {
 				helpPermalink: 'import/notion',
 				formatDescription: 'Export your Notion workspace to HTML format.',
 			},
+			'notion-api': {
```
kepano: Move the Notion API importer above the .zip importer in the list

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+			.addToggle(toggle => toggle
+				.setValue(true) // Default to enabled
+				.onChange(value => {
+					this.incrementalImport = value;
+				}));
+
+		// List pages and toggle selection buttons
+		const listPagesSetting = new Setting(this.modal.contentEl)
+			.setName('Select pages to import')
+			.setDesc('Click the button below to list all pages and databases you can import.');
```
tgrosinger: Not necessarily below. Maybe something like this? 

> Click "Refresh list" to see all of the importable pages and databases. If a page or database is missing, verify the Notion integration has access to it.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+				}));
+
+		// Incremental import setting
+		new Setting(this.modal.contentEl)
+			.setName('Incremental import')
+			.setDesc('Skip files that already exist with the same notion-id. Disable to always import (will rename duplicates).')
+			.addToggle(toggle => toggle
+				.setValue(true) // Default to enabled
+				.onChange(value => {
+					this.incrementalImport = value;
```
tgrosinger: With this enabled, I imported the Product Roadmap twice with the same settings. It seemed to work except the "Product Roadmap.png" was imported again with a number appended.

src/util.ts
```diff
@@ -32,7 +32,9 @@ export function parseHTML(html: string): HTMLElement {
 }
 
 export function uint8arrayToArrayBuffer(input: Uint8Array): ArrayBuffer {
-	return input.buffer.slice(input.byteOffset, input.byteOffset + input.byteLength);
+	// Slice to ensure we only return the portion of the buffer that corresponds to this view
+	// Use slice which creates a new ArrayBuffer (not SharedArrayBuffer)
+	return input.buffer.slice(input.byteOffset, input.byteOffset + input.byteLength) as ArrayBuffer;
```
tgrosinger, line 37: I'm not sure that this casting is necessary since the function has an explicit return type.

src/formats/notion-api/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Replace a placeholder with actual content
+ * @param content - Markdown content containing placeholders
+ * @param type - Type of placeholder to replace
+ * @param id - Notion ID to match
+ * @param replacement - Replacement string
+ * @returns Updated content
+ */
+export function replacePlaceholder(
```
tgrosinger: This function appears to be unused.

src/formats/notion-api.ts
```diff
@@ -0,0 +1,1893 @@
...
+							const displayName = relatedPageFile.basename; // Just the file name for display
+							const wikiLink = `"[[${fullPath}|${displayName}]]"`;
+							
+							// Replace the page ID with the wiki link in the YAML
+							// The page IDs are stored as array items in YAML
+							newContent = newContent.replace(
+								new RegExp(`"${relatedPageId}"`, 'g'),
+								wikiLink
+							);
+							newContent = newContent.replace(
```
tgrosinger, line 1535: Why is this happening twice?

src/formats/notion-api/vault-helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Functions for interacting with Obsidian vault
+ */
+
+import { App, Vault, normalizePath } from 'obsidian';
+
+/**
+ * Check if a page with the given notion-id already exists in the vault
+ * FIXME: This could cause performance problems if the vault is too large, so consider only performing duplicate detection on the pages/database currently being imported—since the user might actually want to import twice.
+ */
+export async function pageExistsInVault(app: App, vault: Vault, notionId: string): Promise<boolean> {
```
tgrosinger: This function is unused.

src/formats/notion-api/formula-converter.ts
```diff
@@ -0,0 +1,576 @@
...
+ * - length(x) -> (x).length (property access)
+ * - abs(x) -> (x).abs() (method call)
+ * - contains(x, y) -> (x).contains(y) (method call)
+ * - unique(x) -> (x).unique() (method call)
+ * - Global functions stay as-is: max(), min(), if()
+ * 
+ * Important notes:
+ * - length is a PROPERTY (x.length), not a function
+ * - unique() is a METHOD (x.unique()), not a global function
+ * - sum() and average() are NOT supported in Obsidian Bases
```
tgrosinger: These functions are supported on list types in an upcoming release.

src/formats/notion-api/formula-converter.ts
```diff
@@ -0,0 +1,576 @@
...
+			}
+		
+			// Special case: Notion's dateAdd() function -> Obsidian date arithmetic
+			// In Notion: dateAdd(date, amount, unit)
+			// In Obsidian: date + 'amount+unit' (e.g., now() + '1d')
+			if (funcName === 'dateAdd') {
+				changed = true;
+				const args = parseArguments(argsStr);
+				if (args.length === 3) {
+					const dateArg = args[0];
```
tgrosinger: This logic of splitting, removing the quotes, and mapping the units is repeated again directly below. Please pull out into a helper function.

src/formats/notion-api/block-converter.ts
```diff
@@ -0,0 +1,1348 @@
...
+				if (!error.message?.includes('already exists')) {
+					const errorMsg = error instanceof Error ? error.message : String(error);
+					console.error('Failed to create Notion Synced Blocks folder:', error);
+					context.ctx.reportFailed('Create Synced Blocks folder', errorMsg);
+				}
+			}
+		}
+		
+		// Generate unique file path
+		let filePath = normalizePath(`${syncedBlocksFolder}/${fileName}.md`);
```
tgrosinger: Use `getUniqueFilePath` from `vault-helpers.ts`?

src/formats/notion-api/attachment-helpers.ts
```diff
@@ -0,0 +1,358 @@
...
+
+/**
+ * Format attachment link according to vault settings
+ * @param result - Attachment download result
+ * @param vault - Obsidian vault
+ * @param caption - Optional caption/alt text
+ * @param isEmbed - Whether to use embed syntax (!) for images/videos/pdfs
+ * @returns Formatted markdown link
+ */
+export function formatAttachmentLink(
```
tgrosinger: Can some or all of this function be replaced with `app.fileManager.generateMarkdownLink`?

src/formats/notion-api/attachment-helpers.ts
```diff
@@ -0,0 +1,358 @@
...
+		else {
+			return `[${displayText}](${pathWithExt})`;
+		}
+	}
+}
+
+/**
+ * Get file extension from filename
+ */
+function getFileExtension(filename: string): string {
```
tgrosinger: This can be replaced with `filesystem.splitext`.

src/formats/notion-api/database-helpers.ts
```diff
@@ -0,0 +1,1010 @@
...
+	
+	// For incremental import: update existing .base file if it exists
+	const existingFile = vault.getAbstractFileByPath(baseFilePath);
+	if (existingFile instanceof TFile) {
+		// Update existing .base file with latest database properties
+		await vault.modify(existingFile, baseContent);
+		return baseFilePath;
+	}
+	
+	// File doesn't exist, create new one with unique name if needed
```
tgrosinger, line 377: Use `getUniqueFilePath`?

src/formats/notion-api.ts
Xheldon: I forgot to mention, this is indeed a bug, and I have already fixed it.

src/formats/notion-api.ts
Xheldon: This was intentionally designed this way because determining that two attachments are exactly the same is quite difficult. Even if the two images have identical dimensions, size, and even creation dates (some hash calculations depend on this value), I cannot be certain they are the same image. Therefore, to avoid losing user data, I did not use an incremental import strategy for attachments.
Regarding this, do you have any good solutions or suggestions?

src/util.ts
Xheldon: My TS type checker threw an error, telling me that `Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'`, so I had to add a type assertion.   
CleanShot 2025-11-20 at 18 15 53@2x

src/formats/notion-api.ts
Xheldon: This might be a bug. During development I tried using YAML attributes without double quotes, found that the attribute didn’t work, and then added a version with double quotes. I’ll remove the version without double quotes and test to see if it works correctly.

src/formats/notion-api/vault-helpers.ts
Xheldon: Deleted, the commit will be attached later. This is leftover code that I forgot to delete earlier; initially I used this traversal method to find already imported pages, but later refactored to track imported pages using Map/Set, reducing O(n) to O(1).

src/formats/notion-api/formula-converter.ts
Xheldon: Okay, I’ll do the conversion based on the file you gave me. If there are additional functions supported later, just let me know, and I’ll keep working on this conversion issue.

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/main.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/main.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Yes, done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/main.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Right, Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api/utils.ts
Xheldon: Removed, done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api/vault-helpers.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/f73ac30206c9670ed465074c1dbade96d3196662

src/formats/notion-api/formula-converter.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/a1db5e1a6b5bbb88aea907ed2ab46cb9a2504478

src/formats/notion-api/block-converter.ts
Xheldon: Yeah, I should use getUniqueFolderPath and getUniqueFilePath in any unique path place
Done, https://github.com/obsidianmd/obsidian-importer/pull/444/commits/909e3e421d4b3dd21ab17601a5b4fa1c3fc4994e

src/formats/notion-api/database-helpers.ts
Xheldon: Done, https://github.com/obsidianmd/obsidian-importer/pull/444/commits/909e3e421d4b3dd21ab17601a5b4fa1c3fc4994e

src/formats/notion-api/attachment-helpers.ts
Xheldon: No, There have some fallback logic in this function, but I refactor part of this logic to use the build-in interface generateMarkdownLink, see: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/2bb97c05ae1b55de519839d00dbf5fb5a0c558b2

src/formats/notion-api/attachment-helpers.ts
Xheldon: You are right. Done: https://github.com/obsidianmd/obsidian-importer/pull/444/commits/1248bb9b80b37653b5c61cb16529920dca4fad6a

src/formats/notion-api.ts
Xheldon: Here I finally removed the quoted version, because when generating placeholders, the YAML property's Page id is not enclosed in double quotes. See the subsequent commit.

src/formats/notion-api.ts
tgrosinger: I think @kepano mentioned to try to just match based on filename and file size, which sounds good to me.

Xheldon: Okay, I will match based on the attachment’s filename (including path) and file size

Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/444/commits/23144314745fb4c8f4c92fbef2f58e429808f712

PR #445, opened 2025-10-29
Tomboy / GNote importer

farnaby:
Hello, I made a Tomboy / GNote importer extension, mainly for myself as I had a large collection of Tomboy notes and wanted to migrate them to Obsidian. Exporting html from Tomboy and re-importing it in Obsidian didn't work very well, so I wrote my own importer.

It has the following features:

- Maintains nested lists
- Maintains internal links, even if titles contain symbols that have to be escaped
- Maintains most formatting (not perfect in certain edge cases, but practically works very well)
- Detects headings (heuristic)
- Detects TODO lists (heuristic, optional)
- Handling of note titles is configurable
- Metadata (notebook, date) is transferred to md frontmatter

I've also added a hint to the file browser to help the user find the note directories to be imported.

Feel free to use this code as needed!

### DIFF
--- a/src/formats/tomboy/core.ts
+++ b/src/formats/tomboy/core.ts
@@ -291,23 +291,15 @@ export class TomboyCoreConverter {
 				break;
 		}
 
-		if (shouldKeepTitle) {
-			// Keep the title - format it as H1 if it's the first line
-			const lines = markdownContent.split('\n');
-			if (lines.length > 0) {
-				const escapedTitle = this.escapeMarkdownSpecialChars(note.title);
-				if (lines[0].trim() === escapedTitle.trim()) {
-					// Replace the first line with H1 formatted title
-					lines[0] = `# ${note.title}`;
-					markdownContent = lines.join('\n');
-				}
+		const lines = markdownContent.split('\n');
+		const escapedTitle = this.escapeMarkdownSpecialChars(note.title);
+		if (lines.length > 0 && lines[0].trim() === escapedTitle.trim()) {
+			if (shouldKeepTitle) {
+				// Replace the first line with H1 formatted title
+				lines[0] = `# ${note.title}`;
+				markdownContent = lines.join('\n');
 			}
-		}
-		else {
-			// Remove the title - same logic as before, but with proper escaping comparison
-			const lines = markdownContent.split('\n');
-			const escapedTitle = this.escapeMarkdownSpecialChars(note.title);
-			if (lines.length > 0 && lines[0].trim() === escapedTitle.trim()) {
+			else {
 				// Remove the title line and any empty lines that follow it
 				let startIndex = 1;
 				while (startIndex < lines.length && lines[startIndex].trim() === '') {


src/formats/tomboy.ts
```diff
@@ -0,0 +1,134 @@
...
+				// Check for both Tomboy and Gnote on Linux
+				const homeDir = os.homedir();
+				const tomboyPath = path.join(homeDir, '.local', 'share', 'tomboy');
+				const gnotePath = path.join(homeDir, '.local', 'share', 'gnote');
+				
+				// Return whichever exists, preferring Gnote since it is the newer SW
+				try {
+					if (fsPromises && fsPromises.access) {
+						// Check if gnote directory exists
+						fsPromises.access(gnotePath).then(() => {}).catch(() => {});
```
tgrosinger: This should be awaiting the promise, otherwise the try catch will not prevent us from returning the `gnotePath` even if it doesn't exist.

Unfortunately that also mean the caller needs to await this function, which isn't possible from `init`. So it might be best to just default to which ever path is more common and put the other one in the description.

src/formats/tomboy.ts
```diff
@@ -0,0 +1,134 @@
...
+				toggle.setValue(this.todoEnabled)
+					  .onChange((value: boolean) => this.todoEnabled = value);
+			});
+
+		new Setting(this.modal.contentEl)
+			.setName('Keep title in markdown')
+			.setDesc('Choose whether to keep the note title in the markdown content. "Automatic" keeps titles only when special characters are lost in filename conversion.')
+			.addDropdown((dropdown: DropdownComponent) => {
+				dropdown.addOption('automatic', 'Automatic')
+					.addOption('yes', 'Yes')
```
tgrosinger: Maybe instead of `Yes` and `No`, the description for the options could have more detail, for example something like: 

- Automatic
- Keep titles
- Filename only

src/formats/tomboy.ts
```diff
@@ -0,0 +1,134 @@
...
+				ctx.reportFailed(file.fullpath, e);
+			}
+
+			ctx.reportProgress(i + 1, files.length);
+		}
+	}
+
+	private async processFile(ctx: ImportContext, folder: any, file: any): Promise<void> {
+		const xmlContent = await file.readText();
+		this.coreConverter.setTodoEnabled(this.todoEnabled);
```
tgrosinger: Looks like these settings only need to be set once. Maybe call this in `import` before processing the files.

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+	 */
+	private isFullyStrikethrough(sections: Array<ContentSection>): boolean {
+		return sections.every(section => section.xmlPath.includes('strikethrough'));
+	}
+
+	/**
+	 * Check if a note title contains TODO keywords
+	 */
+	private isTodoTitle(title: string): boolean {
+		const todoKeywords = ['TODO', 'Todo', 'To Do', 'todo', 'to do', 'ToDo'];
```
tgrosinger: It might be simpler to check with a regex. `/to ?do/i.matches`

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+	}
+
+	/**
+	 * Recursively extract content sections from XML elements
+	 */
+	private extractContentSections(element: Element, currentPath: string): Array<ContentSection> {
+		const sections: Array<ContentSection> = [];
+
+		// Convert NodeList to Array for iteration
+		Array.from(element.childNodes).forEach(child => {
```
tgrosinger: No need to call `Array.from`

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+
+		// Parse date elements
+		const createDateElement = doc.querySelector('create-date');
+		const lastChangeDateElement = doc.querySelector('last-change-date');
+		
+		let createDate: Date | undefined;
+		let lastChangeDate: Date | undefined;
+
+		if (createDateElement?.textContent) {
+			createDate = this.parseTomboyDate(createDateElement.textContent);
```
tgrosinger: If `parseTomboyDate` returned null if it was passed null, this section of code could be simplified nicely.

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+		}
+
+		const coreText = text.substring(leadingWhitespace.length, text.length - trailingWhitespace.length);
+		return { coreText, leadingWhitespace, trailingWhitespace };
+	}
+
+	/**
+	 * Convert structured content lines to Markdown
+	 */
+	private convertStructuredContent(contentLines: Array<ContentLine>, noteTitle: string): string {
```
tgrosinger: Doesn't look like `noteTitle` is used.

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+			// Check for headings (entire line analysis) - lists already handled above
+			let headingPrefix = '';
+			let isBoldConsumedInHeading = false;
+			if (line.contentSections.length > 0 && listPrefix === '') {
+				const paths = line.contentSections.map(section => section.xmlPath);
+				const allBold = paths.every(path => path.includes('bold'));
+				const allHuge = paths.every(path => path.includes('size:huge'));
+				const allLarge = paths.every(path => path.includes('size:large'));
+
+				if (allBold || allHuge || allLarge) {
```
tgrosinger, line 558: What's going on in this section? Does Tomboy use bold and size to denote the heading level?

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+			result += this.activeFormatStack.pop()!.closeTag;
+		}
+
+		return result;
+	}
+
+	/**
+	 * Calculate formatting changes between current and new format states using stack-based ordering
+	 */
+	private calculateFormattingChanges(currentFormats: Set<string>, newFormats: { [key: string]: boolean }) {
```
tgrosinger, line 440: Since this is only called in one place, I think it would be easier to reason about if it were moved in-line so that closing tags can modify the result directly. Let me know what you think.

src/formats/tomboy/core.ts
```diff
@@ -0,0 +1,605 @@
...
+				break;
+			case 'automatic':
+			default:
+				shouldKeepTitle = informationLost;
+				break;
+		}
+
+		if (shouldKeepTitle) {
+			// Keep the title - format it as H1 if it's the first line
+			const lines = markdownContent.split('\n');
```
tgrosinger: The line split and the escapedTitle can be pulled out of the if block since both cases use them.

tgrosinger (changes requested):
This looks really good, thanks @farnaby!

Please run your code through the linter as well as considering the comments below. Thanks.

farnaby:
Hi @tgrosinger thank you for the detailed review! I'll come back to you as soon as I find time to look into all of the suggestions.

src/formats/tomboy/core.ts
farnaby: Tomboy doesn't actually have any concept of heading levels, it just allows changing the font size and style. So this is a heuristic to deduce meaningful heading levels from the way I was using it (and I suspect most users will do).

src/formats/tomboy.ts
farnaby: Right, this doesn't work. I've simplified it to simply default to the tomboy path and show the gnote path in the description as suggested.

src/formats/tomboy.ts
farnaby: Makes sense.

src/formats/tomboy.ts
farnaby: Makes sense. I'll just put it in import as suggested.

src/formats/tomboy/core.ts
farnaby: I fully agree that the logic as it currently is can be put more concise with a regex.

My thought here was that others might want to extend the list of keywords, or ideally even make it configurable. Just looking for "TODO" in the headings is rather specific to my own usage - others might label their Todo-lists as "Tasks" or use a term in another language.

However, converted it to regexp for now.

src/formats/tomboy/core.ts
farnaby: I see, simplified.

src/formats/tomboy/core.ts
farnaby: Makes sense.

src/formats/tomboy/core.ts
farnaby: You're right, it isn't - removed it from the function signature.

src/formats/tomboy/core.ts
farnaby: Yes. Simplified this code section.

src/formats/tomboy/core.ts
farnaby: The formatting processing feels a bit entangled and I think that there are certainly ways to make it more elegant or consistent. However, to be honest I currently don't have the spare time or motivation to further work on it. It did its job for me, and I think it should do its job for others who migrate their notes from tomboy to obsidian as well, at least much better than the html migration path.
 
Whether or not the code becomes more readable by inlining said function - I'm pretty neutral about it.
For now, I left the code as it was, but feel free to refactor!

Issue #443, opened 2025-10-24
Importer plugin: Evernote failed imports


tsmcgrath:
I am moving years worth of notes from Evernote to Obsidian. Thank you for the importer plugin. It is awesome. However, it does choke on some notes and fails. All with the same error:
`Failed: "The splitting of Seattle — and the Eastside, too — is accelerating" because TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer...`

MacOS Sequoia 15.6.1
MacPro M1 (ARM)

- I've imported thousands of notes. Only 44 errors so far so it is infrequent. But, all the same errors.
- Could be related to issue #25 ?

3 failure examples exported separately as .ENEX files and attached zipped.

[EvernoteImportFailExamples.zip](https://github.com/user-attachments/files/23123398/EvernoteImportFailExamples.zip)

Examples from these original URL's. Web clipped with Evernote web clipper.

1. https://www.seattletimes.com/seattle-news/the-splitting-of-seattle-and-the-eastside-too-is-accelerating/?utm_medium=social&utm_campaign=owned_echobox_tw_m&utm_source=Twitter#Echobox=1694880733-1
2. https://blog.adafruit.com/2023/09/05/icymi-python-on-microcontrollers-newsletter-250-issues-displays-galore-cheat-sheets-and-much-more-circuitpython-python-micropython-icymi-raspberry_pi/
3. https://www.seattletimes.com/life/travel/road-trip-inspiration-nearly-400-miles-of-adventure-in-northeast-or/

Xheldon:
It can be reproduced consistently; I encountered two errors. One is that the note name is too long (e.g., `ICYMI_Python_on_Microcontrollers_Newsletter_250_Issues,_Displays_Galore,_Cheat_Sheets_and_much_mICYMI_Python_on_Microcontrollers_Newsletter_250_Issues,_Displays_Galore,_Cheat_Sheets_and_much_more!_CircuitPython_Python_micropython_ICYMI_@Raspberry_Pi.resources` — is this a note name?)
Another is that when processing attachments, the data is empty.

@tsmcgrath  I just submitted a PR; you can test using the code from this branch, see https://github.com/obsidianmd/obsidian-importer/pull/449

tsmcgrath:
Awesome @Xheldon. Thank you!

[...]

PR #449, opened 2025-11-12
Fix Evernote importer bug #443

Xheldon:
Fix two issues:
1. The import note path may be too long (including the total length of folder + filename), so it is truncated and the rename sequence number is incremented.
2. Notes may lack certain attributes; add fault‑tolerant handling.

### DIFF
--- a/src/formats/yarle/utils/filename-utils.ts
+++ b/src/formats/yarle/utils/filename-utils.ts
@@ -8,6 +8,10 @@ import { ResourceFileProperties } from '../models/ResourceFileProperties';
 import { escapeStringRegexp } from './escape-string-regexp';
 import { extensionForMime } from '../../../mime';
 
+// Filename length constants
+const MAX_NOTE_NAME_LENGTH = 100; // Limit note name length to prevent path issues
+const MAX_RESOURCE_FILENAME_PREFIX_LENGTH = 50; // Maximum length for resource filename prefix
+
 export const normalizeTitle = (title: string) => {
 	return sanitizeFileName(title).replace(/[\[\]\#\^]/g, '');
 };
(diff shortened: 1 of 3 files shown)


src/formats/yarle/utils/filename-utils.ts
```diff
@@ -105,11 +109,28 @@ export const getNoteName = (dstPath: string, note: any): string => {
 			zettelPrefix;
 
 		if (!yarleOptions.useZettelIdAsFilename) {
-			noteName += getFilePrefix(note) !== 'Untitled' ? `${separator}${getFilePrefix(note)}` : '';
+			const filePrefix = getFilePrefix(note);
```
tgrosinger, line 112: Looks like this `getFilePrefix` can be hoisted out of the if condition since both cases use it now.

src/formats/yarle/utils/filename-utils.ts
```diff
@@ -105,11 +109,28 @@ export const getNoteName = (dstPath: string, note: any): string => {
...
+			if (filePrefix !== 'Untitled') {
+				const combined = `${noteName}${separator}${filePrefix}`;
+				// Truncate if combined name is too long
+				if (combined.length > MAX_NOTE_NAME_LENGTH) {
```
tgrosinger, line 116: Maybe just rewrite this so that there is no else case required. Instead, calculate the available space and always add a substr from filePrefix of that length. I think that will allow simplifying this section a bit.

src/formats/yarle/utils/folder-utils.ts
```diff
@@ -21,10 +21,41 @@ export interface NotebookStackProps {
...
+const MAX_ENEX_DIR_LENGTH = 100; // Conservative limit for enex directory name
+
+/**
+ * Check if a path exists and add (1), (2), etc. suffix if needed to avoid conflicts
+ * @param basePath - The directory path where the item will be created
+ * @param name - The desired name (without path)
+ * @param suffix - Optional suffix to append (e.g., '.resources')
+ * @returns A unique name that doesn't conflict with existing items
+ */
+const getUniqueNameForPath = (basePath: string, name: string, suffix: string = ''): string => {
```
tgrosinger, line 35: I think I remember you adding something similar in the notion api PR? Maybe something like this should be a utility format that can be shared between both.

Xheldon: Yes, there's also similar tag‑generation logic. However, because I haven't worked with many other importer implementations, these are the only two functions that can be extracted globally, and I'll address this later.

tgrosinger: I did some looking into this and the use case is a bit different because of the suffix and how one returns the name part and the other returns the full path. For now I'm okay with keeping these separate.

tgrosinger:
I made a couple small changes to address my own feedback and then merged this. Thank you!

PR #454, opened 2025-11-24
Notion API Bugfix

Xheldon:
* Fix the display issue of the Database date attribute in the base
* Fix Page exist error
* Add some logs to pinpoint the issue; they’ll be removed once things are stable.

Note: I use `adapter.exists` for reliable check, to operate the types.json

### DIFF
--- a/src/formats/notion-api/database-helpers.ts
+++ b/src/formats/notion-api/database-helpers.ts
@@ -622,12 +622,12 @@ function sanitizePropertyKey(key: string): string {
  * Returns a mapping of property name -> Obsidian type
  */
 function extractPropertyTypesForTypesJson(
-	dataSourceProperties: any,
+	dataSourceProperties: Record<string, any>,
 	databasePages: Array<PageObjectResponse | PartialPageObjectResponse>
 ): Record<string, string> {
 	const propertyTypes: Record<string, string> = {};
 	
-	for (const [key, prop] of Object.entries(dataSourceProperties as Record<string, any>)) {
+	for (const [key, prop] of Object.entries(dataSourceProperties)) {
 		const propType = prop.type;
 		const propName = prop.name || key;
 		


tgrosinger:
I pushed a couple changes to the master branch which looks like it caused some conflicts. Please rebase your branch.

src/formats/notion-api/database-helpers.ts
```diff
@@ -731,6 +618,146 @@ function sanitizePropertyKey(key: string): string {
 	return key;
 }
 
+/**
+ * Extract property types that should be written to types.json
+ * Maps all Notion property types to Obsidian property types following the same logic as mapDatabaseProperties
+ * Returns a mapping of property name -> Obsidian type
+ */
+function extractPropertyTypesForTypesJson(
+	dataSourceProperties: any,
```
tgrosinger: Please don't use any. This can be `Record<string, any>`. Remove the `as Record<string, any>` from the for loop below.

src/formats/notion-api/database-helpers.ts
```diff
@@ -377,10 +374,17 @@ export async function createBaseFile(params: CreateBaseFileParams): Promise<stri
...
+	// Use adapter.exists for reliable check
+	if (await vault.adapter.exists(baseFilePath)) {
 		// Update existing .base file with latest database properties
-		await vault.modify(existingFile, baseContent);
+		const existingFile = vault.getAbstractFileByPath(baseFilePath);
+		if (existingFile instanceof TFile) {
+			await vault.modify(existingFile, baseContent);
+		}
+		else {
+			// File exists but not recognized as TFile, use adapter to write directly
```
tgrosinger, line 383: Is this an error that was being encountered? What steps reproduce the error?

tgrosinger (changes requested):
As Kepano mentioned, please update how you're reading and writing types.

```
// Read
app.metadataTypeManager.getAssignedWidget(key: string);
// Write
app.metadataTypeManager.setType(key: string, type: string);
```

src/formats/notion-api/database-helpers.ts
Xheldon: I changed this part because, in the `readTypesJson` function of the `src/formats/notion-api/vault-helpers.ts` file, the logic required repeatedly reading and writing the `types.json` file, and I encountered a `File already exist` error. Since I was using the `vault.getAbstractFileByPath` method (similar to the logic here) at the time, I felt that this method seemed to have caching and could not return the latest content in real time after being called. At the same time, I recalled that some users had reported (including a screenshot from @kepano in a Discord chat) encountering similar errors, so I examined all possible situations where this error could occur and changed them to the `valut.adapter.exists` method in similar places.

src/formats/notion-api/database-helpers.ts
Xheldon: Done. https://github.com/obsidianmd/obsidian-importer/pull/454/commits/b3b50876d3947312765350164a73f038cd863276

package-lock.json
```diff
@@ -969,6 +969,14 @@
 				"url": "https://opencollective.com/typescript-eslint"
 			}
 		},
+		"node_modules/@xmldom/xmldom": {
```
tgrosinger, line 972: Why is this being added? I don't see a change in package.json.

tgrosinger: Oh, looks like this is left over from the recent OneNote PR that was merged.

tgrosinger (approved):

tgrosinger:
Looks good to me. Thank you!

PR #459, opened 2025-11-27
Notion API: fix the toggle convert & file name error

Xheldon:
In Notion, if a Toggle's title is a Heading, then it is a Heading block, not Toggle🥲

### DIFF
--- a/src/formats/notion-api/block-converter.ts
+++ b/src/formats/notion-api/block-converter.ts
@@ -15,16 +15,13 @@ import {
 	LinkPreviewBlockObjectResponse,
 	CodeBlockObjectResponse,
 	RichTextItemResponse,
-	Heading1BlockObjectResponse,
-	Heading2BlockObjectResponse,
-	Heading3BlockObjectResponse
 } from '@notionhq/client';
 import { normalizePath, TFile } from 'obsidian';
 import { parseFilePath } from '../../filesystem';
 import { sanitizeFileName } from '../../util';
 import { getBlockChildren, processBlockChildren } from './api-helpers';
 import { downloadAndFormatAttachment, extractAttachmentFromBlock, getCaptionFromBlock } from './attachment-helpers';
-import { BlockConversionContext, AttachmentType, AttachmentBlockConfig } from './types';
+import { BlockConversionContext, AttachmentType, AttachmentBlockConfig, HeaderContentWithRichTextAndColorResponse } from './types';
 import { createPlaceholder, extractPlaceholderIds, PlaceholderType } from './utils';
 import { getUniqueFilePath } from './vault-helpers';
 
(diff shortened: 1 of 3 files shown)


Xheldon:
I made some changes to the `sanitizeFileName` function; it has a wide impact, but logically there aren’t any major issues. @tgrosinger

src/util.ts
```diff
@@ -8,15 +8,21 @@ let windowsTrailingRe = /[\. ]+$/;
...
-		.replace(windowsReservedRe, '')
 		.replace(windowsTrailingRe, '')
+		.replace(windowsReservedRe, '')
 		.replace(startsWithDotRe, '')
 		.replace(badLinkRe, '');
+	
+	// If the result is empty or only whitespace after sanitization, return a default name
+	// This prevents creating files like ".md" (no name) or folders with only spaces
+	const trimmed = sanitized.trim();
+	return trimmed || 'Untitled';
```
tgrosinger, line 25: I think generally this is a good change, however, in `importDatabaseCore` of the Notion API importer it would previously use `Untitled Database` if this function returned an empty value and now it will just return `Untitled`.

src/util.ts
```diff
@@ -8,15 +8,21 @@ let windowsTrailingRe = /[\. ]+$/;
...
 export function sanitizeFileName(name: string) {
-	return name
+	const sanitized = name
 		.replace(illegalRe, '')
-		.replace(controlRe, '')
+		.replace(controlRe, '')  
 		.replace(reservedRe, '')
-		.replace(windowsReservedRe, '')
 		.replace(windowsTrailingRe, '')
+		.replace(windowsReservedRe, '')
```
tgrosinger, line 18: Can you please make this change in `sanitizeFileNameKeepPath` as well to keep them consistent?

src/util.ts
```diff
@@ -8,15 +8,21 @@ let windowsTrailingRe = /[\. ]+$/;
 let startsWithDotRe = /^\./; // Regular expression to match filenames starting with "."
 let badLinkRe = /[\[\]#|^]/g; // Regular expression to match characters that interferes with links: [ ] # | ^
 
+// I think we should first remove illegal characters such as spaces and periods, and then check whether it is a Windows reserved word.
 export function sanitizeFileName(name: string) {
-	return name
+	const sanitized = name
 		.replace(illegalRe, '')
-		.replace(controlRe, '')
+		.replace(controlRe, '')
```
tgrosinger: Extra space added.

src/util.ts
tgrosinger: Oh you already did this. Thanks!

src/formats/notion-api/block-converter.ts
```diff
@@ -458,25 +501,50 @@ export function convertParagraph(block: BlockObjectResponse, context?: BlockConv
...
  */
-export function convertHeading(block: BlockObjectResponse, context?: BlockConversionContext): string {
+export async function convertHeading(block: BlockObjectResponse, context: BlockConversionContext): Promise<string> {
 	// Get indent level if heading is nested in a list
 	const indentLevel = context?.indentLevel || 0;
 	const indent = '    '.repeat(indentLevel); // 4 spaces per indent level
 	
-	let headingText = '';
+	let headingPrefix = '';
+	let headingData: Heading1BlockObjectResponse['heading_1'] | Heading2BlockObjectResponse['heading_2'] | Heading3BlockObjectResponse['heading_3'];
```
tgrosinger:
```suggestion
	let headingData: HeaderContentWithRichTextAndColorResponse;
```

src/formats/notion-api/block-converter.ts
```diff
@@ -458,25 +501,50 @@ export function convertParagraph(block: BlockObjectResponse, context?: BlockConv
...
+	let isToggleable = false;
```
tgrosinger: `isToggleable` can be defined after the if-block below without any casting required.

```ts
const isToggleable = headingData.is_toggleable || false;
```

src/util.ts
Xheldon: I run `npm run lint` before each commit, but occasional spacing and comment indentation issues do occur, possibly due to my editor configuration; I’ll take a closer look.

src/util.ts
Xheldon: Indeed, but compared to checking the return value at every `sanitizeFileName` function call, I don’t think it’s that important. Do you think this part must, as before, return a possible empty string value for backward compatibility?

src/formats/notion-api/block-converter.ts
Xheldon: Yes, that part was overly verbose, so I refactored it, along with the `headingData` above, see: [8496eff](https://github.com/obsidianmd/obsidian-importer/pull/459/commits/8496effe607b9a818f4af803dee0129f029def47)

src/formats/notion-api/block-converter.ts
Xheldon: Done. see: [8496eff](https://github.com/obsidianmd/obsidian-importer/pull/459/commits/8496effe607b9a818f4af803dee0129f029def47)

src/util.ts
Xheldon: done.

src/util.ts
tgrosinger: No, I don't think that's critical. I looked through the other uses and the one I mentioned is the only one that even acted on the return value being empty. This should be just fine I think.

Issue #455, opened 2025-11-24
Airtable importer


kepano:
Import Airtable as bases and files, similar to Notion API importer and CSV importer.

PR #461, opened 2025-11-28
Support Airtable API

Xheldon:
issue https://github.com/obsidianmd/obsidian-importer/issues/455

This PR’s initial submission only implements the basic functionality; formula, Rollup, Lookup, and other related properties have not been properly handled yet. These features will be fixed in later code commits.

If anyone is willing to try this PR, please be sure to back up your Vault.

Some of the code was generated by AI. I have reviewed and tested it myself, but there are parts that need refactoring. I will continue to optimize the code in subsequent commits and will request the author’s review when the timing feels right.


tgrosinger:
I see Kepano added me as a reviewer, but based on the description I will hold off until I hear from Xheldon saying that it's ready. Looking forward to trying it out!

Xheldon:
I think I’ve finally finished this task, the code is ready for review! @tgrosinger 

I tested importing the Base that @kepano shared with me, which contains multiple tables with nearly 3,000 records, and it runs perfectly😎

Xheldon:
It looks like, due to my mistake, I mixed in dirty history. After cherry-picking all necessary changes, I resubmitted a new version.

src/util.ts
```diff
@@ -25,6 +25,38 @@ export function sanitizeFileName(name: string) {
...
+/**
+ * Get a unique file path by appending 1, 2, etc. if needed
+ * Uses the same naming convention as Obsidian's attachment deduplication (space + number)
+ * 
+ * @param vault - Obsidian vault instance
+ * @param parentPath - Parent folder path
+ * @param fileName - File name with extension (e.g., "note.md")
+ * @returns Unique file path that doesn't conflict with existing files
+ */
+export function getUniqueFilePath(vault: Vault, parentPath: string, fileName: string): string {
```
tgrosinger, line 39: Please remove `notion-api/vault-helpers.ts:getUniqueFilePath` if this is being added here instead.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+				.then(textComponent => {
+					textComponent.inputEl.type = 'password';
+				}));
+
+		// Load bases and tables button
+		const loadSetting = new Setting(this.modal.contentEl)
+			.setName('Select tables to import')
+			.setDesc('Click "Load" to browse your Airtable bases and tables.');
+
+		let toggleButtonRef: any = null;  // ButtonComponent from obsidian
```
tgrosinger: These two variables should use `ButtonComponent` from the Obsidian API.

```suggestion
		let toggleButtonRef: ButtonComponent | null = null;
```

Same for the `loadButtonRef`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+
+		// Load button
+		loadSetting.addButton(button => {
+			loadButtonRef = button;
+			button
+				.setButtonText('Load')
+				.onClick(async () => {
+					try {
+						this.loadButton = loadButtonRef;
+						this.toggleSelectButton = toggleButtonRef;
```
tgrosinger: Why are two buttons being assigned to `toggleSelectButton`? It seems like this one probably shouldn't be?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+		let loadButtonRef: any = null;  // ButtonComponent from obsidian
+
+		// Toggle select all/none button
+		loadSetting.addButton(button => {
+			toggleButtonRef = button;
+			button
+				.setButtonText('Select all')
+				.onClick(() => {
+					this.toggleSelectButton = toggleButtonRef;
+					this.handleToggleSelectClick();
```
tgrosinger: This function is only being called from this location. Please inline it's contents here, and remove `toggleSelectButton` from the class. I think `toggleButtonRef` can be removed as well.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+		});
+
+		// Load button
+		loadSetting.addButton(button => {
+			loadButtonRef = button;
+			button
+				.setButtonText('Load')
+				.onClick(async () => {
+					try {
+						this.loadButton = loadButtonRef;
```
tgrosinger: I don't think `loadButtonRef` is necessary. You can assign using `button` here directly. However, I think this assignment should also be moved out of the onClick handler, so `loadButtonRef = button` can be changed to `this.loadButton = button`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+		});
+
+		// Page tree container (using Publish plugin's style with proper hierarchy)
+		// Create the section wrapper
+		const publishSection = this.modal.contentEl.createDiv();
+		publishSection.addClass('file-tree', 'publish-section');
+
+		// Create the change list container
+		this.treeContainer = publishSection.createDiv('publish-change-list');
+		this.treeContainer.style.maxHeight = '200px';
```
tgrosinger: Why are these styles defined like this instead of adding them to styles.css?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+					.setValue(true)
+					.onChange(value => {
+						this.downloadAttachments = value;
+					});
+			});
+
+		// View property name
+		new Setting(this.modal.contentEl)
+			.setName('View property name')
+			.setDesc('Property name to track which views a record belongs to. Each record will have a list of view names it appears in.')
```
tgrosinger, line 237: It took me a few reads to understand what this was saying. What do you think about something like this:

> Obsidian property name for which will contain a list of view names each record appears in.

@kepano might have a better suggestion.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2124 @@
...
+
+		if (!this.loadButton) {
+			return;
+		}
+
+		this.loadButton.setDisabled(true);
+		this.loadButton.setButtonText('Loading...');
+
+		try {
+			const tempCtx = {
```
tgrosinger: This doesn't appear to be a valid ImportContext. Maybe the types requested by `fetchBases` and `fetchTableSchema` should be adjusted so that this narrower set of properties is valid.

We should avoid type coercions like `as unknown as ___`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				this.loadButton.setDisabled(false);
+				this.loadButton.setButtonText('Refresh');
+			}
+		}
+	}
+
+	/**
+	 * Render tree UI
+	 */
+	private renderTree(): void {
```
tgrosinger, line 361: OneNote, Notion API , and now Airtable all have a tree rendering of content which can be selected for import. What do you think about trying to abstract this logic and reuse it across those three importers?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				treeItem.addClass('is-collapsed');
+			}
+			
+			// Allow arrow click even when disabled
+			if (node.disabled) {
+				collapseIcon.style.pointerEvents = 'auto';
+			}
+
+			// Store references for event handler
+			const treeItemRef = treeItem;
```
tgrosinger: I don't think `treeItemRef` is necessary. It looks like you can use `treeItem` in all the places where `treeItemRef` is currently used.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				node.collapsed = !node.collapsed;
+
+				// Get reference if not set yet
+				if (!childrenContainer) {
+					childrenContainer = treeItemRef.querySelector('.tree-item-children') as HTMLElement;
+				}
+
+				// Toggle CSS classes and visibility
+				if (node.collapsed) {
+					collapseIcon.addClass('is-collapsed');
```
tgrosinger: This can be simplified a bit with `toggleClass`.

```
collapseIcon.toggleClass('is-collapsed', node.collapsed);
treeItem.toggleClass('is-collapsed', node.collapsed);
childrenContainer.toggle(!node.collapsed);
```

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+
+		// Inner content (checkbox, icon, title)
+		const treeItemInner = treeItemSelf.createDiv('tree-item-inner file-tree-item');
+
+		// Checkbox
+		const checkbox = treeItemInner.createEl('input', {
+			type: 'checkbox',
+			cls: 'file-tree-item-checkbox'
+		});
+		checkbox.checked = node.selected;
```
tgrosinger, line 458: You can move these into `createEl`.

```
		const checkbox = treeItemInner.createEl('input', {
			type: 'checkbox',
			cls: 'file-tree-item-checkbox',
			attr: {
				checked: node.selected,
				disabled: node.disabled
			}
		});
```

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		// Title
+		const titleEl = treeItemInner.createDiv('file-tree-item-title');
+		titleEl.setText(node.title);
+
+		// Children container
+		const childrenContainer = treeItem.createDiv('tree-item-children');
+
+		// Hide children container if collapsed
+		if (node.collapsed) {
+			childrenContainer.style.display = 'none';
```
tgrosinger:
```
childrenContainer.hide();
```

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			child.selected = selected;
+			child.disabled = selected;
+			this.setChildrenSelection(child, selected);
+		}
+	}
+
+	/**
+	 * Check if all nodes are selected
+	 */
+	private areAllNodesSelected(): boolean {
```
tgrosinger: Remove the inner function, give this function the argument for nodes with a default vault of `this.tree`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			for (const child of node.children!) {
+				this.renderTreeNode(childrenContainer, child, level + 1);
+			}
+		}
+	}
+
+	/**
+	 * Toggle node selection
+	 */
+	private toggleNodeSelection(node: AirtableTreeNode, selected: boolean): void {
```
tgrosinger: I'd suggest the name `setNodeSelection`, since it's not actually blindly toggling.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+
+		return checkNode(this.tree);
+	}
+
+	/**
+	 * Select or deselect all nodes
+	 */
+	private selectAllNodes(selected: boolean): void {
+		const processNode = (node: AirtableTreeNode) => {
+			if (!node.disabled) {
```
tgrosinger, line 528: Is there a situation where a node will be disabled, but not the children? Maybe this should return here.

Separately, I'm also wondering if this should still make necessary changes to make sure all children are selected, even if disabled. But it makes sense to not deselect if disabled.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+						if (!allFieldsMap.has(field.name)) {
+							allFieldsMap.set(field.name, field);
+							fieldExamples.set(field.name, this.generateExampleValue(field));
+						}
+					}
+				}
+				
+				// Recursively collect from children (for base nodes)
+				if (node.children && node.children.length > 0) {
+					collectFields(node.children);
```
tgrosinger, line 598: `collectNodes` also recursively gathers children. Are we going through all the children twice because of this?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+					});
+				}
+			}
+			else if (node.type === 'table' && node.metadata?.baseId) {
+				// Single table selected - find or create its base group
+				const baseId = node.metadata.baseId;
+				
+				// Find the base node to get the base name
+				let baseName = '';
+				for (const baseNode of this.tree) {
```
tgrosinger: Move this inside the if-block below.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			baseName: string;
+			tables: Array<{
+				tableName: string;
+				primaryFieldId: string;
+				fields: AirtableFieldSchema[];
+				views: AirtableViewInfo[];
+			}>;
+		}>();
+
+		for (const node of selectedNodes) {
```
tgrosinger, line 861: Similar to another comment I left, `selectedNodes` is going to have both the bases and their children since it gathers nodes recursively. If a base is selected it seems like we're going to add it and all of the children, and then as we continue iterating through `selectedNodes`, we'll attempt to add all the children again. Am I reading that right?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				if (ctx.isCancelled()) {
+					ctx.status('Import cancelled.');
+					return;
+				}
+
+				// ============================================================
+				// PHASE 2: Create files for this base
+				// ============================================================
+				try {
+					await this.createFilesForBase(ctx, folder.path);
```
tgrosinger, line 825: Is `folder.path` the same as `this.outputLocation`? Maybe just use that instead.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			new Notice(`Import failed: ${error.message}`);
+		}
+	}
+
+	/**
+	 * Group selected nodes by their base
+	 * Returns a Map where key is baseName and value contains baseId, baseName, and tables array
+	 */
+	private groupSelectedNodesByBase(selectedNodes: AirtableTreeNode[]): Map<string, {
+		baseId: string;
```
tgrosinger: This should be turned into a type interface since it's used in several locations, such as `fetchBaseData`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		}
+
+		return baseGroups;
+	}
+
+	/**
+	 * Clear data from previous base to free memory
+	 * Note: Some data is preserved across bases:
+	 * - recordIdToPath: needed for cleanupAirtableIds at the end
+	 * - allFieldsForTypeInference: needed for updatePropertyTypes at the end
```
tgrosinger, line 913: So if I have two bases with the same field name of different types, the first one will win?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		// Find primary field - this is used as note title/filename, not as a formula column
+		const primaryField = fields.find(f => f.id === primaryFieldId);
+		const primaryFieldName = primaryField?.name || null;
+		
+		// Process fields in original order, tracking which are formulas
+		// This preserves Airtable's field order in the .base file
+		const formulas: Map<string, string> = new Map(); // field name -> obsidian formula
+		
+		// Helper to resolve field name from ID
+		const resolveFieldName = (fieldId: string): string | undefined => {
```
tgrosinger: It doesn't seem like this helper adds anything over calling the `get` method directly.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			return this.globalFieldIdToNameMap.get(fieldId);
+		};
+		
+		// Helper to build map expression: note["LinkedField"].map(value.asFile().properties["TargetField"])
+		const buildMapExpression = (linkedFieldName: string, targetFieldName: string): string => {
```
tgrosinger: This is only used once, so let's just in-line it.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		
+		// Add fields in original order (excluding primary field which is file.name)
+		for (const field of fields) {
+			// Skip the primary field (it's represented by file.name)
+			if (field.id === primaryFieldId) {
+				continue;
+			}
+			
+			// Check if this field was converted to a formula
+			if (formulas.has(field.name)) {
```
tgrosinger: Not hugely important in this case, but I generally prefer to try to factor out the commonalities between logic branches so that it's easier to see what will always be true. For example, this could instead be written something like this, which makes it easy to see that it will always sanitize the name, and it will always push to propertyColumns.

```
const sanitizedName = this.sanitizePropertyName(field.name);
const columnName = formulas.has(field.name) ? `formula.${sanitizedName}` : sanitizedName;
propertyColumns.push(columnName);
```

The same thing could be done below when adding to `baseConfig.properties`.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		}
+		if (failedCount > 0) {
+			console.warn(`Failed to clean airtable-id from ${failedCount} file(s)`);
+		}
+	}
+
+	/**
+	 * Create a single .base file for the table with multiple views
+	 */
+	private async createViewBaseFiles(ctx: BaseFileContext): Promise<void> {
```
tgrosinger: I find this name a little misleading since it's only creating one file. Can it be called simply `createBaseFile`, or maybe `createBaseFileForTable`?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			// Handle other objects (shouldn't happen for primary fields, but just in case)
+			if (typeof value === 'object') {
+				return JSON.stringify(value);
+			}
+			return String(value);
+		};
+		
+		// Get primary field value (processed)
+		// Airtable always uses each table's primary field as note title
+		const primaryFieldValue = extractStringValue(recordFields[primaryFieldName]);
```
tgrosinger: `primaryFieldValue` is not used again. Just declare `title` here.
`const title = extractStringValue(...)`

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+		if (!title || title.trim() === '') {
+			title = 'Untitled Record';
+		}
+		
+		let sanitizedTitle = sanitizeFileName(title);
+		
+		let filePath = normalizePath(`${tablePath}/${sanitizedTitle}.md`);
+		
+		// Check for incremental import - skip if same record already exists
+		const shouldSkip = await this.handleIncrementalImportCheck(filePath, recordId, async () => {
```
tgrosinger: We're doing things in two places if there is a match:
1. In the async callback
2. In the if-block below.

Remove the callback from this check, and perform the progress reporting in the if-block below before returning.

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+			});
+			
+			// If formula was converted (returns null), use the computed value for templates
+			// Formula fields won't be in YAML (they'll be in .base file), but templates may reference them
+			if (convertedValue === null && field.type === 'formula') {
+				convertedValue = fieldValue;
+			}
+			
+			// Resolve linked records for template data
+			if (field.type === 'multipleRecordLinks' && Array.isArray(fieldValue)) {
```
tgrosinger: Should this be before the call to `convertFieldValue` since it does not require the result?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				const fieldValue = recordFields[field.name];
+
+				// Skip null/undefined
+				if (fieldValue === null || fieldValue === undefined) {
+					continue;
+				}
+
+				// Convert field value with direct linked record resolution
+				// Use globalFieldIdToNameMap for lookup/rollup fields that reference other tables
+				let convertedValue = convertFieldValue({
```
tgrosinger, line 1359: Can calling this twice be avoided? Maybe it can be retrieved from the `templateData`?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+				});
+
+				// Skip if convertedValue is null/undefined/empty string (e.g., formula was converted and will be in .base file)
+				if (convertedValue === null || convertedValue === undefined || convertedValue === '') {
+					continue;
+				}
+
+				// Resolve linked records IMMEDIATELY from recordIdToTitle
+				if (field.type === 'multipleRecordLinks' && Array.isArray(fieldValue)) {
+					convertedValue = fieldValue.map((linkedRecordId: string) => {
```
tgrosinger: This block is also duplicated. Is there a way this section can be deduplicated a bit?

src/formats/airtable-api.ts
```diff
@@ -0,0 +1,2148 @@
...
+						console.warn(`[Airtable] Skipping complex object for property "${propertyName}"`);
+						continue;
+					}
+					frontMatter[propertyName] = propertyValue;
+				}
+			}
+		}
+
+		// Apply body template
+		const bodyContent = this.templateConfig
```
tgrosinger: Unless the code is updated to make use of `templateData` when building the frontmatter, this should avoid building templateData if there is no templateConfig or if the bodyTemplate is empty.

Since the primary use-case of importing form Airtable is to put the data in the frontmatter, I think it will be fairly common for the bodyTemplate to be empty.

tgrosinger (commented):
Finished reading through the main logic, but I still need to review the helper/converter files.

Most of my comments are minor code style. I have not tested it yet, but the logic seems solid.

Looking really great so far. Thank you @Xheldon!

src/formats/airtable-api/types.ts
```diff
@@ -0,0 +1,263 @@
...
+	recordId: string; // Source record ID
+	fieldName: string; // Field name in frontmatter
+	linkedRecordIds: string[]; // Linked record IDs
+	linkedTableId?: string; // Target table ID
+}
+
+/**
+ * Lookup placeholder for post-processing
+ */
+export interface LookupPlaceholder {
```
tgrosinger: `RecordProcessingContext`, `LookupPlaceholder`, `RollupPlaceholder`, and `CreateBaseFileParams` appear to be unused.

src/formats/airtable-api/attachment-helpers.ts
```diff
@@ -0,0 +1,224 @@
...
+		
+		// Create the file
+		const normalizedPath = normalizePath(targetPath);
+		await vault.createBinary(normalizedPath, response.arrayBuffer);
+		
+		// Return path without extension for wiki links
+		const pathWithoutExt = targetPath.replace(/\.[^/.]+$/, '');
+		
+		return {
+			path: pathWithoutExt,
```
tgrosinger: This doesn't feel like the right data to be returning because there are two instances below where the extension is extracted from the filename and then appended to the end of this pathWithoutExt.

Maybe just return the full path? Then the users of this data can call the various filepath utilities to get the exact format required.

src/formats/airtable-api/formula-converter.ts
```diff
@@ -0,0 +1,786 @@
...
+ *   Example: array.filter(value > 2) NOT array.filter(v => v > 2)
+ * - Obsidian only supports one regex function: /pattern/.matches(string)
+ *   REGEX_EXTRACT and REGEX_REPLACE are NOT supported
+ * - ERROR() and BLANK() are converted to literal values:
+ *   ERROR() -> "!ERROR" (string literal to indicate error)
+ *   BLANK() -> "" (empty string)
+ * - String search functions FIND and SEARCH are NOT supported (no indexOf method)
+ * - SUBSTITUTE with 4 arguments (index parameter) is NOT supported, only 3-arg version works
+ * 
+ * Obsidian global functions (only 7):
```
tgrosinger, line 34: I'm not sure if they are needed here, but there are more global functions in Obsidian:

- random
- list
- link
- duration
- image
- icon
- file
- html
- escapeHTML

src/formats/airtable-api/formula-converter.ts
```diff
@@ -0,0 +1,786 @@
...
+ * @param airtableFormula - The Airtable formula expression
+ * @param fieldIdToNameMap - Map of field IDs to field names (for resolving field references)
+ */
+export function convertAirtableFormulaToObsidian(
+	airtableFormula: string,
+	fieldIdToNameMap?: Map<string, string>
+): string | null {
+	let result = airtableFormula;
+	
+	if (!canConvertFormula(result)) {
```
tgrosinger, line 249: It looks like all the places that call `convertAirtableFormulaToObsidian` already check `canConvertFormula`. Is there a way we can avoid doing that work twice? Maybe they don't need to check first and can just treat null from this function as unable to convert.

src/formats/airtable-api/formula-converter.ts
```diff
@@ -0,0 +1,786 @@
...
+	let changed = true;
+	let maxIterations = 20;
+	let iterations = 0;
+	
+	while (changed && iterations < maxIterations) {
+		changed = false;
+		iterations++;
+		
+		// Find function calls (case-insensitive)
+		const funcNamePattern = /(?<![.\w])([A-Z_][A-Z0-9_]*)\s*\(/gi;
```
tgrosinger: Pull this out into a const so it's not recompiled every iteration.

src/formats/airtable-api/formula-converter.ts
```diff
@@ -0,0 +1,786 @@
...
+				inString = false;
+			}
+		}
+		else {
+			if (char === '"' || char === '\'') {
+				inString = true;
+				stringChar = char;
+				current += char;
+			}
+			else if (char === '(' || char === '[') {
```
tgrosinger, line 728: Why are square brackets considered here?

src/formats/airtable-api/formula-converter.ts
```diff
@@ -0,0 +1,786 @@
...
+
+/**
+ * Find the matching closing parenthesis for an opening parenthesis
+ */
+function findMatchingParen(str: string, openPos: number): number {
+	let depth = 1;
+	let inString = false;
+	let stringChar = '';
+	
+	for (let i = openPos + 1; i < str.length; i++) {
```
tgrosinger, line 761: Not strictly necessary, but if you wanted you might be able to abstract some of this functionality into a helper function which walks along a string, accumulating characters and performing an action unless in quotes.

src/formats/airtable-api.ts
Xheldon: These two button issues, including the logic below. I previously ran into a problem with the Notion API where the `this` pointer was lost, so I added a double safeguard here. I’ll try using only a local reference to see if it works.

src/formats/airtable-api.ts
Xheldon: `this.handleToggleSelectClick` is indeed used only once and should be placed in the callback of this `onClick`, but the assignment of `this.toggleSelectButton` cannot be removed because that value is used in the `Load` button callback (`renderTree/loadTree`). I need to access that value across functions (by default the button is hidden, and it only appears after clicking the `Load` button).

The reason for adding a `toggleButtonRef` is that I encountered a case with the Notion API earlier and asked about it on Discord; the screenshot is shown below：

CleanShot_2025_12_10_at_12_04_56@2x

So my solution in the Notion API was also aimed at correctly setting `this.toggleSelectButton`, using a local variable within the closure; to access the component across functions, I used a reference to `this.toggleSelectButton`.

I wonder if my explanation is clear to you?

src/formats/airtable-api.ts
Xheldon: Reason same as above.

src/formats/airtable-api.ts
Xheldon: Yes, this is a lazy move... I directly copied the changes Kepano made to my code in the Notion API. The proper way would be to put it in `style.css`. I’ll handle this part of the Notion API styles together:

CleanShot_2025_12_10_at_12_13_50@2x

src/formats/airtable-api.ts
Xheldon: The copy in this spot can finally be reviewed by @kepano for the code—--just like last time.

src/formats/airtable-api.ts
Xheldon: That's a good idea; I can start working on it. `id`, `name`, `children`, and other tree nodes are generic. However, do you think the expand/collapse logic should also be extracted as generic logic? 
I'm fairly familiar with the Notion API and Airtable and can handle it, but OneNote might require a bit more effort since I'm not very familiar with the OneNote node structure.

src/formats/airtable-api.ts
Xheldon: Yes, actually the ctx here only uses the `status` function; I implemented a `StatusReporter` type specifically for this.

src/formats/airtable-api.ts
Xheldon: Alright, modify it together with the Notion API.

src/formats/airtable-api.ts
Xheldon: I merged `toggleNodeSelection` and `setNodeChildrenSelection`

src/formats/airtable-api.ts
Xheldon: Here I kept the same logic as the Notion API: once a parent is selected, its child is automatically selected and cannot be deselected. I recommend leaving this logic unchanged, because Airtable has a Linked Records column type (similar to Notion’s Relation), and if you don’t import every table in a base, the table containing a referenced column might not be found.

src/formats/airtable-api.ts
Xheldon: No, because Airtable's directory tree hierarchy has only two levels; in fact, only the root node Base reaches this point to iterate over `table children`, and this won't be repeated.

src/formats/airtable-api.ts
Xheldon: This situation will not occur. First, only Base nodes have a children property; second, the if branch here distinguishes node types to treat them differently.

```// Base node
const baseNode: AirtableTreeNode = {
	id: base.id,
	title: base.name,
	type: 'base',
	parentId: null,
	children: [], // <-------- Children
	selected: false,
	disabled: false,
	collapsed: true,
};

// Table node
const tableNode: AirtableTreeNode = {
	id: `${base.id}:${table.id}`,
	title: table.name,
	type: 'table',
	parentId: base.id,
	selected: false,
	disabled: false,
	metadata: {
		baseId: base.id,
		tableName: table.name,
		primaryFieldId: table.primaryFieldId,
		fields: table.fields,
		views: table.views,
	},
};
```

src/formats/airtable-api.ts
Xheldon: The information for `folder` comes from `const folder = await this.getOutputFolder()`, and `getOutputFolder()` handles the case where `this.outputLocation` is empty (in which case `folder.path` is `/`).

src/formats/airtable-api.ts
Xheldon: Yes, do you have a better way?

src/formats/airtable-api.ts
Xheldon: The function call method will be a bit more concise... However, you're right—there are only two calls where this isn’t necessary, and I’ll change them to direct calls.

src/formats/airtable-api.ts
Xheldon: OK, I'll rewrite it concisely using the ternary operator.

src/formats/airtable-api.ts
Xheldon: Yes, you're right, I changed it to `createBaseFile`.

src/formats/airtable-api.ts
Xheldon: Yeah, It would be more reasonable to place the `convertFieldValue` call after the if branch.

src/formats/airtable-api.ts
Xheldon: Yes, I added a cache and will use it directly from now on.

src/formats/airtable-api.ts
Xheldon: It has already been handled in the logic changes above, adding cache usage.

src/formats/airtable-api/types.ts
Xheldon: Yes, I tried using AI (CodeX) and also encountered this issue, and have already updated.

src/formats/airtable-api.ts
Xheldon: Exactly right, I'll tweak this part of the logic later; we should check whether `this.templateConfig.bodyTemplate` exists beforehand to avoid executing unnecessary code.

src/formats/airtable-api/attachment-helpers.ts
Xheldon: Indeed, your proposal is more reasonable.

src/formats/airtable-api/formula-converter.ts
Xheldon: I understand. I’ve gone through the Airtable and Obsidian documentation piece by piece (including the List methods you previously sent me that are about to be implemented, such as max/min), and all the global functions you mentioned have been taken into account. The comment here indicates the Obsidian global functions that will be available after the Airtable conversion.

src/formats/airtable-api/formula-converter.ts
Xheldon: Array literal: SUM([1, 2, 3], 4) – commas inside the square brackets are not argument separators  
Transformed field reference: note["Field Name"] – although there’s no comma here, you still need to correctly track bracket depth

src/formats/airtable-api/formula-converter.ts
Xheldon: This logic is currently only called in two places, so we can keep things as they are for now.

src/util.ts
Xheldon: Done: [eb45b2d](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/eb45b2d82606e28ca10d4fccc618fffee0d1abcf)

src/formats/airtable-api.ts
Xheldon: Done: [c37d90c](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/c37d90cadf051dc420fe537b7fbad593e53667c2)

src/formats/airtable-api.ts
Xheldon: Done: [c37d90c](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/c37d90cadf051dc420fe537b7fbad593e53667c2)

src/formats/airtable-api.ts
Xheldon: Done: [5c596da](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/5c596da48e1cd2a17d80887f86f3fe5cc8a12379)

src/formats/airtable-api.ts
Xheldon: And right now this logic only appears in two places; maybe I’ll wait until later—like when handling the Craft Docs API—when extraction would be more valuable.

src/formats/airtable-api.ts
Xheldon: Done: [417fe69](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/417fe692e07c589e0fb5cb66ff010f81567d348b)

src/formats/airtable-api.ts
Xheldon: Done: [417fe69](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/417fe692e07c589e0fb5cb66ff010f81567d348b)

src/formats/airtable-api.ts
Xheldon: Done: [680a8e5](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/680a8e502402719bd7df4ecde6d0f6da00b38144)

src/formats/airtable-api.ts
Xheldon: Done: [c1b5dac](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/c1b5dac3e8e24089de0d9f293bccf21d0e4d3628)

src/formats/airtable-api.ts
Xheldon: Done: [1a265b2](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/1a265b290641927306db8f22291d49ea8f143e1b)

src/formats/airtable-api.ts
Xheldon: Done: [9d36378](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/9d36378902710cfc30e161a65e555ad889404e79)

src/formats/airtable-api.ts
Xheldon: Refactor, Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Right, done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Refactor, done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Refactor, done: [8770b9f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/8770b9f8b33e800aca6e5225fb0d99240553af26)

src/formats/airtable-api.ts
Xheldon: Changed as above

src/formats/airtable-api.ts
Xheldon: Done: [fd63701](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/fd637013b52c6c8a4cdeb161ff8ccbc0ebd233e8)

src/formats/airtable-api/attachment-helpers.ts
Xheldon: Done: [229e045](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/229e045b75cd93c0bbbb43675d501220fdd37389)

src/formats/airtable-api/formula-converter.ts
Xheldon: Done: [6e98db9](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/6e98db9e946dc70e6723b5cd053ef66f24c680d2)

src/formats/airtable-api/formula-converter.ts
Xheldon: Done: [73bfd3f](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/73bfd3fb9ce90a7b22a7c3bd109acf5a86c8d0a4)

src/formats/airtable-api.ts
Xheldon: That's not correct; we made a mistake. As long as the attr attribute exists, it is treated as true and checked, even if it's `checked="false"`, so it still needs to be set manually.
Recover: [bf0057e](https://github.com/obsidianmd/obsidian-importer/pull/461/commits/bf0057e2d13259de35a49a757ed45c363b46cbc1)

src/formats/airtable-api.ts
tgrosinger: Thank you!

src/formats/airtable-api.ts
tgrosinger: Ah good catch. Thanks for letting me know.

src/formats/airtable-api.ts
tgrosinger: Sounds good, thanks.

src/formats/airtable-api/formula-converter.ts
tgrosinger: Got it, thank you.

src/formats/airtable-api.ts
tgrosinger: No, I don't think there's a better approach given the global nature of property types in Obsidian. Just making sure I understood correctly.

tgrosinger (commented):
@Xheldon, I pushed a commit to your branch for the `toggleButtonRef` and `loadButtonRef`. I didn't have any problems with it losing the reference using this method of assignment. Please let me know what you think and if you find a situation where this doesn't work.

src/formats/airtable-api.ts
tgrosinger: I pushed a commit which changes this. Let me know what you think.

tgrosinger:
Nice, looks like pretty much all of the feedback has been addressed. I'm going to do a little testing, and hopefully we can get this merged shortly!

tgrosinger:
- [ ] Reminder for us to add a page to the help docs when this is released.

tgrosinger:
Some thoughts from testing:

- Can you see the current Group By and preserve that in the import?
- Same for the sort? Bases support sorting by multiple properties like Airtable.
- In the gallery view, can you set the image property so that it matches what is used in Airtable?
- In the gallery view, can you set the properties so that it matches what is shown in Airtable? It may be different from what is shown in the grid/table view.
- In the grid view, can you match the column summaries?
- There's something wrong with Rollups and Linked properties I think. Repro steps:
  - Create a database with two tables.
  - In the first table, create a link to the second table. Allow selecting multiple values.
  - In the first table, create a lookup of a numerical column from the second table.
  - In the second table, add some rows, being sure to put some values in the numerical column from the lookup.
  - In the first table, add some values to the link column. Be sure to add more than one value in a couple of the cells.
  - In the first table create a rollup with a formula like `SUM(values)` on the numerical column from the second table.
  - Observed: The lookup column is not a list data type. Just showing the two strings concatenated.
  - Observed: The rollup column is not performing the sum operation. Also shows the concatenated string value.

Xheldon:
@tgrosinger 

> Can you see the current Group By and preserve that in the import?
Same for the sort? Bases support sorting by multiple properties like Airtable.
In the gallery view, can you set the image property so that it matches what is used in Airtable?
In the gallery view, can you set the properties so that it matches what is shown in Airtable? It may be different from what is shown in the grid/table view.

NO. What you’re referring to are all view‑level details—field order, sorting, grouping, and even the cover settings for Gallery views. The Airtable API does not expose this view‑layer configuration information.

> In the grid view, can you match the column summaries?

I don't quite understand what this means; could you explain it in more detail?

> There's something wrong with Rollups and Linked properties I think....

Unfortunately, Rollup itself doesn't return its formula information via the API, so I can only retrieve the values of the specified columns from its corresponding records and then concatenate them for display. See: https://airtable.com/developers/web/api/field-model#rollup

This is the information about Airtable's Rollup:

CleanShot_2025_12_23_at_21_17_40@2x


Below is the PostMan API response of the table schema, and it indeed does not contain any Formula information:
CleanShot_2025_12_23_at_21_16_45@2x

tgrosinger:
> No. What you’re referring to are all view‑level details

Ah okay, that's too bad. Thank you for looking into this.

> I don't quite understand what this means; could you explain it in more detail?

Summaries are displayed in the footer of a table in Obsidian. Here's what they look like:

image

[Obsidian summaries docs](https://help.obsidian.md/bases/syntax#Summaries)

In Airtable they are displayed in the header like this:

image

> Unfortunately, Rollup itself doesn't return its formula information via the API

Darn. Maybe we should just omit Rollup columns then, since they aren't going to display the correct information. We'll need to note that in the docs.

Xheldon:
> In Airtable they are displayed in the header like this:

I see, this also appears to be a view configuration, so the API can't retrieve the information either...

As for omitting the Rollup column or some columns, I'm not sure if that's reasonable, because some formula columns depend on information from other columns; if a column is missing, the formula column will also break.

So my suggestion is to keep the status quo and inform the users. If users need this Rollup calculation, they can manually adjust the relevant functions—after all, I've already converted the Rollup to an Obsidian Formula, and as for how to use it, users can write the formula themselves. What does @kepano   think?

PR #467, opened 2025-12-20
Import title, created, and updated dates from Evernote ENEX files

### LINKED ISSUE #409: Evernote title, created, and updated metadata not imported
The markdown files created when I run the Evernote importer don't file yaml data for the title, nor for the updated and created dates.

Is it possible to add these?

zarino:
I noticed that obsidian-importer wasn’t respecting the created and updated dates that I asked Evernote to include in my ENEX exports. It turns out, it was correctly extracting the metadata, but the dates were missing from the `default-template.ts`, so weren’t added to the frontmatter. By adding the dates to the template, they now get correctly assigned to the frontmatter, and are used elsewhere in the Obsidian app, for example, when you hover over notes in the file explorer, or sort the file explorer by date created or updated.

If the ENEX file was generated without created/updated dates (this is an option the user can specify when exporting from Evernote), then obsidian-importer’s pre-existing template handling will just silently exclude those properties from the frontmatter, so you won‘t end up with an empty `created` or `updated` date.

Then I spotted that [someone else had already requested this feature](https://github.com/obsidianmd/obsidian-importer/issues/409), and had also mentioned that the importer failed to import the `title` from Evernote exports. On this latter point, they weren’t quite right – the importer _did_ correctly import the `<title>` tag from the ENEX file, but it then sanitised that title to remove any characters that would be illegal in a filename (such as the forward slash, `/`), and used that santised version everywhere. My second commit assigns the un-sanitised title as a frontmatter property. Now, when combined with the [front-matter-title plugin](https://github.com/snezhig/obsidian-front-matter-title), this gives a neat solution for displaying imported note titles exactly how they originally appeared in Evernote.

I wasn’t sure whether you want contributors to bump versions, or whether you’d rather do that yourselves. I can remove that commit from the branch if you’d prefer to merge without it.

Fixes #409.

### DIFF
--- a/src/formats/yarle/utils/templates/apply-functions/apply-conditional-template.ts
+++ b/src/formats/yarle/utils/templates/apply-functions/apply-conditional-template.ts
@@ -1,6 +1,9 @@
+import { escapeYamlValue } from '../../yaml-utils';
+
 export const applyConditionalTemplate = (text: string, P: any, newValue?: string): string => {
+	const escapedValue = escapeYamlValue(newValue);
 	return text
-		.replace(new RegExp(`${P.CONTENT_PLACEHOLDER}`, 'g'), newValue || '')
+		.replace(new RegExp(`${P.CONTENT_PLACEHOLDER}`, 'g'), escapedValue)
 		.replace(new RegExp(`${P.START_BLOCK}`, 'g'), '')
 		.replace(new RegExp(`${P.END_BLOCK}`, 'g'), '');
 };
--- a/src/formats/yarle/utils/templates/apply-functions/apply-content-template.ts
+++ b/src/formats/yarle/utils/templates/apply-functions/apply-content-template.ts
@@ -5,7 +5,7 @@ import { applyTemplateOnBlock } from './apply-template-on-block';
 import { getTemplateBlockSettings } from './get-templateblock-settings';
 
 export const applyContentTemplate = (noteData: NoteData, inputText: string, check: Function): string => {
-	const contentTemplateSettings = getTemplateBlockSettings(inputText, check, P, noteData.content);
+	const contentTemplateSettings = getTemplateBlockSettings(inputText, check, P, noteData.content, true);
 
 	return applyTemplateOnBlock(contentTemplateSettings);
 };
--- a/src/formats/yarle/utils/templates/apply-functions/apply-template-on-block.ts
+++ b/src/formats/yarle/utils/templates/apply-functions/apply-template-on-block.ts
@@ -1,4 +1,5 @@
 import { TemplateBlockSettings } from '../template-settings';
+import { escapeYamlValue } from '../../yaml-utils';
 
 export const applyTemplateOnBlock = ({
 	template,
@@ -7,12 +8,14 @@ export const applyTemplateOnBlock = ({
 	endBlockPlaceholder,
 	valuePlaceholder,
 	value,
+	skipYamlEscaping,
 }: TemplateBlockSettings): string => {
 	if (value && check()) {
+		const finalValue = skipYamlEscaping ? value : escapeYamlValue(value);
 		return template
 			.replace(new RegExp(`${startBlockPlaceholder}`, 'g'), '')
 			.replace(new RegExp(`${endBlockPlaceholder}`, 'g'), '')
-			.replace(new RegExp(`${valuePlaceholder}`, 'g'), value);
+			.replace(new RegExp(`${valuePlaceholder}`, 'g'), finalValue);
 
 	}
 	const reg = `${startBlockPlaceholder}([\\d\\D])(?:.|(\r\n|\r|\n))*?(?=${endBlockPlaceholder})${endBlockPlaceholder}`;
--- a/src/formats/yarle/utils/templates/apply-functions/get-templateblock-settings.ts
+++ b/src/formats/yarle/utils/templates/apply-functions/get-templateblock-settings.ts
@@ -1,6 +1,6 @@
 import { TemplateBlockSettings } from '../template-settings';
 
-export const getTemplateBlockSettings = (text: string, check: Function, T: any, value?: string): TemplateBlockSettings => {
+export const getTemplateBlockSettings = (text: string, check: Function, T: any, value?: string, skipYamlEscaping?: boolean): TemplateBlockSettings => {
 	return {
 		template: text,
 		check,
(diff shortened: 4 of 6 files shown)


tgrosinger:
Hi @zarino, thanks for making this contribution!

In my testing it looks like your change does successfully add the created, updated, and title values to the frontmatter. I'm not sure if adding the created and updated there is the best solution though.

In the current release, when I run an import it appears that the "last modified" time on the file is set correctly, but the created time is not. Your change adds these times to the frontmatter, but this does not fix the file ctime.

If you create a markdown file with the following content it will display a Base that demonstrates what I mean:

````
```base
views:
  - type: table
    name: Table
    order:
      - file.name
      - file.ctime
      - created
      - file.mtime
      - updated
```
````

Ideally, we could find a fix that preserves the file ctime, instead of adding it to the frontmatter.

**Comment by zarino** [comment]
Thanks for the review, @tgrosinger.

Yes, you’re right! Well, in fact, I get even better on the current (1.8.2) release – both the ctime _and_ mtime of files created during the import are correctly set. Here’s the four files created after I import [a small `Test.enex` file](https://gist.github.com/zarino/4b573882169ed19f10afd35d0c19f311) under plugin version 1.8.2, Obsidian version 1.10.6, macOS 15.7.2:

```
$ stat -x Test/*                                                
  File: "Test/_resources"
Access: Tue Dec 23 13:42:38 2025
Modify: Tue Dec 23 13:42:38 2025
Change: Tue Dec 23 13:42:38 2025
 Birth: Tue Dec 23 13:42:38 2025
  File: "Test/Charter of the United Nations  UN Charter.md"
Access: Tue Dec 23 13:42:40 2025
Modify: Mon Dec 22 00:00:00 2025
Change: Tue Dec 23 13:42:39 2025
 Birth: Tue May  7 01:00:00 2024
  File: "Test/Declaration of Independence of Cyberspace.md"
Access: Tue Dec 23 13:42:40 2025
Modify: Sun Dec 21 00:00:00 2025
Change: Tue Dec 23 13:42:39 2025
 Birth: Thu Feb  1 00:00:00 1996
  File: "Test/The Declaration of the Rights of Man and of the Citizen.md"
Access: Tue Dec 23 13:42:41 2025
Modify: Sat Dec 20 00:00:00 2025
Change: Tue Dec 23 13:42:39 2025
 Birth: Sun Oct 30 01:00:00 2022
  File: "Test/The Universal Declaration of Human Rights.md"
Access: Tue Dec 23 13:42:40 2025
Modify: Tue Dec 23 00:00:00 2025
Change: Tue Dec 23 13:42:39 2025
 Birth: Sun Apr 14 01:00:00 2019
```

Odd that the ctimes didn’t get set for you. Different OS perhaps?

What initially led me to assume that the times _weren’t_ being correctly set (and therefore that I should add them to the frontmatter) was that _sorting_ the imported files by time didn’t work – despite the correct times appearing in the hover tooltip:

https://github.com/user-attachments/assets/5c1fb928-59ca-4b4d-a833-8430ac1d4d72

Maybe, instead, this is a bug in Obsidian itself?

Out of interest – if you [import the `Test.enex` file](https://gist.github.com/zarino/4b573882169ed19f10afd35d0c19f311), are you able to sort the contents by date?

tgrosinger:
Okay, there's a few things going on here. Let me ask a few questions to make sure I'm understanding what's happening correctly.

With the latest release build of the Importer plugin:

- Does the tooltip show the correct modified and created times?
- Does the Base show the correct modified and created times?
- Does the file explorer sort correctly based on the times displayed?

And then with the changes in this PR, how do the answers to those questions change?

In my testing on Linux, I am seeing the modified time set correctly with and without your changes, but the created time is never set. I haven't been able to reproduce the sorting issue.

**Edit:** It sounds like it is expected that the ctime is not persisted on Linux and Android, but on other platforms it should work.

For reference, setting the created and modified values in the frontmatter should not affect the times show in the file explorer tooltip or how the files are sorted.

zarino:
Thanks @tgrosinger. I realise this thread may be straying somewhat from the original pull request (!!) so let me know if it’s better for me to migrate to an issue tracker somewhere else.

Something really weird is going on.

A few minutes ago, I reset back to obsidian-importer 1.8.2 (latest release) (`git checkout 9d96e29 && npm run build`), imported the `Test.enex`, and while the tooltip times (both created and modified) were correct and the Base showed the correct created/modified times, the sorting of the file explorer didn’t work … Then I deleted the files, reverted back to my modified build of the plugin (`git checkout e53b31e && npm run build`), restarted Obsidian and reimported, expecting it to fix the sorting – but it didn’t! Sorting by date didn’t work with my modified code either 🤔

Confused (sorting _definitely_ worked when I first submitted this PR, otherwise I wouldn’t have submitted it), I cleared the Obsidian vault again, double-checked that I’d checked out the correct commit and re-built (I had), restarted and re-imported, and this time, sorting worked. Suspicious that something weird must be going on, I repeated my first step again—resetting to the latest release (`git checkout 9d96e29 && npm run build`) and re-importing, and sorting the file explorer by the dates now works 🤯 

I’ve since cleared the vault, quit Obsidian, switched between the two versions of the plugin, and re-imported a few times, and I can’t get back to a state where sorting the file explorer by date _doesn’t_ work. Hmmm.

FWIW, I’m doing all of this under macOS 15.7.2, running Obsidian 1.10.6, with a bunch of core plugins active (Backlinks, Bases, Bookmarks, Canvas, Command palette, File recovery, Files, Graph view, Note composer, Outgoing links, Outline, Page preview, Properties view, Quick switcher, Search, Sync, Tags view, Templates, Word count) but no community plugins other than Obsidian Importer itself. My Obsidian repo is stored at `~/Documents/<repo-name>`, but I don’t have iCloud set to sync my Desktop or Documents folders, so no file date shenanigans should be happening there, and I don’t have Obsidian Sync set up yet either.

It _seems_ like sometimes sorting the file explorer by created/modified date works, and sometimes it doesn’t, but I can’t reliably replicate it, and in any case, I don’t have any evidence that it’s specifically to do with the obsidian-importer plugin, so I’d understand if you’d just rather close this PR and move on!

Maybe I might re-work this branch into something that just (optionally?) sets a `title` frontmatter property with the un-sanitised titles from the ENEX file, so people like me, who want forward slashes etc to work in their note titles, can go on to run the [front-matter-title plugin](https://github.com/snezhig/obsidian-front-matter-title) on those imported notes and get a smoother transition from Evernote. Let me know if you’d be open to that.

tgrosinger:
Hm, that's very interesting. I don't have a Mac to continue testing the date/time issue but my team members on macs have not been able to reproduce the issue you're noting yet.

I think adding a toggle for adding the unmodified title to the frontmatter would be great. If you'd like to slim down this PR to that feature, and make it controllable with a setting we could definitely get that merged.

Oh, also please remove the version bump. I'll add that in as a separate commit once we're ready to release.

Thanks again!

Issue #377, opened 2025-04-22
Bug Report: Google Keep Import Fails Silently if Keep Folder Already Exists


Kaos-07:
Bug Report: Google Keep Import Fails Silently if Keep Folder Already Exists

**When using the Google Keep import feature in the Obsidian Importer plugin, the import fails silently if a folder named the same as what you choose already exists in the vault.** No warning is shown to the user. The plugin UI reports all zeros (0 notes, 0 attachments, etc.), giving the impression that the ZIP was invalid or empty. The only hint of the problem is visible via the Developer Console, where an error appears.

To Reproduce:

Have a folder named Keep already present in your Obsidian vault.

Use the Importer plugin to import a correctly structured Google Takeout ZIP (with a Keep/ folder at root).

Click "Import from Google Keep."

No notes are imported. The UI shows "0 notes, 0 attachments imported."

Check Developer Tools → Console to see the real error.

Expected:
A user-facing warning or prompt (e.g., “A folder named Keep already exists. Overwrite, merge, or cancel?”), or at minimum an error message shown in the plugin UI.

Environment:

OS: win10
Obsidian version: v1.8.9
Importer plugin version: v1.6.19

_This silent failure makes it look like something is wrong with the ZIP file or the import process, causing a lot of confusion. It should be more transparent._

Issue #385, opened 2025-05-15
Hopefully there can be a better import of the time in the keep import


mmoneyy:
If you use json to import, there will be no last update time, if you use html, there is no update time of the document attribute, time and date, which is the key for us to sort the document, find the right file, I hope to be better to import the time and date

PR #472, opened 2025-12-24
Google Keep: Add created/modified timestamps & handle duplicates

keanucz:
This fixes issues #377 and #385. I've tested it against my local vault (by importing my own Keep takeout with a dozen or so attachments and 675 notes) and this PR adds the following functionality:

1. Adds `keepCreatedAt` and `keepUpdatedAt` to the frontmatter of all imported notes, output using ISO8601.
2. Adds a dropdown for how duplicate notes should be handled (e.g. if the user reimports the same takeout file but with substantial changes months later for example) - the default behaviour is to skip any already imported duplicate notes.

I've never contributed to an Obsidian plugin before so I'm happy to take any pointers or suggestions for how my implementation could be refined!

### DIFF
--- a/src/format-importer.ts
+++ b/src/format-importer.ts
@@ -146,21 +146,41 @@ export abstract class FormatImporter {
 
 		let { vault } = this.app;
 
-		let folderPath = this.outputLocation;
+		let folderPath = this.outputLocation.trim();
 		if (folderPath === '') {
 			folderPath = '/';
 		}
 
-		let folder = vault.getAbstractFileByPath(folderPath);
+		const normalizedFolderPath = normalizePath(folderPath);
+		let folder = vault.getAbstractFileByPath(normalizedFolderPath) ?? vault.getAbstractFileByPathInsensitive(normalizedFolderPath);
 
 		if (folder === null || !(folder instanceof TFolder)) {
-			await vault.createFolder(folderPath);
-			folder = vault.getAbstractFileByPath(folderPath);
+			try {
+				await vault.createFolder(normalizedFolderPath);
+			}
+			catch (e) {
+				// On case-insensitive filesystems (e.g. Windows), `getAbstractFileByPath`
+				// may not find an existing folder if the casing doesn't match, and then
+				// `createFolder` throws because it already exists. Try again with an
+				// insensitive lookup before failing.
+				folder = vault.getAbstractFileByPathInsensitive(normalizedFolderPath);
+				if (folder instanceof TFolder) {
+					this.outputFolder = folder;
+					return folder;
+				}
+				throw e;
+			}
+
+			folder = vault.getAbstractFileByPath(normalizedFolderPath) ?? vault.getAbstractFileByPathInsensitive(normalizedFolderPath);
 		}
 
 		if (folder instanceof TFolder) {
 			this.outputFolder = folder;
 			return folder;
+		}
+
+		if (folder) {
+			throw new Error(`Output path "${normalizedFolderPath}" exists but is not a folder.`);
 		}
 
 		return null;
--- a/src/formats/keep-json.ts
+++ b/src/formats/keep-json.ts
@@ -1,8 +1,8 @@
-import { FrontMatterCache, Notice, Setting, TFolder } from 'obsidian';
+import { FrontMatterCache, Notice, normalizePath, Setting, TFile, TFolder } from 'obsidian';
 import { PickedFile } from '../filesystem';
 import { FormatImporter } from '../format-importer';
 import { ATTACHMENT_EXTS, ImportContext } from '../main';
-import { serializeFrontMatter } from '../util';
+import { sanitizeFileName, serializeFrontMatter } from '../util';
 import { readZip, ZipEntryFile } from '../zip';
 import { KeepJson } from './keep/models';
 import { sanitizeTag, sanitizeTags, toSentenceCase } from './keep/util';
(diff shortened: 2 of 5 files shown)


src/format-importer.ts
```diff
@@ -146,23 +146,43 @@ export abstract class FormatImporter {
...
-			folder = vault.getAbstractFileByPath(folderPath);
+			try {
+				await vault.createFolder(normalizedFolderPath);
+			}
+			catch (e) {
+				// On case-insensitive filesystems (e.g. Windows), `getAbstractFileByPath`
+				// may not find an existing folder if the casing doesn't match, and then
+				// `createFolder` throws because it already exists. Try again with an
+				// insensitive lookup before failing.
+				folder = vault.getAbstractFileByPathInsensitive(normalizedFolderPath);
```
tgrosinger, line 166: The change made above on 155 seems like it should cover this case and remove the need to repeat the lookup here.

The only case I can think of where this would have an effect is if you are on a case-sensitive OS and the provided path is a file, but an insensitive match finds a folder. I might argue in that case we should just be throwing an error, not finding the insensitive-matching folder on a case-sensitive OS.

src/main.ts
```diff
@@ -486,6 +486,10 @@ export class ImporterModal extends Modal {
 						try {
 							await importer.import(ctx);
 						}
+						catch (e) {
+							ctx.reportFailed('Import', e);
```
tgrosinger, line 490: Were you running into an importer that was throwing errors? Maybe we should fix that more directly?

src/formats/keep-json.ts
```diff
@@ -16,13 +16,61 @@ const NOTE_EXTS = ['json'];
...
 export class KeepImporter extends FormatImporter {
 	importArchivedSetting: Setting;
 	importTrashedSetting: Setting;
+	existingFileSetting: Setting;
 	importArchived: boolean = false;
 	importTrashed: boolean = false;
+	existingFileBehavior: ExistingFileBehavior = 'skip';
 
 	init() {
+		this.importArchived = this.importArchived ?? false;
```
tgrosinger, line 70: What is the purpose of these lines?

src/formats/keep-json.ts
```diff
@@ -45,6 +93,19 @@ export class KeepImporter extends FormatImporter {
 				});
 			});
 
+		this.existingFileSetting = new Setting(this.modal.contentEl)
+			.setName('If note or attachment already exists')
+			.setDesc('When importing into the same folder again, choose to skip, overwrite, or create a duplicate.')
```
tgrosinger, line 98:
```suggestion
			.setDesc('When importing into the same folder again, choose how to handle content that was previously imported.')
```

src/formats/keep-json.ts
```diff
@@ -111,7 +174,14 @@ export class KeepImporter extends FormatImporter {
...
 		const keepJson = JSON.parse(content) as KeepJson;
-		if (!keepJson || !keepJson.userEditedTimestampUsec || !keepJson.createdTimestampUsec) {
+		if (!keepJson || typeof keepJson !== 'object') {
+			ctx.reportFailed(fullpath, 'Invalid Google Keep JSON');
+			return;
+		}
+
+		const createdMs = normalizeEpochMs(keepJson.createdTimestampUsec);
+		const editedMs = normalizeEpochMs(keepJson.userEditedTimestampUsec);
+		if (!createdMs && !editedMs) {
```
tgrosinger, line 184: Should this be `||`?

src/formats/keep-json.ts
```diff
@@ -82,8 +143,10 @@ export class KeepImporter extends FormatImporter {
 			}
 			else if (ATTACHMENT_EXTS.contains(extension)) {
 				ctx.status('Importing attachment ' + name);
-				await this.copyFile(file, assetFolderPath);
-				ctx.reportAttachmentSuccess(fullpath);
+				const imported = await this.copyFile(file, assetFolderPath, ctx);
+				if (imported) {
+					ctx.reportAttachmentSuccess(fullpath);
```
tgrosinger, line 148: Maybe this should be moved into `copyFile` instead? Then it would be next to the other reporting calls in that function.

src/formats/keep-json.ts
```diff
@@ -124,23 +194,58 @@ export class KeepImporter extends FormatImporter {
 			return;
 		}
 
-		await this.convertKeepJson(keepJson, folder, basename);
-		ctx.reportNoteSuccess(fullpath);
+		const imported = await this.convertKeepJson(keepJson, folder, basename, fullpath, ctx);
+		if (imported) {
+			ctx.reportNoteSuccess(fullpath);
```
tgrosinger, line 199: Same comment as above. Move this into `convertKeepJson`?

src/formats/keep-json.ts
```diff
@@ -124,23 +194,58 @@ export class KeepImporter extends FormatImporter {
...
 		// First let's gather some metadata
 		let frontMatter: FrontMatterCache = {};
 
+		// Store both:
+		// - the original Keep fields (as microseconds, matching the JSON field names)
+		// - ISO strings for easy reading/searching in Obsidian
+		const createdMs = normalizeEpochMs(keepJson.createdTimestampUsec);
+		const editedMs = normalizeEpochMs(keepJson.userEditedTimestampUsec) ?? createdMs;
+
+		if (createdMs !== undefined) frontMatter['keepCreatedAt'] = new Date(createdMs).toISOString();
```
tgrosinger, line 246: I've actually been having a very similar conversation about this in !467.

I don't think we need to store these times in the frontmatter. They should be set on the file correctly when creating the file. Setting it on the file means that in a Base you can view the `file.mtime` and `file.ctime`, filter using them, or even transform it into another format.

tgrosinger (commented):
Thank you @keanucz for helping to improve the Keep importer! I've left some comments and questions. Let me know what you think.

kepano:
This should be already partially supported in 2.2.0, and other fields can be added in Templates in the future #609

PR #476, opened 2025-12-29
Add delta sync for Notion API incremental imports

parthalon025:
## Summary

  Implements one-way delta sync for the Notion API importer, enabling efficient re-imports that only update changed content. This enhancement builds on the existing incremental import feature by adding intelligent timestamp-based change detection.

  ## Changes

  - **Delta sync implementation**: Compares Notion's `last_edited_time` with local `notion-updated` frontmatter timestamps to detect changes
  - **Smart update detection**: Only updates files when Notion content is newer than the local copy
  - **Enhanced progress reporting**: Displays separate counts for created, updated, and skipped files
  - **Backward compatibility**: Gracefully handles files imported before this feature (treats them as "unknown age" and skips)

  ## Key Implementation Details

  - Adds `notion-created` and `notion-updated` timestamps to frontmatter on all imported files
  - Refactored `shouldSkipExistingFile()` → `shouldSkipOrUpdateFile()` with decision logic
  - Uses ISO 8601 timestamp format for reliable chronological comparisons
  - One-way sync: Notion → Obsidian (local changes are overwritten on update)

  ## Modified Files

  - `src/formats/notion-api/api-helpers.ts`: Timestamp extraction and comparison logic
  - `src/formats/notion-api.ts`: Main import flow with update detection
  - Progress UI updates to show "X created, Y updated, Z skipped"

  ## Benefits

  - Reduces unnecessary file operations by skipping unchanged files
  - Enables efficient workspace-wide re-imports to catch recent changes
  - Maintains data integrity with timestamp-based version tracking
  - Improves user experience with clearer progress reporting

  ## Testing

  Tested with:
  - Fresh imports (all files created with timestamps)
  - Re-imports with no changes (all files skipped)
  - Re-imports with Notion updates (only changed files updated)
  - Backward compatibility with pre-existing imports

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ data.json
 # Exclude files specific to personal development process
 .hotreload
 .devtarget
+
+# Exclude Claude Code project instructions and artifacts
+CLAUDE.md
+.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,3 @@ data.json
 # Exclude files specific to personal development process
 .hotreload
 .devtarget
-
-# Exclude Claude Code project instructions and artifacts
-CLAUDE.md
-.claude/
--- a/src/formats/notion-api.ts
+++ b/src/formats/notion-api.ts
@@ -1443,6 +1443,7 @@ export class NotionAPIImporter extends FormatImporter {
 							await this.vault.modify(existingFile, fullContent);
 							console.log(`[UPDATE FILE] Successfully updated: ${finalPath}`);
 							this.pagesUpdatedCount++;
+							ctx.reportNoteUpdate(sanitizedTitle);
 						}
 						else {
 							throw new Error(`File not found or not a file: ${finalPath}`);
--- a/src/formats/notion-api.ts
+++ b/src/formats/notion-api.ts
@@ -7,11 +7,12 @@ import { parseFilePath } from '../filesystem';
 
 // Import helper modules
 import { createPlaceholder, PlaceholderType } from './notion-api/utils';
-import { 
-	makeNotionRequest, 
-	fetchAllBlocks, 
-	extractPageTitle, 
+import {
+	makeNotionRequest,
+	fetchAllBlocks,
+	extractPageTitle,
 	extractFrontMatter,
+	extractNotionTimestamps,
 	hasChildPagesOrDatabases
 } from './notion-api/api-helpers';
 import { convertBlocksToMarkdown } from './notion-api/block-converter';
@@ -69,6 +70,7 @@ export class NotionAPIImporter extends FormatImporter {
 	private relationPlaceholders: RelationPlaceholder[] = [];
 	// Progress counters: separate tracking for pages and attachments
 	private processedPagesCount: number = 0; // Total processed (imported + skipped) for progress tracking
+	private pagesUpdatedCount: number = 0; // Track pages updated during delta sync
 	private attachmentsDownloaded: number = 0;
 	// Track Notion ID (page/database) to file path mapping for mention replacement
 	// Stores path relative to vault root without extension: "folder/subfolder/Page Title"
@@ -979,6 +981,7 @@ export class NotionAPIImporter extends FormatImporter {
 			this.processedDatabases.clear();
 			this.relationPlaceholders = [];
 			this.processedPagesCount = 0;
+			this.pagesUpdatedCount = 0;
 			this.attachmentsDownloaded = 0;
 	
 			// Note: getSelectedNodeIds() already populated this.selectedNodeIds and this.totalNodesToImport
@@ -1052,7 +1055,15 @@ export class NotionAPIImporter extends FormatImporter {
 				await this.cleanupNotionIds(ctx);
 			}
 
-			ctx.status('Import completed successfully!');
+			// Report update statistics for delta sync
+			let completionMessage = 'Import completed successfully!';
+			if (this.pagesUpdatedCount > 0) {
+				const updateMessage = `Updated ${this.pagesUpdatedCount} page${this.pagesUpdatedCount > 1 ? 's' : ''}`;
+				console.log(`[Delta Sync] ${updateMessage}`);
+				completionMessage = `Import completed successfully! ${updateMessage}.`;
+			}
+
+			ctx.status(completionMessage);
 		
 		}
 		catch (error) {
@@ -1169,7 +1180,8 @@ export class NotionAPIImporter extends FormatImporter {
 			let pageFolderPath: string; // Folder for child pages/databases
 			let mdFilePath: string;
 			let shouldSkipParentFile = false; // Flag to track if parent file should be skipped
-		
+			let shouldUpdateFile = false; // Flag to track if file should be updated (delta sync)
+
 			if (hasChildren) {
 				// Create folder structure for pages with children
 				// The folder will contain the page content file and child pages/databases
@@ -1187,11 +1199,15 @@ export class NotionAPIImporter extends FormatImporter {
 					await this.createFolders(pageFolderPath);
 				}
 			
-				// Check if file already exists with same notion-id
+				// Check if file needs to be skipped, updated, or created
 				const fileName = `${sanitizedTitle}.md`;
 				const potentialFilePath = normalizePath(`${pageFolderPath}/${fileName}`);
-				shouldSkipParentFile = await this.shouldSkipExistingFile(potentialFilePath, pageId, ctx);
-			
+				const pageLastEdited = page.last_edited_time;
+				const fileDecision = await this.shouldSkipOrUpdateFile(potentialFilePath, pageId, pageLastEdited, ctx);
+
+				shouldSkipParentFile = fileDecision.action === 'skip';
+				shouldUpdateFile = fileDecision.action === 'update';
+
 				mdFilePath = potentialFilePath;
 			}
 			else {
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -25,3 +25,7 @@ data.json
 # Exclude files specific to personal development process
 .hotreload
 .devtarget
+
+# Exclude Claude Code project instructions and artifacts
```
tgrosinger: Let's keep this file generic and avoid adding files specific to individual setups. Please instead be sure to remove those files before commiting.

src/main.ts
```diff
@@ -94,6 +96,10 @@ export class ImportContext {
 				this.importedCountEl = el.createDiv({ cls: 'importer-stat-count', text: this.notes.toString() });
 				el.createDiv({ cls: 'importer-stat-name', text: 'imported' });
 			});
+			el.createDiv('importer-stat mod-updated', el => {
```
tgrosinger, line 99: I'm not sure yet that we should add this, since only one importer would be using it. I think we should continue using the regular imported count for now.

src/formats/notion-api/api-helpers.ts
```diff
@@ -388,6 +387,29 @@ export async function extractFrontMatter(
 	return frontMatter;
 }
 
+/**
+ * Extract notion timestamps from file frontmatter
+ * Used for delta sync to determine if file needs updating
+ * @param fileContent - The raw file content including frontmatter
+ * @returns Object with created and updated timestamps, or null if not found
+ */
+export function extractNotionTimestamps(fileContent: string): {
```
tgrosinger, line 398: Instead of pulling this out of the file content using string matching, please use the metadata cache to read it from the parsed frontmatter.

Change the arguments to accept `app: App`, then

```ts
app.metadataCache.getFileCache(file).frontmatter...
```

src/formats/notion-api/api-helpers.ts
```diff
@@ -318,15 +318,14 @@ export async function extractFrontMatter(
...
-	// Note: created_time and last_edited_time are not added to frontmatter
-	// Users can enable these if needed by uncommenting the code below
-	// if (page.created_time) {
-	// 	frontMatter.created = page.created_time;
-	// }
-	// if (page.last_edited_time) {
-	// 	frontMatter.updated = page.last_edited_time;
-	// }
+
+	// Always add timestamps for delta sync support
```
tgrosinger: We should only write these if incremental import is enabled.

src/formats/notion-api.ts
```diff
@@ -186,7 +188,7 @@ export class NotionAPIImporter extends FormatImporter {
 		// Incremental import setting
 		new Setting(this.modal.contentEl)
 			.setName('Incremental import')
-			.setDesc('Adds a notion-id property to pages so that future imports can skip pages that have already been imported.')
+			.setDesc('Enables delta sync: tracks notion-id and timestamps in frontmatter to skip unchanged pages and update only pages modified in Notion since last import. Keeps your vault synchronized with Notion changes.')
```
tgrosinger: I think it's important to mention that local changes will be overwritten. How about something like this:


```suggestion
			.setDesc('Adds a notion-id and timestamps in frontmatter to skip pages with no changes in Notion since the last import. Local changes will be overwritten if the copy in Notion was modified.')
```

src/formats/notion-api.ts
```diff
@@ -1052,7 +1055,15 @@ export class NotionAPIImporter extends FormatImporter {
 				await this.cleanupNotionIds(ctx);
 			}
 
-			ctx.status('Import completed successfully!');
+			// Report update statistics for delta sync
+			let completionMessage = 'Import completed successfully!';
+			if (this.pagesUpdatedCount > 0) {
+				const updateMessage = `Updated ${this.pagesUpdatedCount} page${this.pagesUpdatedCount > 1 ? 's' : ''}`;
+				console.log(`[Delta Sync] ${updateMessage}`);
```
tgrosinger, line 1062: Remove this `console.log` since it's redundant with the message shown in the UI.

src/formats/notion-api.ts
```diff
@@ -1934,57 +1981,79 @@ export class NotionAPIImporter extends FormatImporter {
...
+	 * @returns Object indicating action: 'skip', 'update', or 'create'
 	 */
-	private async shouldSkipExistingFile(
+	private async shouldSkipOrUpdateFile(
 		filePath: string,
 		notionId: string,
+		pageLastEdited: string,
 		ctx: ImportContext
-	): Promise<boolean> {
+	): Promise<{ action: 'skip' | 'update' | 'create' }> {
```
tgrosinger: This doesn't need to be an object.

```suggestion
	): Promise<'skip' | 'update' | 'create'> {
```

src/formats/notion-api.ts
```diff
@@ -1934,57 +1981,79 @@ export class NotionAPIImporter extends FormatImporter {
...
 		const file = this.vault.getAbstractFileByPath(normalizePath(filePath));
 		if (!file || !(file instanceof TFile)) {
-			return false; // File doesn't exist, don't skip
+			return { action: 'create' }; // File doesn't exist
 		}
 
 		// Read file and extract notion-id from frontmatter
 		try {
 			const content = await this.vault.read(file);
 			const notionIdMatch = content.match(/^notion-id:\s*(.+)$/m);
```
tgrosinger, line 2008: While this function is being updated, it would be good to change this to read out of the metadataCache too. See the other related comment for an example.

src/formats/notion-api.ts
```diff
@@ -1934,57 +1981,79 @@ export class NotionAPIImporter extends FormatImporter {
...
-		
-			if (notionIdMatch) {
-				const existingNotionId = notionIdMatch[1].trim();
-				if (existingNotionId === notionId) {
-					// Same notion-id, skip this file
-					const { basename } = parseFilePath(filePath);
-					ctx.reportSkipped(basename, 'already exists with same notion-id');
-				
-					// IMPORTANT: Register this skipped file in notionIdToPath mapping
```
tgrosinger, line 1968: Please do not remove these comments since they are still relevant and valuable.

src/formats/notion-api.ts
```diff
@@ -1934,57 +1981,79 @@ export class NotionAPIImporter extends FormatImporter {
...
-					// This ensures that relation/mention links can find this page even though it wasn't imported in this session
-					// Without this, we would fail to resolve relations to previously imported pages
-					const filePathWithoutExtension = filePath.replace(/\.md$/, '');
-					this.notionIdToPath.set(notionId, filePathWithoutExtension);
-				
-					// IMPORTANT: Scan for unresolved placeholders from previous imports
```
tgrosinger, line 1974: Please do not remove these comments since they are still relevant and valuable.

src/formats/notion-api.ts
```diff
@@ -1934,57 +1981,79 @@ export class NotionAPIImporter extends FormatImporter {
...
+
+			const existingNotionId = notionIdMatch[1].trim();
+			if (existingNotionId !== notionId) {
+				return { action: 'create' }; // Different notion-id, will rename
+			}
+
+			// Same notion-id found - check if update needed
+			const timestamps = extractNotionTimestamps(content);
+
+			if (!timestamps) {
```
tgrosinger: I would be nice to combine the two skip cases to help prevent accidents in the future causing their logic to drift.

```ts
if (!timestamps || pageLastEdited > timestamps.updated) {
    // Skip
}
else {
    // Needs update
}
```

tgrosinger (changes requested):
Thank you for your work to improve the Notion API importer! I love the idea of making the incremental import aware of last edit times. I haven't tested your changes yet, but it sounds like you have done a good amount already so I just left a few suggestions below:

rgaufman:
Also in general if the importer can import some metadata about each note, like created_at updated_at and if possible other things like created_by that would be awesome! - Right now when notes are imported, there's no way to tell when they were created/edited. I created a separate issue about it here: 

https://github.com/obsidianmd/obsidian-importer/issues/478

Xheldon:
Great idea! @parthalon025 Currently, the Notion API importer’s incremental import works at the file/folder level and doesn’t go down to the file‑content level, so to implement the logic you mentioned, you’d need to keep the notion-id and also retain the notion-updated property for comparison in the incremental import options.

Also, I’m curious what use case drives this need, because typically when users want to incrementally import Notion into Obsidian, they’re doing it to back up Notion content locally, which means they usually don’t modify the original Notion content nor the Obsidian content.

Furthermore, according to @kepano 's view, he seems to dislike stuffing too many Notion importer‑related YAML attributes into the imported content. @kepano what do you think?

Xheldon:
Additionally, I'd like to add that Notion's criteria for detecting modifications to Pages might be very lenient (I haven't tested it); for example, adding comments, moving the cursor, resizing images, and changing layout positions are all considered modifications to Pages. I'm not sure if this is expected.

tgrosinger:
> Also in general if the importer can import some metadata about each note, like created_at updated_at and if possible other things like created_by that would be awesome! - Right now when notes are imported, there's no way to tell when they were created/edited. I created a separate issue about it here:
> 
> #478

Let's please stick with the file ctime/mtime instead of adding these values to the frontmatter.

parthalon025:
> Great idea! @parthalon025 Currently, the Notion API importer’s incremental import works at the file/folder level and doesn’t go down to the file‑content level, so to implement the logic you mentioned, you’d need to keep the notion-id and also retain the notion-updated property for comparison in the incremental import options.
> 
> 
> 
> Also, I’m curious what use case drives this need, because typically when users want to incrementally import Notion into Obsidian, they’re doing it to back up Notion content locally, which means they usually don’t modify the original Notion content nor the Obsidian content.
> 
> 
> 
> Furthermore, according to @kepano 's view, he seems to dislike stuffing too many Notion importer‑related YAML attributes into the imported content. @kepano what do you think?

My primary use case for this addition is to establish a local mirror of my Notion workspace for local AI applications and Claude Code execution. I continue to utilize Notion as my primary “second brain,” comprising over 4,000 pages. This approach mitigates the MCP overhead associated with token usage and ensures that my local copy remains synchronized with any modifications made to the imported pages.

PR #486, opened 2026-01-03
Upgrade TypeScript, ESLint, typescript-eslint and @types/node

ikeyan:
This PR modernizes the TypeScript + lint toolchain and fixes the type issues that show up with newer upstream typings.

## Changes
- **TypeScript:** bump to **5.9.3**
- **ESLint:** bump to **9.39.2**, while keeping the existing legacy `.eslintrc.js`:
  - `npm run lint` now runs ESLint with `ESLINT_USE_FLAT_CONFIG=false` so `.eslintrc.js` continues to be used
- **typescript-eslint:** bump to **8.51.0** (using the consolidated `typescript-eslint` package)
- **Stylistic rules migration:** move deprecated formatting rules (`quotes`, `semi`, `indent`, etc.) to ESLint Stylistic:
  - add `@stylistic/eslint-plugin` (v3; compatible with legacy config)
  - switch formatting rules in `.eslintrc.js` to `@stylistic/*`
- **Node typings + buffer-related fixes:**
  - bump `@types/node` to **20.19.27**
  - tighten `Buffer` / `Uint8Array` generics and convert `Buffer` → `ArrayBuffer` where required
- **Bug fix:** correct `nodeBufferToArrayBuffer` default `length` when `offset` is non-zero

## Compatibility note
ESLint v9 and typescript-eslint v8 require Node.js **^18.18.0 || ^20.9.0 || >=21.1.0**. This PR does not add a Node version field to `package.json` because the repo currently does not specify one.

## Follow-up (out of scope)
This PR intentionally does **not** migrate to ESLint flat config (`eslint.config.*`). A follow-up PR can do that separately.


.eslintrc.js
```diff
@@ -1,45 +1,44 @@
...
+	plugins: ['@typescript-eslint', '@stylistic'],
 	extends: ['plugin:@typescript-eslint/recommended'],
 	parserOptions: {
 		ecmaVersion: 2018,
 		sourceType: 'module',
 	},
 	rules: {
 		'@typescript-eslint/interface-name-prefix': 'off',
-		'@typescript-eslint/no-unused-vars': ['error', { 'vars': 'all', 'args': 'none' }],
+		'@typescript-eslint/no-unused-vars': ['error', { 'vars': 'all', 'args': 'none', 'caughtErrors': 'none' }],
```
tgrosinger: I don't think we need to add this. I would prefer an unused `err` variable be removed.

.eslintrc.js
```diff
@@ -1,45 +1,44 @@
...
 		'@typescript-eslint/no-explicit-any': 'off',
 		'@typescript-eslint/ban-ts-comment': 'off',
 		// '@typescript-eslint/no-use-before-define': ['error', { 'functions': false, 'classes': false }],
 		'@typescript-eslint/no-empty-function': 'off',
-		'@typescript-eslint/ban-types': 'off',
+		'@typescript-eslint/no-unsafe-function-type': 'off',
```
tgrosinger, line 17: Maybe we should create a new issue to address the Yarle code that requires this so that we can remove it. But I agree with your current PR of setting this for now and actually fixing it in another PR later.

.eslintrc.js
```diff
@@ -1,45 +1,44 @@
...
+		'@typescript-eslint/no-wrapper-object-types': 'off',
```
tgrosinger: Is this needed? If I remove this an run `npm lint` I don't any errors.

tgrosinger (commented):
Thanks for submitting this PR again with a narrower focus on the dependency updating. Just a couple small questions/comments.

.eslintrc.js
ikeyan: OK. I prefer it too. (I tried my best to keep everything the same)

.eslintrc.js
ikeyan: Ok, removed.

tgrosinger:
Thank you!

PR #487, opened 2026-01-03
Migrate legacy ESLint config to flat config (ESLint v9)

ikeyan:
Follow-up to #486.

This PR migrates the project from the legacy `.eslintrc.js` format to ESLint’s flat config (`eslint.config.mjs`), and updates the `lint` script accordingly.

Changes:
- Replaced `.eslintrc.js` with `eslint.config.mjs` (flat config).
- Updated `npm run lint` to use the default flat-config flow (removed `ESLINT_USE_FLAT_CONFIG=false` and `--ext .ts`).
- Bumped `@stylistic/eslint-plugin` to a version compatible with ESLint v9.
- Moved the previous “CLI-excluded” generated file into config ignores (`src/z-worker-inline.js`).
- Includes one formatting-only change produced by the updated lint tooling.

Notes:
- Rule intent is kept equivalent to the previous config; no intentional behavior changes.
- `src/z-worker-inline.js` is generated/minified output and should not be linted.

This PR is also based on `master`. Until #486 is merged, the diff will include the changes from #486 as well.


tgrosinger:
I merged the previous PR, so I think this has some conflicts now from including duplicate commits.

ikeyan:
@tgrosinger Thanks, rebased onto new master.

tgrosinger (approved):
Thank you!

PR #488, opened 2026-01-05
Evernote: Optionally import titles and dates as frontmatter values

### LINKED ISSUE #409: Evernote title, created, and updated metadata not imported

zarino:
Follow-up to #467, which also fixes #409.

Two changes:

- Optionally imports the original note title, and creation and modification dates, from ENEX files, as frontmatter values on imported notes.
- Also YAML-escapes those values (well, all frontmatter values, in ENEX imports, apart from tag lists) to avoid special characters (eg: a `title` string containing an unescaped `: `) from invalidating a note’s frontmatter.

screenshot

I used this myself, along with [obsidian-front-matter-title](https://github.com/snezhig/obsidian-front-matter-title) and [notebook-navigator](https://github.com/johansan/notebook-navigator), to smooth my transition from Evernote. I figured it may be useful to others in a similar situation.

The code around Yarle template-handling is pretty complex, so I’ve done my best to work out how to integrate the YAML escaping. But if there’s a different way you’d like me to re-attempt it, let me know.

I guess, also, there’s nothing about these features which are especially Evernote-specific. Users importing from other sources may equally want their YAML frontmatter to be escaped, and even for note titles and dates to be imported as frontmatter values. If that’s the case, I’m happy to generalise this code to work for _all_ importers, but I may need some pointers from you on where best to make these changes.

Happy also to re-submit these commits as two separate PRs, as I realise they’re _technically_ solving two separate issues.

### DIFF
--- a/src/formats/yarle/utils/templates/apply-functions/apply-template-on-block.ts
+++ b/src/formats/yarle/utils/templates/apply-functions/apply-template-on-block.ts
@@ -1,4 +1,5 @@
 import { TemplateBlockSettings } from '../template-settings';
+import { escapeYamlValue } from '../../yaml-utils';
 
 export const applyTemplateOnBlock = ({
 	template,
@@ -7,12 +8,14 @@ export const applyTemplateOnBlock = ({
 	endBlockPlaceholder,
 	valuePlaceholder,
 	value,
+	skipYamlEscaping,
 }: TemplateBlockSettings): string => {
 	if (value && check()) {
+		const finalValue = skipYamlEscaping ? value : escapeYamlValue(value);
 		return template
 			.replace(new RegExp(`${startBlockPlaceholder}`, 'g'), '')
 			.replace(new RegExp(`${endBlockPlaceholder}`, 'g'), '')
-			.replace(new RegExp(`${valuePlaceholder}`, 'g'), value);
+			.replace(new RegExp(`${valuePlaceholder}`, 'g'), finalValue);
 
 	}
 	const reg = `${startBlockPlaceholder}([\\d\\D])(?:.|(\r\n|\r|\n))*?(?=${endBlockPlaceholder})${endBlockPlaceholder}`;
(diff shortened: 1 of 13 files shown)


src/formats/yarle/utils/templates/apply-functions/apply-template-on-block.ts
```diff
@@ -7,12 +8,14 @@ export const applyTemplateOnBlock = ({
 	endBlockPlaceholder,
 	valuePlaceholder,
 	value,
+	skipYamlEscaping,
 }: TemplateBlockSettings): string => {
 	if (value && check()) {
+		const finalValue = skipYamlEscaping ? value : escapeYamlValue(value);
```
tgrosinger, line 14: Instead of adding this new flag, I think it would be simpler and easier to follow if `value` passed into this function is already escaped if it needs to be. So for the two templates where escaping is enabled now, it needs to instead escape `noteData.content` or `noteData.title` before calling `getTemplateBlockSettings`.

src/formats/yarle/utils/templates/apply-functions/get-templateblock-settings.ts
```diff
@@ -1,12 +1,13 @@
 import { TemplateBlockSettings } from '../template-settings';
 
-export const getTemplateBlockSettings = (text: string, check: Function, T: any, value?: string): TemplateBlockSettings => {
+export const getTemplateBlockSettings = (text: string, check: Function, T: any, value?: string, skipYamlEscaping?: boolean): TemplateBlockSettings => {
```
tgrosinger, line 3: Remove `skipYamlEscaping`. See related comment.

tgrosinger (changes requested):
Thanks for these changes, it looks great. Just a couple comments.

Could you please also remove the ctime/mtime frontmatter properties and leave just the title property? I just did a little testing and it seems like the Evernote importer is properly setting the ctime/mtime on the file itself, so for now I'm requesting that this not be duplicated.

Thank you!

kepano:
This should be reimplemented as a generic template feature that all importers can use, see #609.

PR #491, opened 2026-01-12
Add Apple Journal Importer

### LINKED ISSUE #230: Import from Journal
Apple Journal App[1]

Have not seen any current solution. 

Similar to day one or Apple notes it allows easy logging of daily things and photos. 

But since there is already Apple notes supported, maybe journal is not completely different?

[1]: https://en.wikipedia.org/wiki/Journal_(Apple)

wzs:
Hey! I'm a new Obsidian users and wanted to add a proper way to migrate my Apple Journal entries, along with some metadata
- fixes #230

### Features

- **HTML parsing** - imports individual journal entry HTML files from the [exported data](https://support.apple.com/guide/iphone/print-and-export-entries-iph4cad323fe/ios)
- **Date extraction** - parses the formatted date header
- **Frontmatter metadata** - _optionally_ captures metadata including: state-of-mind, contacts mentioned, locations visited, music/media, workouts activity

### DIFF
--- a/src/formats/apple-journal.ts
+++ b/src/formats/apple-journal.ts
@@ -1,10 +1,10 @@
-import { htmlToMarkdown, moment, Notice, Setting } from 'obsidian';
+import { htmlToMarkdown, moment, normalizePath, Notice, Setting, TFile } from 'obsidian';
 import type { FrontMatterCache, TFolder } from 'obsidian';
 import type { PickedFile } from '../filesystem';
 import { fs, os, path } from '../filesystem';
 import { FormatImporter } from '../format-importer';
 import type { ImportContext } from '../main';
-import { parseHTML, serializeFrontMatter } from '../util';
+import { parseHTML, sanitizeFileName, serializeFrontMatter } from '../util';
 
 const DATE_FORMAT = 'dddd, D MMMM YYYY';
 const DEFAULT_OUTPUT_FOLDER = 'Journal';
@@ -33,8 +33,18 @@ const OVERLAY_TEXT_SELECTORS = [
 	'.mediaCategory',
 ];
 
+const DUPLICATE_HANDLING = {
+	Skip: 'skip',
+	ImportUpdated: 'import-updated',
+	CreateCopy: 'create-copy',
+} as const;
+
+type DuplicateHandling = (typeof DUPLICATE_HANDLING)[keyof typeof DUPLICATE_HANDLING];
+const DEFAULT_DUPLICATE_HANDLING = DUPLICATE_HANDLING.ImportUpdated;
+
 export class AppleJournalImporter extends FormatImporter {
 	private frontMatterEnabled = true;
+	private duplicateHandling: DuplicateHandling = DEFAULT_DUPLICATE_HANDLING;
 
 	init(): void {
 		const defaultImportPath = detectDefaultEntriesPath();
@@ -58,6 +68,20 @@ export class AppleJournalImporter extends FormatImporter {
 				toggle.onChange(value => {
 					this.frontMatterEnabled = value;
 				});
+			});
+
+		new Setting(this.modal.contentEl)
+			.setName('Handle duplicate files')
+			.setDesc('How to handle entries that already exist in the vault.')
+			.addDropdown(dropdown => {
+				dropdown
+					.addOption(DUPLICATE_HANDLING.Skip, 'Skip import')
+					.addOption(DUPLICATE_HANDLING.ImportUpdated, 'Import only updated')
+					.addOption(DUPLICATE_HANDLING.CreateCopy, 'Create a copy')
+					.setValue(DEFAULT_DUPLICATE_HANDLING)
+					.onChange(value => {
+						this.duplicateHandling = value as DuplicateHandling;
+					});
 			});
 
 		this.addOutputLocationSetting(DEFAULT_OUTPUT_FOLDER);
@@ -88,8 +112,10 @@ export class AppleJournalImporter extends FormatImporter {
 
 			try {
 				ctx.status(`Importing note ${file.basename}`);
-				await this.importEntry(folder, file);
-				ctx.reportNoteSuccess(file.fullpath);
+				const imported = await this.importEntry(ctx, folder, file);
+				if (imported) {
+					ctx.reportNoteSuccess(file.fullpath);
+				}
 			}
 			catch (error) {
 				ctx.reportFailed(file.fullpath, error as Error);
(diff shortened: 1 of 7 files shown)


src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+			return;
+		}
+
+		ctx.reportProgress(0, this.files.length);
+		for (let index = 0; index < this.files.length; index++) {
+			if (ctx.isCancelled()) return;
+
+			const file = this.files[index];
+			if (file.name === 'index.html') {
+				ctx.reportProgress(index + 1, this.files.length);
```
tgrosinger, line 109: We should also call `ctx.reportSkipped` here.

src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+ */
+function normalizeAssetType(item: Element): string | undefined {
+	const className = Array.from(item.classList).find(cls => cls.startsWith('assetType_'));
+	if (!className) return undefined;
+
+	const rawType = className.slice('assetType_'.length);
+	if (!rawType) return undefined;
+
+	const normalized = rawType
+		.replace(/([a-z])([A-Z])/g, '$1-$2')
```
tgrosinger: Can these values contains numbers? Maybe the first group should be `\w`

src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+
+const DATE_FORMAT = 'dddd, D MMMM YYYY';
+const DEFAULT_OUTPUT_FOLDER = 'Journal';
+const IGNORED_ASSET_TYPES = new Set<string>(['photo', 'live-photo', 'video']);
+const ASSET_TYPE_ALIASES = new Map<string, string>([
+	['generic-map', 'location'],
+	['multi-pin-map', 'location'],
+]);
+const BODY_PARAGRAPH_SELECTOR = '.p2, .p3';
+const OVERLAY_TEXT_SELECTORS = [
```
tgrosinger, line 22: Can you please add a comment about where these values came from? If in a future version of Apple Journal this list needs to be updated I'd like to know where to look.

src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+
+	const normalized = rawType
+		.replace(/([a-z])([A-Z])/g, '$1-$2')
+		.replace(/_/g, '-')
+		.toLowerCase();
+
+	return ASSET_TYPE_ALIASES.get(normalized) ?? normalized;
+}
+
+function parseOverlayTokens(item: Element): string[] {
```
tgrosinger, line 278: It would be helpful for readers to add a comment with a couple examples of the type of data that the `collectOverlayText` will be retrieving and then splitting.

src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+	const items = Array.from(source.querySelectorAll('.assetGrid .gridItem'));
+
+	for (const item of items) {
+		const assetType = normalizeAssetType(item);
+		if (!assetType || IGNORED_ASSET_TYPES.has(assetType)) continue;
+
+		const tokens = parseOverlayTokens(item);
+		if (tokens.length === 0) continue;
+
+		const normalizedTokens = assetType === 'state-of-mind'
```
tgrosinger: Can you please add a comment about why `state-of-mind` tokens are lowercased but the rest are not?

src/formats/apple-journal.ts
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+
+	private async importEntry(folder: TFolder, file: PickedFile): Promise<void> {
+		const htmlContent = await file.readText();
+		const documentEl = parseHTML(htmlContent);
+		const frontMatter: FrontMatterCache = {};
+
+		if (this.frontMatterEnabled) {
+			const tokens = collectFrontMatterTokens(documentEl);
```
tgrosinger: Can this instead be `frontMatter = collectFrontMatterTokens(documentEl)`, since up to this point we haven't used `frontMatter`? Then we can avoid a data copy.

tgrosinger (commented):
This looks really good. Just a few comments. Thank you for adding this!

src/formats/apple-journal.ts
wzs: looking at my exports - no; but it shouldn't hurt - added

src/formats/apple-journal.ts
wzs: sure! I couldn't find any docs for these - so all of that is derived from analyzing my personal exports

src/formats/apple-journal.ts
wzs: just my personal preference - removed that, will keep these tokens as-they-are

wzs:
@tgrosinger thanks for looking over this. I've addressed your comments and also added some options for handling duplicates (based on on Apple Notes options)

src/formats/apple-journal.ts
```diff
@@ -121,7 +147,39 @@ export class AppleJournalImporter extends FormatImporter {
 			}
 		}
 
-		await this.saveAsMarkdownFile(folder, file.basename, mdContent);
+		const sanitizedName = sanitizeFileName(file.basename);
+		const folderPath = folder.path === '/' ? '' : folder.path;
+		const fullPath = normalizePath(
```
tgrosinger: Please use `path.join`. That will also simplify this handling and make it so you can always use `path.join(folderPath, sanitizedName + '.md')`.

src/formats/apple-journal.ts
```diff
@@ -121,7 +147,39 @@ export class AppleJournalImporter extends FormatImporter {
...
+			folderPath ? `${folderPath}/${sanitizedName}.md` : `${sanitizedName}.md`
+		);
+		const existingFile = this.vault.getAbstractFileByPath(fullPath)
+			?? this.vault.getAbstractFileByPathInsensitive(fullPath);
+
+		if (this.duplicateHandling === DUPLICATE_HANDLING.CreateCopy) {
```
tgrosinger, line 156: I believe calling `saveAsMarkdownFile` will throw an error if the file already exists, not create a copy.

wzs: it seem to behave as expected for me (like Apple Notes)
- import a single entry
- import the same entry again with `Create copy`
- a new md file is created (with suffix ` 1`)

tgrosinger: Yup, you're right. I must have misread the upstream code somewhere. Thanks for testing!

tgrosinger:
Merged - thank you @wzs!

zzeitt:
Hello there, I found that the exported journal folder is unselectable when using "Apple Journal Import". Any suggestion about how to import my journals? Thanks in advance.imageimage

PR #496, opened 2026-01-25
Make tsconfig stricter and reflect build script

ikeyan:
Changes in tsconfig:
- rootDir=src to avoid including files outside src
- moduleResolution=bundler and noEmit=true since we use esbuild
- enabled some of [the strict mode family options](https://www.typescriptlang.org/tsconfig/#strict):
  - strictFunctionTypes
  - noImplicitThis
  - alwaysStrict
  - useUnknownInCatchVariables

Other changes are for resolving new type errors.


tgrosinger (approved):

tgrosinger:
Looks great, thank you!

PR #500, opened 2026-02-07
Enable some typescript-eslint recommended rules

ikeyan:
Enabled these rules:

- @typescript-eslint/no-unsafe-function-type
- @typescript-eslint/no-this-alias
- @typescript-eslint/prefer-as-const

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -13,17 +13,12 @@ export default defineConfig(
 		},
 
 		rules: {
-			'@typescript-eslint/interface-name-prefix': 'off',
 			'@typescript-eslint/no-unused-vars': ['error', { 'vars': 'all', 'args': 'none' }],
 			'@typescript-eslint/no-explicit-any': 'off',
 			'@typescript-eslint/ban-ts-comment': 'off',
 			// '@typescript-eslint/no-use-before-define': ['error', { 'functions': false, 'classes': false }],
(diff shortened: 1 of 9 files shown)


eslint.config.mjs
```diff
@@ -13,17 +13,12 @@ export default defineConfig(
 		},
 
 		rules: {
-			'@typescript-eslint/interface-name-prefix': 'off',
```
ikeyan, line 16: This rule was removed and its alternative is [@typescript-eslint/naming-convention](https://typescript-eslint.io/rules/naming-convention/), which is not recommended.

eslint.config.mjs
```diff
@@ -13,17 +13,12 @@ export default defineConfig(
...
 			'@typescript-eslint/no-unused-vars': ['error', { 'vars': 'all', 'args': 'none' }],
 			'@typescript-eslint/no-explicit-any': 'off',
 			'@typescript-eslint/ban-ts-comment': 'off',
 			// '@typescript-eslint/no-use-before-define': ['error', { 'functions': false, 'classes': false }],
-			'@typescript-eslint/no-empty-function': 'off',
```
ikeyan, line 21: [@typescript-eslint/no-empty-function](https://typescript-eslint.io/rules/no-empty-function/) is not recommended now

PR #503, opened 2026-02-16
fix: sanitize page titles and display names used as folder path segments in OneNote importer

aaron-meyers:
OneNote import was not sanitizing page titles and section/group names when creating folders which could result in unexpected subfolder levels, names with leading or trailing whitespace etc.

Tested locally.

### DIFF
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -1,6 +1,6 @@
 import { OnenotePage, SectionGroup, User, PublicError, Notebook, OnenoteSection } from '@microsoft/microsoft-graph-types';
 import { DataWriteOptions, Notice, Setting, TFolder, htmlToMarkdown, ObsidianProtocolData, requestUrl, moment } from 'obsidian';
-import { genUid, extractErrorMessage, parseHTML } from '../util';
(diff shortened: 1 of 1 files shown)


aaron-meyers:
@tgrosinger any concerns with this change? I have some other improvement ideas for the OneNote importer in mind but wanted to get this small fix in before doing anything too drastic 😅

tgrosinger:
Looks good to me overall, except I would prefer if we could avoid the assertions where possible. Perhaps add a fallback instead?

```ts
sanitizeFileName(notebook.displayName ?? 'Untitled Notebook')
sanitizeFileName(page.title ?? 'Untitled')
sanitizeFileName(sectionGroup.displayName ?? 'Untitled')
```

Thank you!

aaron-meyers:
Done. The code appeared to be assuming these were non-empty/non-null already but added the fallback for good measure (this might actually fix a bug because I think it is possible for OneNote pages at least to have an empty title). I used || so that empty string falls back as well.

aaron-meyers:
@tgrosinger can this be merged? thanks!

tgrosinger (approved):
This looks great, thank you!

PR #508, opened 2026-02-27
Fix paragraph indentation issue

### LINKED ISSUE #492: Content missing for indented text blocks during Notion API import
When importing from Notion via the Notion API, any text blocks that are indented (nested as children under another block) are completely omitted from the resulting Markdown file, rather than just losing their formatting.

Image
Image

Xheldon:
Fix #492

Indentation elements below a paragraph were previously ignored; now they are correctly imported as child elements of the paragraph and indented.

But note that some Markdown elements do not support nesting, so you may encounter expected layout issues. For example, if a Paragraph in Notion has a Heading element as a child, when it’s imported into Obsidian it won’t be properly indented and rendered as a Heading; it will just appear as an indented "#" followed by the text content of heading.

### DIFF
--- a/src/formats/notion-api/block-converter.ts
+++ b/src/formats/notion-api/block-converter.ts
@@ -90,7 +90,7 @@ function isEmptyParagraph(block: BlockObjectResponse | null | undefined): boolea
 	if (!block || block.type !== 'paragraph') {
 		return false;
 	}
-	return block.paragraph.rich_text.length === 0;
+	return block.paragraph.rich_text.length === 0 && !block.has_children;
 }
 
 /**
@@ -299,7 +299,7 @@ export async function convertBlockToMarkdown(
 	
 	switch (type) {
 		case 'paragraph':
-			markdown = convertParagraph(block, context);
+			markdown = await convertParagraph(block, context);
 			break;
 		
 		case 'heading_1':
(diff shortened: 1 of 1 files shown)


tgrosinger:
I'm not sure that I understand this change. In obsidian if you indent a block of text, it turns it into a code block. That's not what we want when imported indented text from Notion, right?

Xheldon:
@tgrosinger Are you referring to indentation as a Tab? Currently, the indentation of elements under a paragraph actually adds four spaces, as shown in the figure(ignore the content, i modify it after imported):

Notion content:
image

Obsidian Import it:
image

But it's true that indentation can cause many unexpected problems, such as list nesting. Unordered lists seem fine, but ordered lists don't appear to nest correctly after indentation.

Is this expected?

Xheldon:
Also, I noticed that whether I use a Tab or four spaces for indentation, it doesn't turn into a code block. Am I missing something, or have I set something incorrectly?

tianjiaoding:
Thank you for the efforts! I think the issue that the PR addresses is: currently, when there are indented content under a paragraph in Notion, the imported Obsidian markdown ignores the indented **content** (not only just the indentation). Perhaps it is up to debate how these content should be properly indented in Obsidian, but i think keeping these content is crucial.

PR #509, opened 2026-02-27
Add Reflect JSON importer

### LINKED ISSUE #248: Import from Reflect
My [Reflect](https://reflect.app/) subscription ended, and I'm wanting to migrate to Obsidian. Exporting from Reflect in HTML and then using the importer works okay, but it loses some formatting that was applied. Would be ideal to support importing Reflect JSON

rossbrg:
Fixes #248.

This adds a native Reflect (`.json`) importer so people can move from Reflect to Obsidian without relying on HTML export.

The issue describes formatting loss when importing Reflect via HTML. This importer reads Reflect's JSON export directly and converts the ProseMirror document model to Markdown.

### What is included

- New `ReflectImporter` wired into the importer list (`Reflect (.json)`).
- Conversion support for common Reflect content:
  - headings, paragraphs, blockquotes, horizontal rules, code blocks, embeds
  - bullet lists, task lists, and legacy list nodes
  - backlinks (including aliases), tags, and inline formatting marks
  - images as embeds (with optional local download)
- Import options in the UI:
  - download attachments (opt-in)
  - add YAML tags
  - add YAML created/updated timestamps
  - add YAML title
- Reflect notes are written with preserved note timestamps (`ctime`/`mtime`).

### Hardening work from real export edge cases

- Note filename collision handling is deterministic, while backlink targets stay stable.
- Backlink labels with `]`, `|`, or `\\` fall back to Markdown link form when wiki-link syntax would be unsafe.
- Nested list handling preserves non-paragraph blocks inside list items (for example headings, code blocks, and blockquotes).
- Attachment download handling avoids path collisions and deduplicates repeated image URLs across notes.
- Reflect image proxy URLs (`reflect.academy/_next/image`) are unwrapped to download the original asset URL.
- Import toggle values are initialized and snapshotted at import start to avoid state drift mid-run.

### Files and fixtures

- New importer and converter:
  - `src/formats/reflect-json.ts`
  - `src/formats/reflect/convert.ts`
  - `src/formats/reflect/models.ts`
- Registration and metadata updates:
  - `src/main.ts`
  - `manifest.json`
  - `README.md`
- Regression fixtures:
  - `tests/reflect/sample-reflect-export.json`
  - `tests/reflect/hardening-edge-cases.json`

### Guideline alignment

- Implemented in TypeScript with no new runtime dependencies.
- Importer UI is added through the existing modal/settings pattern.
- Follows the existing importer architecture (`FormatImporter` + registration in `main.ts`).
- Kept import processing sequential to avoid concurrency-related memory risk.

### Testing

- `npm run build`

### Related

- Docs PR: [obsidianmd/obsidian-help#1039](https://github.com/obsidianmd/obsidian-help/pull/1039)


.gitignore
```diff
@@ -25,3 +25,9 @@ data.json
 # Exclude files specific to personal development process
 .hotreload
 .devtarget
+
```
tgrosinger: Please do not add these entries here.

src/main.ts
```diff
@@ -451,6 +459,7 @@ export class ImporterModal extends Modal {
 
 		if (selectedId && importers.hasOwnProperty(selectedId)) {
 			let importer = this.importer = new selectedImporter.importer(this.app, this);
+			importer.init();
```
tgrosinger: Why is this change necessary? This should probably be reverted.

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+}
+
+export class ReflectImporter extends FormatImporter {
+	downloadAttachments: boolean;
+	tagsFrontmatter: boolean;
+	dateFrontmatter: boolean;
+	titleFrontmatter: boolean;
+
+	init() {
+		// Initialize defaults in init() because FormatImporter calls init() from its constructor.
```
tgrosinger, line 26: This contradicts another change in this PR.

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+import { FormatImporter } from '../format-importer';
+import { ImportContext } from '../main';
+import { sanitizeFileName, serializeFrontMatter } from '../util';
+import { sanitizeTag } from './keep/util';
+import { ReflectExport, ReflectNote } from './reflect/models';
+import { convertDocument, ConvertOptions } from './reflect/convert';
+
+const MAX_FILENAME_LENGTH = 200;
+
+function truncateTitle(title: string): string {
```
tgrosinger: Use `truncateText`

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+					);
+					const outputPath = idToOutputPath.get(note.id);
+					if (!outputPath) {
+						throw new Error(`Missing output path for note ${note.id}`);
+					}
+					const outputName = parseFilePath(outputPath).basename;
+
+					// Build frontmatter
+					let content = result.markdown;
+					const frontMatter: Record<string, any> = {};
```
tgrosinger:
```suggestion
					const frontMatter: FrontmatterCache = {};
```

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+			return;
+		}
+
+		const userDNPFormat = this.getUserDNPFormat();
+
+		for (let file of files) {
+			if (ctx.isCancelled()) return;
+
+			ctx.status('Reading ' + file.name);
+			const data = JSON.parse(await file.readText()) as ReflectExport;
```
tgrosinger: This should be in a try/catch and include validation that the imported data matches the expected format.

src/formats/reflect/convert.ts
```diff
@@ -0,0 +1,447 @@
...
+
+	const markdown = convertNodes(nodes, ctx).trim();
+	return { markdown, tags, images };
+}
+
+interface ConvertContext {
+	idToSubject: Map<string, string>;
+	tags: Set<string>;
+	images: ImageInfo[];
+	depth: number;
```
tgrosinger: Appears to be unused.

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+			return 'YYYY-MM-DD';
+		}
+		return plugin.instance.options?.format || 'YYYY-MM-DD';
+	}
+
+	private getNoteTitle(note: ReflectNote, userDNPFormat: string): string {
+		if (note.daily_at) {
+			return moment(note.daily_at).format(userDNPFormat);
+		}
+		return truncateTitle(note.subject);
```
tgrosinger: Provide a fallback title ("Untitled") if the returned value would be empty.


```suggestion
		return truncateTitle(note.subject).trim() || 'Untitled';
```

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+		if (response.status !== 200) {
+			throw new Error(`HTTP ${response.status}`);
+		}
+		return {
+			data: response.arrayBuffer,
+			contentType: response.headers['content-type'] || '',
+		};
+	}
+
+	private async downloadImage(
```
tgrosinger, line 294: Where are images generally downloaded from? Is it a centralized location which would require rate-limiting or backoffs?

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+					const convertOptions: ConvertOptions = {
+						stripInlineTags: this.tagsFrontmatter,
+					};
+					const result = convertDocument(
+						note.document_json,
+						idToSubject,
+						note.subject,
+						convertOptions,
+					);
+					const outputPath = idToOutputPath.get(note.id);
```
tgrosinger, line 440: Consider putting the `note.id` in the frontmatter like is done in other importer format. That would allow incremental imports in the future.

src/formats/reflect-json.ts
```diff
@@ -0,0 +1,347 @@
...
+		};
+		for (const [mime, ext] of Object.entries(map)) {
+			if (mimeType.includes(mime)) return ext;
+		}
+		return '.png';
+	}
+
+	async import(ctx: ImportContext) {
+		// Snapshot option values for this run so they can't drift mid-import.
+		const shouldDownloadAttachments = this.downloadAttachments === true;
```
tgrosinger, line 377: The downloadAttachments toggle is snapshotted at import start (here) with a comment about preventing mid-import drift, but `tagsFrontmatter`, `dateFrontmatter`, and `titleFrontmatter` are read directly from `this` throughout the loop (lines 265, 282, 285, 289).

Either all four settings should be snapshotted for consistency, or none of them need to be — the import modal presumably isn't interactive while the import is running.

tgrosinger (commented):
Thank you for working on adding this new format! Looks like a great start so far.

I left a few comments in line. In addition, please consider padding human-readable error messages to `reportFailed()`. Passing error objects through directly may result in verbose and hard to understand errors, or `[object Object]` being logged. The full error can be logged to the console.

avi-cenna:
@rossbrg Thanks for your work on this; I'm currently in the process of importing Reflect Notes into Obsidian (just using a temporary Python script for the time being).

.gitignore
avi-cenna: FYI if you need to exclude files locally without modifying .gitignore, you can modify your `.git/info/exclude` file.

src/formats/reflect-json.ts
avi-cenna: This is a good question. Reflect assets are stored with a link that looks something like this

https://reflect-assets.app/v1/users/tfgFpw.../999abc79-e45...?key=2227f9…

I doubt they have strict rate limits, but it's probably not a bad idea to put it in there as a safety measure.

.gitignore
rossbrg: thanks, thats on me 🤦

src/formats/reflect-json.ts
rossbrg: Yea, I actually didn't have enough images to even know if there were limits. My tests on my own export only had around 190 images and they all downloaded fine. I'll add some sort of rate limiting

rossbrg:
@tgrosinger Just pushed a follow-up pass for the review comments, appreciate the help.

This removes the local `.gitignore` changes, drops the extra `importer.init()` call, switches title truncation to `truncateText`, adds an `Untitled` fallback, validates the Reflect JSON before import, and makes the `reportFailed()` messages readable while still logging the full error.

I also changed the frontmatter object to `FrontMatterCache`, added `id` to frontmatter, snapshotted all four import settings the same way, and added retry/backoff handling for attachment downloads.

While testing this against my own Reflect export, I ran into a few small converter issues and fixed them here too: hard breaks now render as <br>, tag-only list items are skipped when inline tags are stripped, legacy ordered lists keep their numbering, and headings at the start of a list item stay on the same bullet instead of creating a blank one.

.gitignore
rossbrg: Removed these

src/main.ts
rossbrg: Reverted this

src/formats/reflect-json.ts
rossbrg: This should be consistent again now that the extra `importer.init()` call in `main.ts` is gone

src/formats/reflect-json.ts
rossbrg: Switched to `truncateText`

src/formats/reflect-json.ts
rossbrg: Changed to `FrontMatterCache`

src/formats/reflect-json.ts
rossbrg: Good catch, all four settings are snapshotted at import start now

src/formats/reflect-json.ts
rossbrg: Added `id` to the frontmatter

src/formats/reflect-json.ts
rossbrg: Added retries with backoff for attachment downloads and respect `Retry-After` when the server sends it. Downloads are still sequential.

src/formats/reflect-json.ts
rossbrg: Added the `Untitled` fallback here

src/formats/reflect/convert.ts
rossbrg: Removed the unused `depth` field from `ConvertContext`

src/formats/reflect-json.ts
rossbrg: Added a dedicated parse/validate step here, so malformed exports fail early with a readable error instead of blowing up in the middle of the import.

sigrunixia:
@tgrosinger When this is merged, please merge https://github.com/obsidianmd/obsidian-help/pull/1039 as well.

heytonytan:
@rossbrg thanks for doing this — looking forward to when this reflect importer is merged!

sandstrom:
I'm trying to try out Obsidian and an ability to migrate would be great! 💯 

Thanks so much for doing this! 🙏🏻

frufru666:
Thank you for working on this! Ditching reflect tomorrow and stoked to try it out.

jp-geminum:
Thank you for this 🙏

nimashariatian:
Can we please get this merged?

MachinationsContinued:
I found file attachments are dropped with the current import setup.

`convertNode` handles `image`, and the `default` branch falls back to recursing into
`node.content`. But a `file` node carries its payload entirely in `attrs` with no children,
so it returns `''` and vanishes with nothing in the output or the log.

The node looks like this:

```json
{ "type": "file", "attrs": { "url": "https://…signed…", "fileName": "memo.m4a" } }
```

A `file` node is the same shape as an `image`, so it can reuse the whole download path. The
only difference is that Obsidian renders embeds for images, audio, video and PDF only, so
`![[memo.docx]]` gives an empty embed box where a plain link is wanted:

```ts
case 'file': {
	return convertImage(node, ctx);
}
```

with `ImageInfo` carrying whether to embed, set from the extension:

```ts
const EMBEDDABLE = /\.(png|jpe?g|gif|bmp|svg|webp|avif|mp3|wav|m4a|ogg|flac|3gp|mp4|webm|ogv|mov|mkv|pdf)$/i;

ctx.images.push({ url: src, fileName, placeholder, embed: EMBEDDABLE.test(fileName) });
```

The importer already replaces each placeholder with the saved attachment path once the
download finishes; it just reads the flag there to decide between `![[…]]` and `[[…]]`.

rossbrg:
@MachinationsContinued Good catch, thanks. Confirmed: `file` nodes were dropped with nothing in the log.

Fixed basically as you sketched it. `file` nodes now go through the same download path as images, and the importer picks `![[...]]` vs `[[...]]` from the file it actually saved, so mime and extension inference still count. One difference from your sketch: when downloads are off, non-image files get a labeled plain link instead of ``, since Obsidian only renders external embeds for images. An m4a behind an expired signed URL would show a broken image box either way.

While I was in there I fixed a few more things I hit testing against my own 1,283-note export. Marks that Reflect splits across text nodes now serialize as one range instead of `**a****b**` garbage. Legacy lists aren't double-spaced anymore. Attachment filenames get sanitized before they become wikilinks, and placeholder replacement no longer trips on `$` in URLs. Also trimmed the retry logic down to a simple bounded backoff.

@tgrosinger should be ready for another look when you have a minute.

ssawczyn:
While testing this importer against a large real-world export (1,400+ notes), I found a case where a corrupted image asset gets silently mislabeled rather than failing cleanly.

**Repro:** an `image` node whose asset URL is dead at the source (an external site the note was capturing a screenshot from), but Reflect's own asset proxy still returns `HTTP 200` for it — with `content-type: text/html` and an error page as the body, instead of real image data. Redacted shape (URL genericized, everything else real):

```json
{"type":"image","attrs":{"settled":true,"alt":"","src":"https://reflect-assets.app/v1/users/<redacted>/<redacted>?key=<redacted>","title":"","resizable":false,"fileName":"image.html"}}
```

Note the proxy's own `fileName` already hints at the problem (`image.html`, not `image.png`/`.jpg`).

**What happens:** `fetchAttachmentData` treats the `200` as success, so it never reaches the failure path. `getExtension` can't map `text/html` to an image MIME type, falls through to the URL (no extension on a signed URL like this), and lands on the `isImage ? '.png' : '.bin'` fallback — so the HTML error page gets written to disk as `image.png` and embedded with `![[...]]`. The note ends up with a "successfully downloaded" attachment that's actually a broken-image icon in Obsidian, with nothing in the import log flagging it.

**Suggested fix:** after a successful fetch, validate that `contentType` actually starts with `image/` when `attachment.isImage` is true; treat a mismatch as a failure so it falls into the existing graceful fallback (``) instead of being saved under a false extension.

rossbrg:
@ssawczyn Nice find, thanks for the detailed repro. Confirmed: a 200 with an HTML body sailed straight through and got written to the vault as a "successful" attachment.

Fixed in a1de5c9, with a follow-up in dc3e7e8 for header casing on the requestUrl path. Image downloads now fail when the response has an affirmatively non-image content type, so that case falls back to the remote `` reference and actually shows up in the import log. I loosened your suggested check slightly: missing content-type headers and octet-stream still download, since some storage backends serve real images that way. The mismatch isn't retried since it's deterministic, not transient.

One small note on the repro: with `fileName: "image.html"` present, the pre-fix build saved the error page as `image.html` and plain-linked it, since the embed decision reads the saved extension. The broken `![[image.png]]` embed you described happens when `fileName` is missing. Both were wrong, and both are covered now.

PR #510, opened 2026-03-01
Add file prefix format option to Bear importer

KrauseFx:
Adds a `File prefix format` option to the Bear importer, matching the Apple Notes behavior for note filenames.

This lets users prefix imported Bear note names with creation dates using a format like `YYYY-MM-DD`, or leave it blank for no prefix.

Apple Notes implementation similar to https://github.com/obsidianmd/obsidian-importer/blob/master/src/formats/apple-notes.ts

image

### DIFF
--- a/src/formats/bear-bear2bk.ts
+++ b/src/formats/bear-bear2bk.ts
@@ -1,8 +1,10 @@
-import { DataWriteOptions, normalizePath, Notice, TFile, Setting } from 'obsidian';
+import { DataWriteOptions, moment, normalizePath, Notice, TFile, Setting } from 'obsidian';
 import { path, parseFilePath } from '../filesystem';
 import { FormatImporter } from '../format-importer';
 import { ImportContext } from '../main';
 import { readZip, ZipEntryFile } from '../zip';
+
+const LOCAL_STORAGE_KEY = 'bear-importer-file-prefix';
 
 type Metadata = {
 	id: string;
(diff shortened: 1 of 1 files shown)


src/formats/bear-bear2bk.ts
```diff
@@ -22,11 +24,35 @@ export class Bear2bkImporter extends FormatImporter {
...
+					this.filePrefixFormat = value;
+					localStorage.setItem(LOCAL_STORAGE_KEY, value);
+				};
+
+				t.setValue(storedPrefix);
+				t.setPlaceholder('YYYY-MM-DD');
+				t.onChange(v => updatePrefix(v));
+				// In practice, users may click Import while focus is still in this field.
+				// 'change' may not fire yet, so persist on each keystroke too.
+				t.inputEl.addEventListener('input', () => updatePrefix(t.inputEl.value));
```
tgrosinger, line 53: I do not think this is necessary. In my testing I see that the `onChange` hook is called before this listener every time I press a key.

src/formats/bear-bear2bk.ts
```diff
@@ -77,6 +103,9 @@ export class Bear2bkImporter extends FormatImporter {
 	}
 
 	async import(ctx: ImportContext): Promise<void> {
+		// The importer modal clears/rebuilds its DOM before import runs, so read from
+		// persisted settings here instead of trusting the now-detached input element.
+		this.filePrefixFormat = localStorage.getItem(LOCAL_STORAGE_KEY) || '';
```
tgrosinger, line 108: I'm not sure why this is necessary. `this.filePrefixFormat` is the same before and after this line in my testing.

src/formats/bear-bear2bk.ts
```diff
@@ -174,7 +203,7 @@ export class Bear2bkImporter extends FormatImporter {
 							}
 
 							idMapping[metadata?.id] = {
-								filename: fileName,
+								filename: file.basename,
```
tgrosinger, line 206: Why does this need to change?

tgrosinger (commented):
Thanks for making this improvement!

I'm wondering if the note prefix should apply to attachments too, like it does in the Apple Notes importer format?

kepano:
This should be implemented when we get to #609 so that it's standard across importers

PR #511, opened 2026-03-02
Fix/bear import

randy:
This is intended to fix several issues I was having with the Bear importer.
- list indentation, specifically if the list contained a code block
- tag processing for multi-word tags in Bear
- tag processing for tag-like text in inline code
- tag processing for tag-like text in code blocks
- tag processing for hex color text (e.g., #00ff00)
- other edge-case tag processing issues that may or may not have been a result of my changes
    -  I ended up typing out many "possible" tags in Obsidian, noting whether they were considered tags internally or not, and then putting that logic into the various regex and processing steps.

I tested these changes by running my Bear export of +1600 notes through the changes over and over. My settings for the import were to to flatten nested tags and not to store note identifiers in front matter. The "flatten" option may have side-effects, but from what I understand, as Bear user, I would not expect that sort of result.

### DIFF
--- a/src/formats/bear-bear2bk.ts
+++ b/src/formats/bear-bear2bk.ts
@@ -167,32 +167,31 @@ export class Bear2bkImporter extends FormatImporter {
 
 	private transformOutsideInlineCode(line: string, transformLine: (line: string) => string): string {
 		let inInline = false;
-		let current = '';
-		let result = '';
+		let currentStart = 0;
+		const parts: string[] = [];
 
 		for (let i = 0; i < line.length; i += 1) {
-			const ch = line[i];
-			if (ch === '`') {
-				if (!inInline) {
-					result += transformLine(current);
-					current = '';
-					inInline = true;
-				}
-				else {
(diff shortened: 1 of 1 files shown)


src/formats/bear-bear2bk.ts
```diff
@@ -119,6 +283,7 @@ export class Bear2bkImporter extends FormatImporter {
 							ctx.status('Importing note ' + mdFilename);
 							let mdContent = await entry.readText();
 							mdContent = this.removeMarkdownHeader(mdFilename, mdContent);
+							mdContent = this.fixListIndentation(mdContent);
```
tgrosinger: `fixListIndentation` and `transformOutsideInlineCode` use different methods for detecting a code block and seem to have different functionality. One supports a list-prefixed code block, while the other does not.

I think this can cause normalizeTagsAndCollect to not recognize a codeblock and modify tags within it.

kepano:
Thanks, ended up bringing some of these ideas into #629

Issue #471, opened 2025-12-24
Apple Notes import: issues with attachments, tags, brackets, and more


natkuhn:
I abandoned Evernote back in September. I had been using Evernote since 2010, and I had about about 4,000 notes that I moved to Apple Notes (AN). I am now switching to Obsidian, originally so that I could make my notes legible to Claude. But I am also discovering that it's awesome in its own right. Like Apple notes choked trying to move all those notes to iCloud, so I had to keep the bulk of them on my Mac, and put new ones in iCloud or move some by hand. Obsidian handles that many notes in iCloud with absolutely no problem. So the whole thing, including the importer, has been a godsend.

I did encounter several issues. I fixed most of them after import with help from Claude, but it should be possible to make the importer take care of most of these automatically at import time and it will make the Apple Notes imports significantly better IMO. I also got Claude's help drafting this but I'm going over it carefully before I submit.

### 1. Attachments without file extensions

Some attachments (likely from the Evernote web clipper) have UUID filenames with no extension (e.g., `0A32B83C-3BCC-4F65-B77D-B2EA8D76B37B`). The importer copies these files but the links don't work in Obsidian because it doesn't recognize them without an extension.

Clicking the link offers to create a new note instead of finding the file:

<img width="616" height="341" alt="Image" src="https://github.com/user-attachments/assets/49f511af-9745-4d42-b239-cf47973f9e0e" />

The file type could be detected (e.g., using magic bytes or the `file` command) and an appropriate extension added during import.

And here is the pathname that it came from in Apple Notes:
`/Users/nat/Library/Group Containers/group.com.apple.notes/Accounts/LocalAccount/Media/0A32B83C-3BCC-4F65-B77D-B2EA8D76B37B/1_04BA02DF-71E3-46DF-8A75-59B7CF9248C7/0A32B83C-3BCC-4F65-B77D-B2EA8D76B37B`

### 2. Files without extensions cannot be found via Obsidian UI

Related to # 1: even using Obsidian's search/file browser, files without extensions cannot be found through the UI—despite being present in the attachments directory on disk.

Can't find it in Obsidian:

<img width="616" height="261" alt="Image" src="https://github.com/user-attachments/assets/ac353e3b-aa62-4cc7-a5e6-d11a9f97b04c" />

And yet, it is in the attachments folder (sorry, I see that the name is trimmed down both in the search bar and in the finder, but please take my word for that I just pasted it in):

[...]

PR #516, opened 2026-03-12
fix: Add embed prefix for Apple Notes attachments and scans

heyalchang:
## Summary

- Attachments imported from Apple Notes render as text links instead of inline embeds because the `!` prefix is missing from the generated markdown link
- This adds the `!` prefix in both `convert-note.ts` (attachments) and `convert-scan.ts` (scans/drawings)
- The third `generateMarkdownLink` call in `getInternalLink()` is intentionally left unchanged — it produces note-to-note links (`[[note]]`), not embeds

This matches Apple Notes' own behavior where all attachments are rendered inline.

## Test plan

- Import an Apple Notes database containing image attachments and scanned documents
- Verify that images render inline in the resulting markdown files rather than appearing as `[[filename.jpg]]` text links
- Verify that internal note links (applenotes://...) still render as `[[note]]` links, not embeds

Addresses #471 (item 3: attachments not embedded inline)

### DIFF
--- a/src/formats/apple-notes/convert-note.ts
+++ b/src/formats/apple-notes/convert-note.ts
@@ -371,7 +371,7 @@ export class NoteConverter extends ANConverter {
 
 		const attachment = await this.importer.resolveAttachment(id, attr.attachmentInfo!.typeUti);
 		let link = attachment
-			? `\n${this.app.fileManager.generateMarkdownLink(attachment, parentNotePath)}\n` 
+			? `\n!${this.app.fileManager.generateMarkdownLink(attachment, parentNotePath)}\n` 
 			: ` **(error reading attachment)**`;
 		
 		if (this.importer.includeHandwriting && row.ZHANDWRITINGSUMMARY) {
--- a/src/formats/apple-notes/convert-scan.ts
+++ b/src/formats/apple-notes/convert-scan.ts
@@ -32,7 +32,7 @@ export class ScanConverter extends ANConverter {
 			if (!file) file = await this.importer.resolveAttachment(row.ZMEDIA, row.ZTYPEUTI);
 
 			if (file) {
-				links.push(this.importer.app.fileManager.generateMarkdownLink(file!, parentNotePath));
+				links.push('!' + this.importer.app.fileManager.generateMarkdownLink(file!, parentNotePath));
 			}
 			else {
 				return '**Cannot decode scan**';


mirnovov:
Note this issue is due to [a breaking change in Obsidian 1.10.0](https://obsidian.md/changelog/2025-10-01-desktop-v1.10.0/).

xml:
Big "Me Too" here: I can confirm the bug breaks every single image, scan, PDF in my 8000-note Apple Notes import on MacOS 26.4.1. Replacing the `!` in a few of them restores proper behavior. I can provide additional test/validation if helpful.

Further, I've reviewed the changes that @mirnovov mentioned in 1.10.0, and this bug is expected based on those changes. This is a no-brainer: 

* the 1.10.0 changelog (2025-10-01) literally says: "FileManager#generateMarkdownLink no longer prepends '!' on attachment  links." 
* This PR restores them

I was going to write my own PR, but found this one first. Also, AFAIK, @mirnovov is the original author of Apple Notes importer (thank you!), so his endorsement is helpful, right?

I'd really love to migrate to Obsidian, but I can't/won't until this is fixed and I can have greater confidence in the transition. (Yes, I recognize I could task an agent with fixing all the embeds, but I'm trying to go 'upstream-first' here, and it also kinda just prevents me from getting any further in my validation that the rest of the import works.)

xml:
Also note that this addresses Issue #540 , which may be a more concise expression of this particular issue than #471, which is more of a grab-bag and more complicated to review.

soid:
I understand the Importer plugin may not be something most regular Obsidian users rely on frequently, but as a new user this creates a pretty rough first experience. I found this issue while researching the correct way to embed attachments. I’ll try building from this branch in the meantime, but I wanted to emphasize that this matters a lot for new users who are actively exploring and evaluating the ecosystem.

rondawes:
I just saw that the importer had been updated and hoped that the update fixed the image import issue. But it still does not work properly. I can manually add the ! and get an image to display inline but that's not practical for the amount of Apple notes I need to import. I'm not very familiar with GitHub to know if the fix posted above was included in this most recent Importer version or not. If it was included, it does not fix the issue. If it was not, please include it as soon as possible. Thanks.

tgrosinger (approved):

tgrosinger:
Thank you!

Issue #517, opened 2026-03-13
Apple Notes: import failed because incorrect header check


chamzzzzzz:
Failed: "xxxxxxx" because incorrect header check

Xheldon:
Could you include the macOS version, the Apple Notes version (if any), the Obsidian Importer version, the Obsidian version, and any characteristics of the notes (e.g., special symbols in the title) for troubleshooting?

As far as I know, the Apple Notes data structure has gone through several upgrades, and there may be issues on some devices.

urosch:
I also have the same problem.

Apple Notes: Version 4.13 (3146.101.28)
Importer plugin: 1.8.4

Working on a MacOs 26.4.1

Nothing special about failed notes that would help...

chamzzzzzz:
The issue affects only a small number of notes that were created several years ago and have never been modified since. These notes contain only simple plain text with no complex elements.

If you duplicate one of the problematic notes, the new copy does not trigger any errors.

If you edit and save the original problematic note directly, the error also stops occurring.

Apple Notes: 4.11, 4.13
Obsidian: 1.12.7
Obsidian Importer: 1.8.4
macOS: Sonoma 14.3.1, Tahoe 26.2

chamzzzzzz:
Sorry, I accidentally closed this issue.

Xheldon:
> The issue affects only a small number of notes that were created several years ago and have never been modified since. These notes contain only simple plain text with no complex elements.此问题仅影响少量在几年前创建且从未修改过的笔记。这些笔记仅包含简单的纯文本，没有复杂元素。
> 
> If you duplicate one of the problematic notes, the new copy does not trigger any errors.如果你复制其中一条有问题的笔记，新的副本不会触发任何错误。
> 
> If you edit and save the original problematic note directly, the error also stops occurring.如果直接编辑并保存原始有问题的笔记，错误也会停止出现。
> 
> Apple Notes: 4.11, 4.13 Apple 备忘录：4.11，4.13 Obsidian: 1.12.7 Obsidian：1.12.7 Obsidian Importer: 1.8.4 Obsidian 导入器：1.8.4 macOS: Sonoma 14.3.1, Tahoe 26.2 macOS：Sonoma 14.3.1，Tahoe 26.2

The phenomenon you described looks a lot like legacy data compatibility, and it seems Apple has already implemented a migration—just needs a trigger for the migration, such as when editing.

[...]

PR #518, opened 2026-03-13
Apple Notes: fix "incorrect header check" decompression error (#517)

chamzzzzzz:
Use unzipSync instead of gunzipSync to automatically detect and handle different compression formats (gzip, deflate) used by Apple Notes.

The gunzipSync method only handles gzip format, but Apple Notes may uses different compression formats (gzip, deflate) which has a different header format causing "incorrect header check" when decompressing note data.


agjones:
this fixes issue i had on my end, would love to see this merged

chamzzzzzz:
Since there were unrelated commits, I closed the original PR and opened a new one. #535

PR #521, opened 2026-03-22
Add Apple Notes highlight color export

Yonderanim:
This PR adds support for importing text highlights from Apple Notes. 

Apple Notes stores emphasis colors in field 14 of the `AttributeRun` protobuf message. This update extracts that field and maps the 5 available colors (Purple, Pink, Orange, Mint, Blue) to corresponding HTML `background-color` styles, allowing highlights to be preserved in the exported Markdown files.

### DIFF
--- a/src/formats/apple-notes/convert-note.ts
+++ b/src/formats/apple-notes/convert-note.ts
@@ -7,6 +7,7 @@ import {
 	ANAttributeRun,
 	ANBaseline,
 	ANColor,
+	ANEmphasisColor,
 	ANConverter,
 	ANDocument,
 	ANFontWeight,
@@ -104,7 +105,7 @@ export class NoteConverter extends ANConverter {
 			else if (attr.attachmentInfo) {
 				converted += await this.formatAttachment(attr, parentNotePath);
 			}
-			else if (attr.superscript || attr.underlined || attr.color || attr.font || this.multiRun == ANMultiRun.Alignment) {
+			else if (attr.superscript || attr.underlined || attr.color || attr.emphasisColor || attr.font || this.multiRun == ANMultiRun.Alignment) {
 				converted += await this.formatHtmlAttr(attr);
 			}
 			else {
@@ -200,6 +201,7 @@ export class NoteConverter extends ANConverter {
 
 		if (attr.font?.pointSize) style += `font-size:${attr.font.pointSize}pt;`;
 		if (attr.color) style += `color:${this.convertColor(attr.color)};`;
+		if (attr.emphasisColor) style += `background-color:${this.convertEmphasisColor(attr.emphasisColor)};`;
 
 		if (attr.link && !NOTE_URI.test(attr.link)) {
 			if (style) style = ` style="${style}"`;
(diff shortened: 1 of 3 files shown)


johntheo (approved):
LGTM

mirnovov:
The hardcoded emphasis colours you have chosen don’t seem to have good contrast in both dark and light modes. 

One possible solution would be to use Obsidian’s CSS variables, though it does make things less portable outside Obsidian. Alternatively, you could choose colours with partial transparency, which would work since the original Apple Notes colours seem to also be semi-transparent.

src/formats/apple-notes/convert-note.ts
```diff
@@ -406,6 +408,23 @@ export class NoteConverter extends ANConverter {
...
+			case ANEmphasisColor.Pink:
+				return '#ffd6ea';
+			case ANEmphasisColor.Orange:
+				return '#ffe1bf';
+			case ANEmphasisColor.Mint:
+				return '#d9f5e8';
+			case ANEmphasisColor.Blue:
+				return '#d7e6ff';
+			default:
+				return '#fff3a3';
```
mirnovov: Per my earlier comment, the colours should be compatible with both Obsidian's dark and light modes IMO. (I'm not an Obsidian employee, but I did develop the original Apple Notes Importer)

```suggestion
			case ANEmphasisColor.Purple:
				return '#a357d750';
			case ANEmphasisColor.Pink:
				return '#ea455a50';
			case ANEmphasisColor.Orange:
				return '#f09a3850';
			case ANEmphasisColor.Mint:
				return '#59c4bd50';
			case ANEmphasisColor.Blue:
				return '#3c7df550';
			default:
				return '#f0d63850';
```

Yonderanim:
HI @mirnovov,

Thank you very much for the thoughtful feedback, and thank you for developing the original Apple Notes Importer in the first place — this tool is so, so incredibly helpful.

That is a very good point. I do not use dark mode myself, so I honestly did not consider the contrast issue. Also, I should mention that I am not a programmer; all of these changes were made by ChatGPT. My main goal here was simply to make sure the highlight information from Apple Notes would not be lost during import.

At this stage, the different highlight colors are being preserved (or tagged) as different background colors in the exported .md files, so users can still identify them inside Obsidian, and then use find-and-replace to further refine the styles to their liking.

I really appreciate your suggestion of using semi-transparent colors. It’s a much more elegant solution that maintains compatibility across different themes. I've updated the code with your suggested hex values.

Update:
I just tried a few things, and it seems that using the CSS snippets below (generated by Gemini) will produce highlight styles almost identical to those in Apple Notes for both light and dark themes.

```
mark[style*="#a357d750"], span[style*="#a357d750"],
mark[style*="#ea455a50"], span[style*="#ea455a50"],
mark[style*="#f09a3850"], span[style*="#f09a3850"],
mark[style*="#59c4bd50"], span[style*="#59c4bd50"],
mark[style*="#3c7df550"], span[style*="#3c7df550"] {
    display: inline !important;
    padding: 0 2px !important;
    margin: 0 -2px !important;
    border-radius: 6px !important;
    -webkit-box-decoration-break: clone !important;
    box-decoration-break: clone !important;
}

/* 1. Purple */
mark[style*="#a357d750"], span[style*="#a357d750"] {
    background-color: rgba(186, 85, 211, 0.15) !important;
    color: #c23cff !important;
}

/* 2. Pink */
mark[style*="#ea455a50"], span[style*="#ea455a50"] {
    background-color: rgba(255, 99, 132, 0.15) !important;
    color: #ff3b5c !important;
}

/* 3. Orange */
mark[style*="#f09a3850"], span[style*="#f09a3850"] {
    background-color: rgba(255, 165, 0, 0.15) !important;
    color: #ff9800 !important;
}

/* 4. Mint */
mark[style*="#59c4bd50"], span[style*="#59c4bd50"] {
    background-color: rgba(64, 224, 208, 0.15) !important;
    color: #00bfa6 !important;
}

/* 5. Blue */
mark[style*="#3c7df550"], span[style*="#3c7df550"] {
    background-color: rgba(66, 133, 244, 0.15) !important;
    color: #1a73e8 !important;
}
```

PR #528, opened 2026-04-08
OneNote: Improve reliability and enterprise account support

### LINKED ISSUE #313: One note import sing in button is tot showed
IMG_0067

Hello,
There is no sing-in button to MS on import dialog in portrait. Landscape is ok.

iOS 17.7 iPhone 15 pro.

### LINKED ISSUE #334: Importer Broken for OneNote - Limited how fast can be imported
Hello!

I have a fresh install of Obsidian notes and im trying to import my Onenote notebooks from my M365 account.

However, when trying to import the data after logging in it just says "Microsoft Onenote has limited how fast notes can be imported please try again in 30 minutes to continue importing."

After waiting over an hour and trying again multiple times it is still not working.  What could be causing this and how can we fix it?

### LINKED ISSUE #390: Constant error message about rate limiting
I have tried this for days now, and it's constantly notifying me that I have to wait 30 minutes because Microsoft OneNote has limited how fast notes can be imported. However, I have not even selected a notebook to import. It's purely the initial initiation. So, how should I move forward in this?

### LINKED ISSUE #398: Importer skipping items that are not previously imported
Whilst investigating https://github.com/obsidianmd/obsidian-importer/issues/334, I came across a strange bug where when importing via Microsoft OneNote, the Importer will skip items that it has supposedly previously imported, and yet those items are nowhere to be found within the Obsidian vault.

I double-checked the directory it wants to import into and I'm perplexed.

### LINKED ISSUE #440: Importer for OneNote does not import anything
When running the importer for OneNote (I have multiple M365 accounts and tried importing from different accounts - always same result), the importer immediately shows the "Import data into Obsidian" dialog, stating "0 imported, 0 attachments, 0 remaining, 0 skipped, 0 failed".

As on my company account I have multiple OneNotes, some being on my personal Employee-OneDrive, others being on Teams-Sharepoints, I would actually also like to chose which OneNotes to import.

gurusince88:
## Summary

This PR improves the reliability of the OneNote importer, particularly for enterprise M365/SharePoint accounts.

**Addresses:** #440, #398, #390, #334, #313

## Changes

- **OAuth scopes**: Add `notes.read.all` for M365/SharePoint enterprise accounts (fixes #440, #334)
- **Error messages**: Show specific messages for 401/403/429/504 errors instead of generic rate-limit message (fixes #334)
- **Exponential backoff**: Replace fixed 60s retry with exponential backoff + jitter for better rate limit handling (fixes #390)
- **Frontmatter tracking**: Switch from `data.json` to `onenote-id` frontmatter (like Notion API importer) to fix false "previously imported" skips (fixes #398)
- **Mobile UI**: Ensure sign-in button is visible on mobile portrait (fixes #313)

## Test plan

- [x] Fresh import with personal Microsoft account
- [x] Re-import with existing files (should skip correctly)
- [x] Re-import after deleting some files (should re-import deleted)
- [x] Test with M365 enterprise account (if available)
- [x] Verify mobile portrait shows sign-in button

## Breaking changes

Notes imported after this change will have `onenote-id` in their frontmatter. The old `data.json` tracking is no longer used, so users who previously imported notes will need to either:
1. Re-import with "Skip previously imported" disabled, or
2. Their old notes won't be recognized as previously imported

### DIFF
--- a/src/formats/onenote.ts
+++ b/src/formats/onenote.ts
@@ -1,5 +1,5 @@
 import { OnenotePage, SectionGroup, User, PublicError, Notebook, OnenoteSection } from '@microsoft/microsoft-graph-types';
-import { DataWriteOptions, Notice, Setting, TFolder, htmlToMarkdown, ObsidianProtocolData, requestUrl, moment } from 'obsidian';
+import { DataWriteOptions, Notice, Setting, TFile, TFolder, htmlToMarkdown, ObsidianProtocolData, requestUrl, moment, normalizePath } from 'obsidian';
 import { genUid, extractErrorMessage, parseHTML } from '../util';
 import { FormatImporter } from '../format-importer';
 import { ATTACHMENT_EXTS, AUTH_REDIRECT_URI, ImportContext } from '../main';
@@ -10,7 +10,7 @@ import { MathMLToLaTeX } from 'mathml-to-latex';
 
 const LOCAL_STORAGE_KEY = 'onenote-importer-refresh-token';
 const GRAPH_CLIENT_ID: string = '66553851-08fa-44f2-8bb1-1436f121a73d';
-const GRAPH_SCOPES: string[] = ['user.read', 'notes.read'];
+const GRAPH_SCOPES: string[] = ['user.read', 'notes.read', 'notes.read.all'];
 // Regex for fixing broken HTML returned by the OneNote API
 const SELF_CLOSING_REGEX = /<(object|iframe)([^>]*)\/>/g;
 // Regex for fixing whitespace and paragraphs
@@ -80,6 +80,10 @@ export class OneNoteImporter extends FormatImporter {
 	rememberMe = false;
 	refreshToken?: string;
 	lastSuccessfulFetchTime: number = performance.now();
+	// Exponential backoff configuration
+	private readonly baseRetryDelayMs: number = 1000;
+	private readonly maxRetryDelayMs: number = 60000;
+	private readonly rateLimitBaseDelayMs: number = 30000; // 30s base for rate limits
 
 	async init() {
 		this.addOutputLocationSetting('OneNote');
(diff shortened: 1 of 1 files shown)


Lenivvenil:
Thanks for tackling this — the error message improvements and exponential backoff are solid. A few things worth addressing before merge:

---

**🔴 Performance: O(n×m) file reads in `findExistingFileByOnenoteId`**

```typescript
const files = this.vault.getMarkdownFiles().filter(f => f.path.startsWith(outputFolder.path));
for (const file of files) {
    const content = await this.vault.read(file); // full disk read per file, per page
```

This reads every file in the output folder for every page being imported. With 500 OneNote pages and 500 existing files, that's 250,000 sequential disk reads. CONTRIBUTING.md calls this out explicitly: *"Be performance minded. Your code will be used in vaults with 10,000 or even 100,000 notes."*

The fix is to build an in-memory `Map<onenoteId, filePath>` once before the import loop by scanning existing files, then do O(1) lookups per page:

```typescript
// Before the import loop:
const existingIds = new Map<string, string>();
for (const file of this.vault.getMarkdownFiles()) {
    if (!file.path.startsWith(outputFolder.path)) continue;
    const content = await this.vault.read(file);
    const match = content.match(/^onenote-id:\s*["']?([^"'\n]+)["']?$/m);
    if (match) existingIds.set(match[1].trim(), file.path);
}

// Inside the loop:
if (!this.importPreviouslyImported && page.id && existingIds.has(page.id)) {
    progress.reportSkipped(page.title!, 'previously imported');
    continue;
}
```

---

**🔴 Rate limit retries are now bounded — regression from current behaviour**

Original code deliberately did not increment `retryCount` on 429:
```typescript
// don't increment the retryCount because we were told to backoff
retryCount  // <-- original
```

This PR changes it to `retryCount + 1`, which means after `MAX_RETRY_ATTEMPTS` (5) rate-limit responses the import hard-fails. Users with large notebooks who hit Microsoft's throttling — the exact users this PR aims to help — will now get failures instead of eventual success.

Suggest keeping 429 retries unbounded (don't increment `retryCount`) while still applying exponential backoff. The 90-minute stall watchdog already provides an upper bound.

---

**🔴 `shouldSkipExistingFile` is dead code**

The method is defined (~line 1107) but never called — `import()` calls `findExistingFileByOnenoteId` instead. Either remove it or wire it in.

---

**🟡 `notes.read.all` requested for all users**

Personal account users don't need `notes.read.all` and will see a more aggressive permission consent dialog with no benefit. Consider showing a targeted message on `40004` error — *"If you're using a work/school account, you may need your admin to grant additional permissions"* — rather than requesting the scope upfront for everyone.

---

**🟡 YAML quoting on `onenote-id`**

OneNote page IDs contain `!` and other characters that YAML parsers may interpret:

```typescript
// Current — may break YAML parsers:
const frontmatter = `---\nonenote-id: ${page.id}\n---\n\n`;

// Safer:
const frontmatter = `---\nonenote-id: "${page.id}"\n---\n\n`;
```

The regex match should also handle quoted values: `/^onenote-id:\s*["']?([^"'\n]+)["']?$/m`

---

**🟡 Migration: existing users will silently re-import everything**

Switching from `data.json` ID tracking to frontmatter means all users who previously imported will lose their tracking state on upgrade and re-import every page. Worth calling out in the PR description with a clear migration note, and potentially adding a one-time migration step that reads old `previouslyImportedIDs` from `data.json` and uses them as a fallback when no frontmatter is found.

---

The error branching in `showContentAreaErrorMessage`, the backoff for non-rate-limit errors, and the mobile UI fix all look good. Happy to discuss any of the above.

kepano:
Superseded by #611

PR #550, opened 2026-05-07
Improve Onenote importer reliability and observability

philou:
- Fix: section fetch failure no longer aborts the whole import — a return in the error handler was replaced with continue, so a failing section is reported and skipped while remaining sections proceed normally.
- Fix: handwritten note SVG files — filenames are now sanitized (removing characters invalid on some filesystems), and existing files are overwritten on reimport instead of throwing a "file already exists" error.
- Fix: SVG write failures are now visible to the user — previously swallowed silently; they now appear in the import log alongside other failures.
- Feature: import progress logging — [OneNote Importer]-prefixed console messages trace each section and note as it is imported, making it easier to diagnose where an import stalls or skips content.
- Refactor: selectedIds stores full section context — previously a flat string[] of section IDs, it is now a SelectedSection[] that carries the full hierarchical path (notebook > section group > section). This eliminates a post-hoc tree walk to recover names, and makes all log and user-facing messages show the full path (e.g. Learning > History > Books) instead of just the section name.


kepano:
Superseded by #611

PR #581, opened 2026-06-22
Fix OneNote progress for skipped pages

fnnikita:
## Summary

When the OneNote importer skips a previously imported page, it currently reports the skip and then continues without updating the progress counter. This can leave the remaining count and progress bar behind the actual number of processed pages.

This change updates progress before continuing when a previously imported page is skipped.

## Validation

- `npx -y pnpm@10.20.0 run build`
- `npx -y pnpm@10.20.0 test`


kepano:
Fixed via bad0a2b8114c7b504c8536c9685514b4c031fa10

PR #585, opened 2026-06-25
Notion API: import data URL attachments

### LINKED ISSUE #539: Failed to process base64 attachment while importing from Notion
While importing via Notion API, a bunch of errors like this appeared in the log:
```
Failed: "Attachment: gif;base64,iVBORw0KGgoAAAA..." because ClientRequest only supports http: and https: protocols
```

As the log doesn't show the name of the page in question, it is not even possible to know which page got its attachment lost during importing, hence recover manually. So a related suggestion is to display the page or .md file name associated with each error.

markoub:
Fixes #539.

## Summary
- Decode Notion API `data:` attachment URLs locally instead of passing them to `requestUrl`, which only supports HTTP(S).
- Infer missing filename extensions from the data URL media type.
- Use the current page title fallback for unnamed data URL attachments so logs and filenames do not include base64 payload text.

## Validation
- `pnpm install --frozen-lockfile`
- `pnpm exec tsx --test tests/notion/attachment-helpers.test.ts`
- `pnpm run test`
- `pnpm exec eslint src/formats/notion-api/attachment-helpers.ts tests/notion/attachment-helpers.test.ts`
- `git diff --check origin/master...HEAD`
- `pnpm run build`

### DIFF
--- a/src/formats/notion-api/attachment-helpers.ts
+++ b/src/formats/notion-api/attachment-helpers.ts
@@ -10,6 +10,11 @@ import { splitext, parseFilePath } from '../../filesystem';
 import { extensionForMime } from '../../mime';
 import { NotionAttachment, AttachmentResult, BlockConversionContext, FormatAttachmentLinkParams } from './types';
 
+interface DownloadedAttachment {
+	arrayBuffer: ArrayBuffer;
+	contentType?: string;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


kepano:
Closed in favor of f3600d437d82446caa80cbc54eba03cae0a677c7

PR #631, opened 2026-08-14
perf: cache Airtable ID lookups

rpreslar4765:
Summary


kepano:
I don’t think this change improves performance. It adds extra setup for one-time searches through small lists, while the import's real time is spent on network requests and file writes. It's adding complexity without any noticeable benefit.

